PR #21, opened 2015-07-30
Small fixes to bugs found while learning to work with Oskari

hatapitk:


hatapitk:
We use Slf4J at Luke and I noticed that stack traces did not reach our log files; 777ed52 fixes that.

ZakarFin:
Good job, thanks!

PR #22, opened 2015-08-17
Allow using unmodified content-resources for custom DBs

hatapitk:
Populating the Oskari database as documented in http://oskari.org/documentation/backend/database-populate would require maintaining a custom fork of content-resources project. Even if the changes would be small this is something I wanted to avoid. The changes in this commit allows us to maintain our application specific definitions in a separate Git repository and still take advantage of improvements made to oskari-server. An additional advantage of these changes is that we can now run all of the schema-modifying operations from a system that does not have Maven installed.

Please check if this approach seems sensible to you (changes to README.md have all the details). It should at least not break any existing workflows but of course if there already is a better way of achieving the same goal we could use that as well.

### DIFF
--- a/content-resources/README.md
+++ b/content-resources/README.md
@@ -7,6 +7,10 @@ The database properties can be modified in oskari-server/content-resources/src/m
 Also these database-settings can be overridden by using -Doskari.env=myenv parameter on maven call.
 This reads db.properties as base, reads an additional properties file db-myenv.properties and overrides any keys found in base properties with the env-specific properties.
 
+Finally it is possible to provide properties using any .properties file by specifying the path to the file
+as the first command line argument. This is useful if you wish to run the DB populator as a standalone unmodified
+.jar file.
+
 ## Populating database content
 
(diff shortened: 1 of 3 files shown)


ZakarFin:
Thanks for the pull request! Can you take a look at the approach used on ELF and ASDI application setups:
- https://github.com/arctic-sdi/oskari-server-extensions
- https://github.com/elf-oskari/oskari-server-extensions/tree/develop/elf-resources

These haven't been tested using runnable jar, only as part of the webapp init. Both of these define overriding app-default.json in the webapp so no additional command line params are needed. Do you think your approach brings additional benefits?

ZakarFin:
I like the overlay consept and while custom property values could already be defined in an external classpath resource as oskari-ext.properties, this enables one to define a path for the properties. I'll merge this after testing nothing gets broken in our setups.

ZakarFin:
Merged into develop. Going to master in about 2 weeks.

PR #23, opened 2015-08-21
Fix missing primary key from oskari_permission

hatapitk:
Did not see a reason why "id" did not have primary key constraint. We use pgAdmin at Luke and it requires that either primary key or OIDs must be present in order to allow editing the data through the graphical UI.


ZakarFin:
Thanks! Added this as a flyway script so existing setups will automatically get the upgrade: https://github.com/nls-oskari/oskari-server/commit/b189cc26fe6799eabf731b716f9977de7e53fac3

PR #24, opened 2015-09-02
Layers and views were not found from overlay directory

hatapitk:
This will fix the existing feature to work with layers and views that were missed.


ZakarFin:
Thanks!

PR #26, opened 2015-10-11
Fixed the tests for the control-base and JSON syntax.

keskival:
The control-base tests failed when run together, because of incomplete clean up after some tests.
Also cleaned up incorrect JSON.


ZakarFin:
Good job, thanks!

PR #28, opened 2016-01-14
Handle email when changing user attributes as admin

hatapitk:
Previously if the admin modified user properties then user's email address would be deleted.

This change requires a corresponding addition in the Oskari frontend code. I have it here but need to extract it from our own codebase and make a pull request against nls-oskari/oskari


ZakarFin:
Thanks, we will test and merge this and the fronted counterpart shortly

kuosman:
Thanks.  Tested and merged.

PR #29, opened 2016-01-27
Fix oskari.proxy.userlayertile.{user,pass} not being passed to GeoServer

hatapitk:


ZakarFin:
Thanks, we'll test & merge this shortly.

PR #32, opened 2016-05-11
Save new passwords using BCrypt hash instead of MD5

hatapitk:
For login both hash formats are still supported. This change is the service layer change needed for the user registration functionality developed at Luke.


ZakarFin:
Thanks! Sorry for the long wait, but the code has now been merged to develop and will be included in the 1.37 release later this month.

PR #33, opened 2016-06-28
User registration and editing of personal details

hatapitk:
Here is finally the new module for user registration functionality - sorry for the long delay. We still use an older version of oskari-server in Biomassa-atlas and won't have time to upgrade just now. But this is an independent module so I would expect it to work.

If I have missed something please let me or @an9l know and we will try to fix that for you. I will be on vacation for most of the time in July and first half of August.

### DIFF
--- a/control-users/pom.xml
+++ b/control-users/pom.xml
@@ -74,10 +74,5 @@
             <artifactId>spring-webmvc</artifactId>
         </dependency>
         
-        <dependency>
-            <groupId>fi.nls.oskari</groupId>
-            <artifactId>servlet-map</artifactId>
-        </dependency>
-
     </dependencies>
 </project>
--- a/control-users/pom.xml
+++ b/control-users/pom.xml
@@ -63,6 +63,12 @@
             <artifactId>powermock-api-mockito</artifactId>
             <scope>test</scope>
         </dependency>
+        
+        <dependency>
+			<groupId>javax.mail</groupId>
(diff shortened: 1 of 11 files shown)


ZakarFin:
Just starting the merge this for the next version. Is there a template passwordReset.jsp that you could include?

ZakarFin:
Also is there another page for user registration that isn't included?

ZakarFin:
Merged the code with some changes implemented and yet to come:
- Added flyway script to create the email-token database table (core oskari module: V1_39_3__create_table_user_registration.sql)
- Renamed database table oskari_email_expiry to oskari_users_pending
- Migrated from ibatis to mybatis
- Added the dummy/template JSPs that can be overridden as any JSP in Oskari
- The functionality is now contained in oskari-control-users maven module (JSP/Mybatis configs)
- Moved dependency from servlet-map to webapp-map to make it more opt-in in extensions. Also need to enable functionality with 'allow.registration=true' in oskari-ext.properties (opt-in with default app)
- Other mandatory configs are:

	oskari.email.sender=<sender email>
	oskari.email.host=<smtp.domain.com>

- URL changes that I made: /register -> /user/register
- Still need to modify JSPs and change some urls (getting all of them to point under /user/)

	/biomass/user/registrationSuccess ->  ???
	/biomass/user/passwordChanged -> ???

- Still needs some missing property-values for localization etc

ZakarFin:
Thanks! :)

PR #35, opened 2016-11-16
Create servlet-printout/test-output directory

jokimaki:
A lot of servlet-printout tests fail on clean checkout because the test-output directory doesn't exist. This patch creates it before running test cases.

### DIFF
--- a/servlet-printout/src/test/java/fi/nls/oskari/printout/ws/jaxrs/map/WebServiceMapProducerResourceTest.java
+++ b/servlet-printout/src/test/java/fi/nls/oskari/printout/ws/jaxrs/map/WebServiceMapProducerResourceTest.java
@@ -15,6 +15,8 @@ import org.opengis.referencing.operation.TransformException;
 
 import javax.xml.stream.FactoryConfigurationError;
 import javax.xml.stream.XMLStreamException;
+
+import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
@@ -35,6 +37,12 @@ public class WebServiceMapProducerResourceTest {
 
     {
         runner.setClientInfo(clientInfo);
+    }
+
+    @Before
+    public void setupTestOutputDirectory() {
+        File testOutputDir = new File("test-output");
+        testOutputDir.mkdir();
     }
 
     @Before


ZakarFin:
Strange we haven't encountered this. Running mvn clean install works just fine. Which JDK/Maven are you using?

ZakarFin:
Suppose this explains it: https://github.com/nls-oskari/oskari-server/blob/master/servlet-printout/pom.xml#L20. Thanks, I'll take a look at this.

jokimaki:
Note that mvn clean does not remove test-output directory even without skipTests true tag in pom.xml, only target. Please try repeating the issue by deleting test-output manually.

ZakarFin:
Yep. We haven't been running the tests ourselves because there's the skipTests flag on the pom.xml that causes tests not being run on that module... Not sure of the reason for this. Propably just a slip. Anyway my confusion is explained by this since with the skipTests-flag I don't need to have a test-output directory (don't even have it in the first place) and for that reason haven't stumbled into this issue. I'll merge this to the develop-branch and remove the skipTests-flag once tested.

jokimaki:
Great, thanks!

Even after this fix is applied, I'm getting several errors (on Oracle JDK 1.8.0_40/Maven 3.3.9):

```
Tests in error:
  testGeojsPrintTestJsonAsPPTX(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest): Invalid args for test
  testWmtsPrintJsonAsPNG(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest)
  testWmtsPrintZoom3JsonAsPNG(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest)
  testWmtsPrintZoom7JsonAsPNG(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest)
  testWmtsPrintZoom7png8opacityJsonAsPNG(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest)
  testWmtsPrintZoom7jpegopacityJsonAsPNG(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest)
  testWmtsPrintZoom7png8opacityJsonAsPDF(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest)
  testWmtsPrintZoom7png8opacityJsonAndTemplateAsPDF(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest)
  testWmtsLayerPrintZoom7png8opacityAndTemplateWithTableContentAsPDF(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest)
  testWmtsLayerPrintZoom7png8opacityAndTemplateWithPagedTableContentAsPDF(fi.nls.oskari.printout.ws.jaxrs.map.WebServiceMapProducerResourceTest)

Tests run: 34, Failures: 0, Errors: 10, Skipped: 0
```

It seems that PPTX support was removed in 42c6f617798557d7f2d1368afdd53d6ffc45c0b5, but the test are still there and the readme still mentions it.

I'm not sure about the WMTS tests, do they fail bacause of invalid test JSON data or a bug in constructing LayerDefinition.

ZakarFin:
I think the tests just haven't been updated for some time. The WMTS layers config got changed a while back and pptx (and some other like docx maybe) support was dropped on a library update change since we didn't use it anyway and as such can't maintain it properly.

ZakarFin:
Merged the change to develop-branch. There's still a ticket open for making the test actually work

jokimaki:
I was thinking that it might actually be better to create the test files under target instead of test-output, so that they could be deleted with `mvn clean`. But let's take it one step at the time, it would be great to have the test pass with minimal changes first.

By the way, if you always merge pull requests to develop branch first, I would suggest that you change the default branch in GitHub to develop, and/or document this practice in the root README.md. I think it would save you from some manual work if pull requests are created on top of a correct branch :)

PR #36, opened 2016-11-28
Geometry transformation improvements and axis order policy to lon,lat for each projection in the geometries

kessu:
1. geometry transform improvements and AxisOrder management for all projections as lon,lat in geometries

2. New property in oskari-ext.properties for file import
   Default source Crs, if source crs is not found in import file (SHP and MIF)
   userlayer.default.source.epsg=EPSG:3067

3. Improvements in DescribeFeatureType parser / wfs 2.0

4. Improments in Mapclick buffertolerance calculation when map crs is geographical (degree units)

5. Optional setup for wfs layers  / "reverseXY" --> lat,lon order in wfs service; longSrsName --> long srsName syntax in GetFeature
   e.g. in oskari_maplayer.attributes column
   {
    "reverseXY": {
      "EPSG:4326":true
    },
    "longSrsName":{
      "EPSG:4326":true
    }
   }


ZakarFin:
Thanks, I'll check these probably next week. Tried the reverseXY on WFS already. Worked like a charm!

ZakarFin:
Rebased the commits to current develop head (found under Commits on Dec 14, 2016) to make them easier to track if there's problems later on, but everything seems to work perfectly regardless of the coordinate order being enforced or not.  Thanks!

PR #38, opened 2017-02-17
Wfs data export functionality for writing common spatial file formats…

kessu:
Wfs data export functionality for writing common spatial file formats

::: Environmetnt Preparations :::

1. Install GDAL library for geotools in the Transport servlet platform
   - http://oskari.org/documentation/backend/configuring-dataset-import
   - version 2.1.2 is used in the tests
2. Setup security for GDAL library
   - https://www.identrust.com/certificates/trustid/root-download-x3.html
   - >keytool -importcert -alias identrust -keystore  
       .../Java/jdk1.8.0_60/jre/lib/security/cacerts" -file identrust.crt
3. Create "export" path under <Jetty home> with write access
    (Default temporary folder is used, if not created or folder is not available)

::: Test use :::

Remark: There is not yet UI in Oskari for this functionality

1. Open WFS layer in Oskari and find out its layer id
   - console> Oskari.getSandbox().findAllSelectedMapLayers();
2. Open browser console and execute below commands
   var event = Oskari.getSandbox().getEventBuilder('WFSSetExport')('<format>','<layerId>');
 
   Oskari.getSandbox().notifyAll(event);
   e.g.
   var event = Oskari.getSandbox().getEventBuilder('WFSSetExport')('GPX','26');
 
   Oskari.getSandbox().notifyAll(event);
3. Format value is OGR GDAL code (http://www.gdal.org/ogr_formats.html)
   - some formats tested succefully
     GEOJSON, Geopackage (GPKG), GPX, KML, ESRI Shapefile, Mapinfo file, PDF

4. Result file(s) are written into /export directory  in transport environment and file id is sent to Oskari
   - result is a single file  or file directory 

::: TODO :::
1. Set export UI into featuredata2 bundle ?
2. Add download permission check into frontend UI and also into Transport service
   - now there is only common view permission check in Transport
3. New action to deliver the export file(s) to the user
   - add copyright.txt file to result set files
   - zip result set
   - deliver the download link to the user

::: REMARKS :::
1. Each format in ORG Gdal library may have special limits for the result file
   - property field name length (Shape)
   - property fields (e.g. GPX)
   - supported property types
   - supported geometry types (e.g. GPX: Point,LineString and MultiLineString)
   - look at more details http://www.gdal.org/ogr_formats.html

2. Some workarounds has been made in WFSExport.java to handle those above mentioned limits
   - property field names has been truncated for Shape file (OGR truncate didn't work)
   - Object type properties are strigified ( not optimal solution)
   - Geometry type is retyped in GPX format in certain cases

### DIFF
--- a/service-map/src/main/java/fi/nls/oskari/map/geometry/ProjectionHelper.java
+++ b/service-map/src/main/java/fi/nls/oskari/map/geometry/ProjectionHelper.java
@@ -7,6 +7,9 @@ import fi.nls.oskari.domain.geo.Point;
 import fi.nls.oskari.log.LogFactory;
 import fi.nls.oskari.log.Logger;
 import fi.nls.oskari.util.JSONHelper;
+import org.geotools.data.DataUtilities;
+import org.geotools.data.Query;
+import org.geotools.data.simple.SimpleFeatureSource;
 import org.geotools.feature.FeatureCollection;
 import org.geotools.feature.FeatureIterator;
 import org.geotools.geojson.geom.GeometryJSON;
@@ -280,4 +283,36 @@ public class ProjectionHelper implements PointTransformer {
             featuresIter.close();
         }
     }
+
+    /**
+     *  Transforms all geometries in Featurecollection
+     *  REMARK crs (sourceCRS) must be defined in feature default geometrydescriptor
+     *  (fc.getSchema().getGeometryDescriptor().getCoordinateReferenceSystem() must be not null
+     * @param featureCollection
+     * @param sourceCrs
+     * @param targetCrs
+     * @return
+     */
+    public static FeatureCollection<SimpleFeatureType, SimpleFeature> transformFeatureCollection(FeatureCollection<SimpleFeatureType,
+            SimpleFeature> featureCollection, String sourceCrs, String targetCrs) {
+
+        if(sourceCrs.toUpperCase().equals(targetCrs.toUpperCase())){
+            return featureCollection;
+        }
+
+        try {
+            CoordinateReferenceSystem crsSource = CRS.decode(sourceCrs, true);
+            CoordinateReferenceSystem crsTarget = CRS.decode(targetCrs, true);
+            Query query = new Query();
+            query.setCoordinateSystem(crsSource);
+            query.setCoordinateSystemReproject(crsTarget);
+            SimpleFeatureSource source = DataUtilities.source(featureCollection);
+            return source.getFeatures(query);
+        }
+        catch (Exception e) {
+            log.error(e, "Transform featureCollection geometries failed! Params:: sourceSRS: ", sourceCrs, "targetSRS: ", targetCrs);
+        }
+        return featureCollection;
+    }
+
 }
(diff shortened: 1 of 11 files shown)


ZakarFin:
Thanks! I'll check this out after the next release.

service-map/src/main/java/fi/nls/oskari/map/geometry/ProjectionHelper.java
```diff
@@ -280,4 +283,36 @@ public static void swapGeometryXY(FeatureCollection<SimpleFeatureType, SimpleFea
...
+            Query query = new Query();
+            query.setCoordinateSystem(crsSource);
+            query.setCoordinateSystemReproject(crsTarget);
+            SimpleFeatureSource source = DataUtilities.source(featureCollection);
+            return source.getFeatures(query);
+        }
+        catch (Exception e) {
+            log.error(e, "Transform featureCollection geometries failed! Params:: sourceSRS: ", sourceCrs, "targetSRS: ", targetCrs);
+        }
+        return featureCollection;
```
ZakarFin: This should throw an (Runtime)Exception rather return the initial un-reprojected featureCollection.

servlet-transport/src/main/java/fi/nls/oskari/transport/TransportService.java
```diff
@@ -719,6 +724,32 @@ private void setFilter(SessionStore store, String json, Map<String, Object> para
...
+        if (store.containsLayer(layerId)) {
+            // TODO: Check export/load permission
+            Job job = createOWSMapLayerJob(createResultProcessor(parseRequestId(params)),
+                    JobType.EXPORT, store, layerId, false, true, false, false);
+            if (job != null) {
+                jobs.add(job);
+            }
+
+        }
+    }
```
ZakarFin: This should probably throw an exception or otherwise signal the frontend that we're not actually going to do anything about the request (in else branch):
            throw new ServiceRuntimeException("Layer is not selected",
                    WFSExceptionHelper.ERROR_SET_PROCESS_REQUEST_FAILED, WFSExceptionHelper.WARNING_LEVEL);

service-map/src/main/java/fi/nls/oskari/map/geometry/ProjectionHelper.java
ZakarFin: Actually the other methods in the class return null in this case so this should do the same

ZakarFin:
Closing the frontend PR (oskariorg/oskari-frontend#37) without merging since it needs some kind of UI and other refining. Frontend changes for requesting the export:

Addition for bundles/mapping/mapwfs2/service/Mediator.js

```
/**
* @method setExport
* @param {String} format
*
* sends message to /service/wfs/setExport
*/
setExport: function (layerId, format) {
    this.sendMessage('/service/wfs/setExport', {
        'format': format || 'GEOJSON',
        'layerId': layerId
    });
}
```
Response readable from '/wfs/export' channel with "fileId" in the data. 

The server code will be merged once we have the time to do it properly. Thanks!

ZakarFin:
The more recent GeoTools have removed GDAL binding support so we'll have to find another way of doing this.

PR #45, opened 2017-04-24
Always proxy GetLegendGraphic requests.

tsallinen:
Changed WMS service handling so that GetLegendGraphic request are always proxied. This is to avoid making non secure HTTP requests in HTTPS enabled Oskari installations.

This is a additional fix for https://github.com/oskariorg/oskari-docs/issues/19

What is still missing from complete solution is proxying at WMTS legend image URLs (and maybe ArcGIS REST too). According to the spec, these can be arbitrary URL:s so different kind of logic is required.


ZakarFin:
Thanks!

PR #50, opened 2017-05-09
Feature/print

jampukka:
Add new printout functionality


service-print/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+            <groupId>fi.nls.oskari.service</groupId>
+            <artifactId>oskari-map</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>fi.nls.oskari</groupId>
+            <artifactId>service-logging</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>fi.nls.oskari</groupId>
+            <artifactId>shared-test-resources</artifactId>
```
ZakarFin: Should have <scope>test</scope>

PR #51, opened 2017-05-10
Feature/csw date parsing

mtyni:
csw-date-parsing: Tries to parse DateWithType date in yyyy-mm-dd format. If it does not succeed, date is taken directly from xml and put to xmlDate Json variable. (First version of unit test included)


mtyni:
csw-date-parsing: Removed xmlDate field from Json. Now adds xmlDate to date if date is not in correct format. Removed ISODate class and Joda time package from pom.

ZakarFin:
Thanks!

PR #52, opened 2017-05-12
fix/quality-metadata

mtyni:
fix/quality-metadata: fixed parsing thematic classification correctness and quality topological consistency.


ZakarFin:
Thanks! This fixes an issue where data quality fields were not parsed correctly from CSW response.

PR #54, opened 2017-05-17
Initial Openlayers 3 geoportal view

ZakarFin:
Adds (needs config to opt-in, isn't added for every install) a new OpenLayers 3 based geoportal view with bundles that are already support for OpenLayers 3 and the rest as is (with possible OL2 implementations). This is can be used as a starting point for implementing OL3 versions for the rest of the geoportal components. Enable by adding a property to oskari-ext.properties:

     flyway.sample.1_0_11.skip=false

See MigrationGuide.md for details.

### DIFF
--- a/MigrationGuide.md
+++ b/MigrationGuide.md
@@ -14,6 +14,26 @@ If you don't want it to be added to the sample you can add this property to oska
 
     flyway.sample.1_0_10.skip=true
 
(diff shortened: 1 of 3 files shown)


ZakarFin:
Note! the view itself is not working properly as is, but it's a starting point for OpenLayers 3 based geoportal view implementation

Issue #30, opened 2016-01-29
User layers not deleted, when user account is deleted.


an9l:
When admin deletes the user account, the corresponding user layers data, which are added by that user, are not deleted from database. 
https://github.com/nls-oskari/oskari-server/blob/master/control-admin/src/main/java/fi/nls/oskari/control/admin/UsersHandler.java

And also same might be the case with analysis and myplace data, but not tested.

ZakarFin:
I think this is a bit larger issue. We should have some mechanic for figuring out services that handle user-generated content. That way we don't need to hardcode references to these services on the user handling.

ZakarFin:
Fixed in version 1.42.1

PR #57, opened 2017-05-22
Feature/registration improvements

mtyni:
feature/registration-improvements:
-User account related email templates, subjects (en/fi/sv) and link expiry time are read from properties
-Emails are read from jetty_home/resources (en/fi/sv)
-If email properties are not set, emails are read from server application folders and titles are read from locales
-Error messages shown to user if can't send email, email is not configured or email is in use
-Checking user name is now case-insensitive
-Link to set password is now in html format
-Register selection removed from edit user information
-Removed placeholders (default texts) from change password
-Changed some locales and removed unused locales
-Deletion of user option provided to user
-Deleting all user content, also from view, user layer, my places, user indicators and analysis services with separate service specific Oskari-annotated classes with UserContentService-interface.
-Removed unused view service deletions.
-Added some deletions by uuid to services.
-Account link opens My Data / My Account (contains also frontend changes)
-Link to edit profile to My Account
-Removed edit success notification when opening edit user

Requires changes to properties files and adding account related emails!


control-users/src/main/java/fi/nls/oskari/control/users/RegistrationUtil.java
```diff
@@ -25,7 +26,8 @@ public static Timestamp createExpiryTime(){
         Calendar calender = Calendar.getInstance();
         Timestamp currentTime = new java.sql.Timestamp(calender.getTime().getTime());
         calender.setTime(currentTime);
-        calender.add(Calendar.DAY_OF_MONTH, 2);
+        Integer expiryTimeProperty = Integer.parseInt(PropertyUtil.get("oskari.email.link.expirytime"));
```
ZakarFin: Use int expireDays =  PropertyUtil.getOptional("oskari.email.link.expirytime", 2) instead.

control-users/src/main/java/fi/nls/oskari/control/users/service/MailSenderService.java
```diff
@@ -61,33 +64,64 @@ public final void sendEmail(EmailMessage emailMessage, final String uuid, final
...
         emailMessage.setContent(content);
-        sendEmail(emailMessage, uuid, serverAddress);
+        sendEmail(emailMessage, user.getUuid(), serverAddress, language);
+    }
+    private String readFile(String file) throws ServiceException {
+        InputStream in = null;
+        String contents = null;
+        try {
+            in = PropertyUtil.class.getResourceAsStream("/"+file);
+            BufferedInputStream bis = new BufferedInputStream(in);
```
ZakarFin: Use String contents =  IOHelper. readString(in); instead

service-myplaces/src/main/java/fi/nls/oskari/myplaces/MyPlaceMapper.java
```diff
@@ -21,4 +22,6 @@
     List<MyPlaceCategory> findByIds(@Param("list") List<Long> idList);
     List<MyPlaceCategory> freeFind(Map<String, Object> data);
     List<MyPlaceCategory> findAll();
+    @Delete("delete from categories where uuid #{uid}")
```
ZakarFin: Missing an equal sign on the where statement?

service-users/pom.xml
```diff
@@ -19,6 +19,14 @@
             <groupId>fi.nls.oskari.service</groupId>
             <artifactId>oskari-base</artifactId>
         </dependency>
+		<dependency>
```
ZakarFin: I don't think these are necessary anymore with the annotated UserContentService approach?

control-users/src/main/resources/META-INF/resources/spring-map-jsp/register.jsp
```diff
@@ -298,6 +295,10 @@
                             errorMsg("#email", '<spring:message code="user.registration.error.emailExists"/>');
                         } else if (errorResponse.toLowerCase().indexOf("username") >= 0) {
                             errorMsg("#username", '<spring:message code="user.registration.error.usernameExists"/>');
+                        } else if (errorResponse.toLowerCase().indexOf("address") >= 0) {
+                            errorMsg("#email", '<spring:message code="user.registration.error.sendingFailed"/>');
```
ZakarFin: Add js escaping for all of these:  <spring:message javaScriptEscape="true" .../>

servlet-map/src/main/resources/META-INF/resources/spring-map-jsp/index.jsp
```diff
@@ -121,6 +121,7 @@
             <%-- If logout url is present - so logout link --%>
             <c:when test="${!empty _logout_uri}">
                 <a href="${pageContext.request.contextPath}${_logout_uri}"><spring:message code="logout" text="Logout" /></a>
+                <a href="javascript:void(0);" id="oskari-profile-link"><spring:message code="account" text="Account" /></a>
```
ZakarFin: Maybe use href="${pageContext.request.contextPath}${_registration_uri}" as it works as the user modify page. That way it works properly even if personaldata bundle is not loaded.

ZakarFin: Also add a comment that personaldata is going to look for a link with that id so no-one will change it by mistake

ZakarFin:
This resolves issue #30

ZakarFin:
Also references oskariorg/oskari-docs#29

PR #62, opened 2017-06-06
Feature/sort userlayer properties

okauppinen:
Store user_layer "fields" in JSONArray instead of JSONObject to keep order of the feature properties.

In the user_layer table "fields" json is migrated from JSONObject to JSONArray to keep order of the feature properties.


service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatterUSERLAYER.java
```diff
@@ -54,4 +62,18 @@ private static String getUserLayerTileUrl() {
         }
         return PROPERTY_RENDERING_URL + "&id=";
     }
+
+    // creates JSONArray from JSONArray's JSONObject keys [{key,value},{key2,value2},..] -> [key1, key2,..]
+    private static JSONArray createJSONArrayJSONObjectKeys(final JSONArray json) {
```
jampukka: Would JSONHelper be better place for this?

service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatterUSERLAYER.java
```diff
@@ -54,4 +62,18 @@ private static String getUserLayerTileUrl() {
...
+        try {
+            JSONArray jsarray =  new JSONArray();
+            for(int i = 0; i < json.length(); ++i){
+                JSONObject obj = json.getJSONObject(i);
```
jampukka: Could use JSONArray#get() instead of #getJSONObject() and check that the Object received is an instanceof JSONObject

service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatterUSERLAYER.java
```diff
@@ -54,4 +62,18 @@ private static String getUserLayerTileUrl() {
...
+                jsarray.put(obj.names().get(0));
```
jampukka: To get the first key could call JSONObject#keys() and get the first value from the Iterator (should probably check that atleast one exists)

service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatterUSERLAYER.java
ZakarFin: Normally yes but with this implementation probably not. This assumes that there's always exactly one key for each object in the array.

PR #63, opened 2017-06-07
Feature/add feature count for userlayer import

okauppinen:
The database access library has been updated from Ibatis to Mybatis. UserLayerDbService has been changed to be suitable for new Mybatis implementation. Old iBATIS implementation isn't compatible with new UserLayerDbService.

Layer, style and data inserts are handled in one transaction using mybatis batch executor.

Added error codes (e.g. invalid_file) to response instead of textual messages to support localization in the frontend.

Added feature count to layerJSON response. Also adds a warning object with skipped_features to layerJSON response if feature(s) were skipped (no geometry object or geometry is null) during import.


control-base/src/main/java/fi/nls/oskari/control/data/CreateUserLayerHandler.java
```diff
@@ -76,7 +75,8 @@ public void handleAction(ActionParameters params) throws ActionException {
             File file = unZip(loadItem.getFileitem());
 
             if (file == null) {
-                throw new ActionException("Couldn't find valid import file in zip file");
+                log.error("Couldn't find valid import file in zip file");
+                throw new ActionException ("invalid_file");
```
ZakarFin: Use ActionParamsException so stacktrace isn't logged.

control-base/src/main/java/fi/nls/oskari/control/data/CreateUserLayerHandler.java
```diff
@@ -115,17 +116,22 @@ public void handleAction(ActionParameters params) throws ActionException {
...
+                JSONObject featuresSkipped = new JSONObject();
+                JSONHelper.putValue(featuresSkipped, "featuresSkipped", ulayer.getFeatures_skipped());
+                JSONHelper.putValue(userLayer, "warning", featuresSkipped);
+            }
             response.getWriter().print(userLayer);
 
         }catch (ActionException e) {
-            throw new ActionException("Import failed - " + e.getMessage(),
-                    e);
+            throw new ActionException (e.getMessage());
```
ZakarFin: Maybe just rethrow the original exception?

control-base/src/main/java/fi/nls/oskari/control/data/CreateUserLayerHandler.java
```diff
@@ -115,17 +116,22 @@ public void handleAction(ActionParameters params) throws ActionException {
...
         }
         catch (Exception e) {
-            throw new ActionException("Couldn't get the import file set - " + e.getMessage(),
-                    e);
+            throw new ActionException (e.getMessage());
```
ZakarFin: Any reason why the original exception is not passed?

PR #73, opened 2017-08-09
Feature/spatineo monitor fixes

jampukka:
Rewrite of service-spatineo-monitor module.
This rewrite operates on the Spatineo Monitor API only (Spatineo Serval API support will be moved to another module)


jampukka:
Updated the branch: Remove BackendStatus from service-spatineo-monitor, instead use the already existing counterpart in service-base module. Re-model the BackendStatusService interface, add MyBatis implementation and remove old IBatis implementation. Remove columns and views from Database that are not needed. GetBackendStatusHandler now uses Jackson for JSON serialization.

content-resources/src/main/resources/flyway/oskari/V1_44_0__update_backendstatus_schema.sql
```diff
@@ -0,0 +1,21 @@
...
+-- drop views, handle logic in code not db
+DROP VIEW IF EXISTS portti_backendstatus_allknown;
+DROP VIEW IF EXISTS portti_backendalert;
+
+-- modernise the name of the table
+ALTER TABLE portti_backendstatus RENAME TO oskari_backendstatus;
+
+-- drop columns that are not needed
+ALTER TABLE oskari_backendstatus DROP COLUMN id;
+ALTER TABLE oskari_backendstatus DROP COLUMN source;
```
ZakarFin: Source can be used for identifying which system triggered the status. I would keep it.

service-base/src/main/java/fi/nls/oskari/util/IOHelper.java
```diff
@@ -828,4 +859,47 @@ public static String getParams(Map<String, String> params) {
...
         return urlBuilder.substring(0, urlBuilder.length()-1);
     }
+
+    /**
+     * Very similiar to IOHelper.getParams(Map<String, String> params)
+     * only difference is that this method encodes the keys as well as the values
+     * @param kvps map containing the key value pairs
+     * @return String
+     */
+    public static String formURLEncode(final Map<String, String> kvps) {
```
ZakarFin: Should getParams() implementation just use this one so we would not have a almost identical functions with developers not sure which one to use?

jampukka: Probably, or maybe we could add yet another function formURLEncode(Map, boolean encodeKeys) and this would call it with (kvps, true); and getParams() would call it with (kvps, false);

ZakarFin: Lets remove this and make getParams() encode keys as well. Shouldn't have any problems with that.

content-resources/src/main/resources/flyway/oskari/V1_44_0__update_backendstatus_schema.sql
ZakarFin: On second thought it can be re-added when we actually use it for something so this is ok.

PR #75, opened 2017-08-11
Feature/geolocator countryload

mtyni:
Loads and parses country XML fetched from external geographical locator service. Displays the country name based on user locale and country code from fetched map with search result administrator name.


ZakarFin:
Merged to develop. Will go to master on the next release.

PR #82, opened 2017-09-04
ELF geolocator country parser

mtyni:
ELF geographical locator country parser refactoring and tests. Had some issues with version control, so some formatting changes to print functionality was included.


webapp-map/pom.xml
```diff
@@ -101,6 +101,12 @@
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
         </dependency>
+
+        <dependency>
```
ZakarFin: Remove this dependency

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
```diff
@@ -31,53 +30,118 @@
...
+    }
+
+    public String getAdminNamesForFilter(String country) {
+        List<String> adminNameList = getAdminName(country, true);
+        String adminNames = "";
+        if (adminNameList.size() == 1) {
+            adminNames = adminNameList.get(0);
+        } else {
+            for (int i = 0; i < adminNameList.size()-1; i++) {
+                adminNames = adminNames + adminNameList.get(i) + "<OR>";
```
ZakarFin: This won't work. You should be creating a wfs-filter here.

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorParser.java
```diff
@@ -47,11 +43,12 @@
     public static final String KEY_LOCATIONTYPE_ROLE = "locationType_role";
     public static final String KEY_PARENT_TITLE = "parent_title";
     public static final String KEY_ADMINISTRATOR = "administrator";
-    private static Map<String, String> countryMap = null;
     private Map<String, Double> elfScalesForType = null;
     private Map<String, Integer> elfLocationPriority = null;
     private ELFGeoLocatorSearchChannel channel;
+    private ELFGeoLocatorCountries countries = null;
 
+    private static final String PROPERTY_FORCEXY = "org.geotools.referencing.forceXY";
```
ZakarFin: Seems this is not used. Is there a reason for adding it?

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
```diff
@@ -31,53 +30,118 @@
...
             log.warn("ServiceURL not configured. Add property with key", ELFGeoLocatorSearchChannel.PROPERTY_SERVICE_URL);
         }
+        loadCountryMap();
     }
 
-    public Map<String, String> getCountryMap() throws IOException {
-        // not properly configured
-        if (serviceURL == null) {
-            return countries;
+    public void loadCountryMap () {
```
ZakarFin: Seems like a bad design. Why should this be public if it doesn't return anything?

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
```diff
@@ -31,53 +30,118 @@
...
-        HttpURLConnection conn = channel.getConnection(buf.toString());
+        log.debug("Server request: " + countriesUrl);
+        HttpURLConnection conn = IOHelper.getConnection(countriesUrl);
         String response = IOHelper.readString(conn);
-        log.debug("Server response: " + response);
+        return response;
+    }
+
+    public void parseCountryMap(String response) {
+        if (response == null || response.isEmpty()) {
```
ZakarFin: Should return the parsed result or be made private.

service-search-nls/src/test/java/fi/nls/oskari/search/channel/ELFGeoLocatorSearchChannelTest.java
```diff
@@ -66,6 +58,22 @@ public void testWildcardQueryCreation() throws Exception {
         assertTrue("Should get expected query " + xmlDiff, xmlDiff.similar());
     }
 
+
+    @Test
+    public void testCountrySearch() throws Exception {
+        System.out.println("testCountrySearch");
```
ZakarFin: Remove sout

service-search-nls/src/test/java/fi/nls/oskari/search/channel/ELFGeoLocatorSearchChannelTest.java
```diff
@@ -66,6 +58,22 @@ public void testWildcardQueryCreation() throws Exception {
...
+
+    @Test
+    public void testCountrySearch() throws Exception {
+        System.out.println("testCountrySearch");
+
+        ELFGeoLocatorCountries countries = new ELFGeoLocatorCountries();
+        String response = IOHelper.readString(getClass().getResourceAsStream("ASDIService.xml"));
+        countries.parseCountryMap(response);
+
+        String adminNames = countries.getAdminNamesForFilter("no");
```
ZakarFin: Maybe this should be named getAdminsForCountry and return a set/list?

service-search-nls/src/test/java/fi/nls/oskari/search/channel/ELFGeoLocatorSearchChannelTest.java
```diff
@@ -66,6 +58,22 @@ public void testWildcardQueryCreation() throws Exception {
...
+        assertTrue(adminNames.equals("Norway polar - GN<OR>Norway - GN"));
```
ZakarFin: See comment about wfs-filter

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
mtyni: This refactoring was made to make unit tests possible. Couldn't think of a better solution at the moment.

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
mtyni: This refactoring was made to make unit tests possible. Couldn't think of a better solution at the moment.

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
```diff
@@ -31,53 +30,118 @@
...
                 Node nNode = nList.item(temp);
                 if (nNode.getNodeType() != Node.ELEMENT_NODE) {
                     continue;
                 }
                 Element eElement = (Element) nNode;
-                countries.put(eElement.getElementsByTagName("code").item(0).getTextContent(),
-                        eElement.getElementsByTagName("administrator").item(0).getTextContent());
+                NodeList elements = eElement.getElementsByTagName("administrator");
+                for(int i=0; i<elements.getLength(); i++) {
+                    countryMap.put(eElement.getElementsByTagName("administrator").item(i).getTextContent(),
```
jampukka: Dont call getElementByTagName() inside the loop, you already have the NodeList - call elements.item(i)
Same thing with the next line, grab the value of "code".item(0).getTextContent() before entering the loop.

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
jampukka: This method should return a new Map containing the parsed result. Consider retuning a null or throwing an exception if something fails.

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
```diff
@@ -31,53 +30,118 @@
     private long lastSync;
     private String serviceURL;
     private ELFGeoLocatorSearchChannel channel;
-    private Map<String, String> countries = new HashMap<>();
+    private static Map<String, String> countryMap = new HashMap<>();
```
jampukka: This shouldn't be static

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
```diff
@@ -31,53 +30,118 @@
...
+                log.debug("Could not get countries");
+            }
         }
-        // synced recently
-        if(System.currentTimeMillis() < lastSync + REFRESH_INTERVAL) {
-            return countries;
+        catch (Exception e) {
+            log.debug(e, "Could not get countries");
         }
+        lastSync = System.currentTimeMillis();
```
jampukka: Don't update lastSync if the process fails

mtyni: Isn't this ok as it is, so if it fails it would possibly try to update all the time.

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
mtyni: Is it not good that it refers to the same map in all instances.

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
jampukka: L51 would force it to wait for the REFRESH_INTERVAL even if it fails

service-search-nls/src/main/java/fi/nls/oskari/search/util/ELFGeoLocatorCountries.java
mtyni: How would we set the country map in that case, so we can handle unit tests also.

jampukka: Change the declaration to `public Map<String, String> parseCountryMap(String response);` instead in loadCountryMap() change the reference of the class variable to use the newly parsed Map. In Unit tests validate that the Map looks like what it should

PR #89, opened 2017-09-25
Feature/wcs-client

jampukka:
Groundwork for WCS client library module
Support for creating WCS GetCapabilities, DescribeCoverage, GetCoverage requests
Support for parsing WCS GetCapabilities and DescribeCoverage responses


service-wcs/src/main/java/org/oskari/wcs/capabilities/Capabilities.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public Contents getContents() {
+        return contents;
+    }
+
+    /**
+     * Check if the coverage appears in the GetCapabilities response
+     */
+    public boolean servesCoverage(String coverageId) {
+        if (coverageId == null || coverageId.length() == 0) {
```
ZakarFin: String.isEmpty() is used in other places of Oskari. No biggie - just a heads up.

service-wcs/src/main/java/org/oskari/wcs/capabilities/Capabilities.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+        return contents.getCoverageSummary().stream()
+                .anyMatch(c -> coverageId.equals(c.getCoverageId()));
+    }
+
+    /**
+     * Check if the format appears in the GetCapabilities response
+     */
+    public boolean supportsFormat(String format) {
+        if (format == null || format.length() == 0) {
```
ZakarFin: Same here with length() == 0 -> isEmpty().

service-wcs/src/main/java/org/oskari/wcs/parser/CommonParser.java
```diff
@@ -0,0 +1,103 @@
...
+        String upperCornerStr = XML.getChildText(envelope, "upperCorner").orElseThrow(
+                () -> new IllegalArgumentException("Could not find lowerCorner from Envelope"));
+        double[] upperCorner = parseDoubleArray(upperCornerStr, ' ').orElseThrow(
+                () -> new IllegalArgumentException("Invalid Envelope,"
+                        + " failed to parse doubleList from upperCorner"));
+
+        return new Envelope(srsName, srsDimension, axisLabels, uomLabels, lowerCorner, upperCorner);
+    }
+
+    public static Optional<double[]> parseDoubleArray(String str, char c) {
```
ZakarFin: These parse methods (parseDoubleArray, parseIntArray, parseInt) are very similar to what ConversionHelper already has, but using Optionals. I would suggest that these should be in ConversionHelper, but that would mean a load of dependencies that you don't otherwise need.

Should we add a new utility-module that has no dependencies and could hold classes with these kinds of functions?

service-wcs/src/main/java/org/oskari/wcs/parser/RectifiedGridCoverageParser.java
```diff
@@ -0,0 +1,205 @@
...
+    private static Point parseOrigin(Element gridEnvelopeE, int dimension) {
+        Element originE = XML.getChild(gridEnvelopeE, "origin")
+                .orElseThrow(() -> new IllegalArgumentException(
+                        "Invalid RectifiedGrid, missing origin element"));
+        Element pointE = XML.getChild(originE, "Point").orElseThrow(
+                () -> new IllegalArgumentException("Invalid origin, missing Point element"));
+        return CommonParser.parsePoint(pointE, dimension);
+    }
+
+    private static String[] parseAxisLables(Element gridEnvelopeE, int dimension) {
```
ZakarFin: Typo here parseAxisLables() -> parseAxisLabels()

service-wcs/src/main/java/org/oskari/wcs/request/DescribeCoverage.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Get the HTTP endpoint declared in GetCapabilities for DescribeCoverage
+     * request
+     *
+     * @param wcs capabilities of the service
+     * @param get true if you want the GET endpoint, false if POST
+     * @return the endpoint which might not exist for your binding
+     */
+    public static Optional<String> getDescribeCoverageEndPoint(Capabilities wcs, boolean get) {
```
ZakarFin: Not really sure how I feel about the name "get". How would you feel about "useGet" or having two methods: one for get and another for post?

service-wcs/src/main/java/org/oskari/wcs/parser/CapabilitiesParser.java
```diff
@@ -0,0 +1,197 @@
...
+import org.oskari.wcs.capabilities.ServiceMetadata;
+import org.oskari.wcs.util.WCS;
+import org.oskari.wcs.util.XML;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.xml.sax.SAXException;
+
+public class CapabilitiesParser {
+
+    public static final String ALLOWED_VERSION = "2.0.1";
```
ZakarFin: Should this be used in all the code for WCS here having the "2.0.1" version as string to make it easier to find where version references are?

service-wcs/src/main/java/org/oskari/wcs/request/DescribeCoverage.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param coverageId
+     *            id of the coverage
+     * @return Map with query parameters
+     */
+    public static Map<String, String> toQueryParameters(String coverageId) {
+        return Maps.of(
+                "service", "WCS",
+                "version", "2.0.1",
+                "request", "DescribeCoverage",
```
ZakarFin: Should DescribeCoverage, GetCapabilities etc be declared as an enum and used here (and other similar places)?

service-wcs/src/main/java/org/oskari/wcs/request/GetCoverage.java
```diff
@@ -0,0 +1,250 @@
...
+        TargetAxisExtent[] scaleAxes = ste.axisExtents;
+        int n = scaleAxes.length;
+        String[] values = new String[n];
+        for (int i = 0; i < n; i++) {
+            values[i] = String.format(Locale.US, "%s(%f:%f)", scaleAxes[i].axis, scaleAxes[i].low, scaleAxes[i].high);
+        }
+        return join(values, ',');
+    }
+
+    private static String join(String[] a, char c) {
```
ZakarFin: Might as well be a public util-method in another class?

service-wcs/src/main/java/org/oskari/wcs/util/XML.java
```diff
@@ -0,0 +1,94 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.xml.sax.SAXException;
+
+public class XML {
+
+    public static Document readDocument(InputStream in) throws ParserConfigurationException,
+            SAXException, IOException {
+        DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
```
ZakarFin: Consider configuring dbf with some XEE prevention: 

       dbf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);

https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet

service-wcs/src/main/java/org/oskari/wcs/util/XML.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Optional;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.xml.sax.SAXException;
+
+public class XML {
```
ZakarFin: This class feels like it should be in that no-dependencies util-module that I mentioned in a comment above.

ZakarFin:
Nice work! Check the comments, but looks like this is good for merging.

PR #90, opened 2017-09-25
Support new version of ELF Geolocator ElasticSearch autocomplete API

data-ux:
Structure of ElasticSearch service response has changed and old version is no longer supported.  Version in use now should be version /asdielastic/placenames3/_suggest

New implementation also combines and deduplicates exact & fuzzy matches, sorting the exact matches on top in the response list.

In reference to:
https://jira.nls.fi/browse/AH-3911

### DIFF
--- a/service-search-nls/src/main/java/fi/nls/oskari/search/channel/ELFGeoLocatorSearchChannel.java
+++ b/service-search-nls/src/main/java/fi/nls/oskari/search/channel/ELFGeoLocatorSearchChannel.java
@@ -429,16 +429,22 @@ public class ELFGeoLocatorSearchChannel extends SearchChannel implements SearchA
         if(PROPERTY_AUTOCOMPLETE_URL == null || PROPERTY_AUTOCOMPLETE_URL.isEmpty()) {
             return Collections.emptyList();
         }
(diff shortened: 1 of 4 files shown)


service-search-nls/src/main/java/fi/nls/oskari/search/channel/ELFGeoLocatorSearchChannel.java
```diff
@@ -428,35 +430,50 @@ public void calculateCommonFields(final SearchResultItem item) {
...
-            for (int i = 0; i < jsonHitsArray.length(); ++i ) {
-                String resultName = jsonHitsArray.getJSONObject(i).getJSONObject("_source").getString("name");
-                resultList.add(resultName);
-            }
-            return resultList;
+            jsonObject = new JSONObject(result);
         }
         catch (Exception ex) {
             log.error("Couldn't open or read from connection for search channel!");
             throw new RuntimeException("Couldn't open or read from connection!", ex);
```
ZakarFin: It usually isn't a good idea to log AND throw an exception as it tends to lead to massive amounts of logging for an issue. I'll modify this a bit so we can include it in the next release.

PR #94, opened 2017-09-28
Transport logging improvements

mtyni:
Improvements to transport logging. (Listed in more detail in issue https://jira.nls.fi/browse/AH-4035)


servlet-transport/src/main/java/fi/nls/oskari/pojo/Location.java
```diff
@@ -519,7 +519,7 @@ public ReferencedEnvelope createEnlargedEnvelope(double width, double height) {
     @JsonIgnore
     public ReferencedEnvelope getEnlargedEnvelope() {
         if(this.enlargedEnvelope == null) {
-            log.error("Enlarged envelope not created");
+            log.warn("Enlarged envelope not created");
```
ZakarFin: I think this could be logged as just debug.

ZakarFin:
Note that this includes changes from PR #92 as well. PRs should only include commits for single change.

PR #98, opened 2017-10-05
Moved GetAllRolesHandler from control-examples to control-admin.

tsallinen:


control-admin/src/main/java/fi/nls/oskari/control/admin/GetAllRolesHandler.java
```diff
@@ -33,7 +38,14 @@ public void init() {
     }
 
     @Override
-    public void handleAction(ActionParameters params) throws ActionException {
+    public void preProcess(ActionParameters params) throws ActionException {
+        if (!params.getUser().isAdmin()) {
```
jampukka: Could have used ActionParameters#requireAdminUser();

Also could move the userService == null check from handleGet() to preProcess()

tsallinen: Updated PR as suggested.

jampukka:
Looks good

ZakarFin:
Thanks, control-admin is a more logical place for the functionality.

PR #100, opened 2017-10-10
Add interpolation for layer/metadata extent math

jampukka:
The extent of a layer is described with an envelope in WGS84 coordinates. Currently the envelope is reprojected to the currently used coordinate system by reprojecting the bottom left corner and the upper right corner coordinates and forming another envelope from those. This often creates creates unwanted results especially when dealing with huge extents (for example (-180,45),(180,90)) or when using for example North Pole LAEA projections.

This PR modifies the procedure by generating a polygon from the envelope with (possibly) more than five coordinates (each corner + the first corner again for closing the ring) via linear interpolation. The generated polygon is then transformed to the target coordinate system.

See also:
https://github.com/oskariorg/oskari-docs/issues/4
https://jira.nls.fi/browse/AH-3635
https://jira.nls.fi/browse/AH-4051


ZakarFin:
Ah, need to manually fix the versions in pom as this should be included in 1.44.1

ZakarFin:
New PR for the same code, but more easily mergeable #104

PR #102, opened 2017-10-12
Added DocumentBuilderFactory and TransformerFactory instantiation helpers

tsallinen:
Added new DocumentBuilderFactory and TransformerFactory instantiation methods to XmlHelper.

This was done to centralise security related XML parser configuration and these should be
used in the future to instantiate forementioned factories if possible.

Most of the code was refactored to use these new methods.


jampukka:
Thanks, changes look good. Love that you couldn't resist touching those if() statements :)

ZakarFin:
Awesome, thanks!

PR #105, opened 2017-10-25
Metadata coverage search GeoJSON CRS fix

data-ux:
If GeoJSON doesn't have "crs" property, assume it is EPSG:4326.

### DIFF
--- a/service-csw/src/main/java/fi/nls/oskari/search/channel/MetadataCatalogueQueryHelper.java
+++ b/service-csw/src/main/java/fi/nls/oskari/search/channel/MetadataCatalogueQueryHelper.java
@@ -23,6 +23,8 @@ import org.deegree.ogcwebservices.csw.discovery.Query;
 import org.geotools.feature.FeatureCollection;
 import org.geotools.geojson.feature.FeatureJSON;
 import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.referencing.CRS;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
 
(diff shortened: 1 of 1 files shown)


service-csw/src/main/java/fi/nls/oskari/search/channel/MetadataCatalogueQueryHelper.java
```diff
@@ -293,9 +295,14 @@ private String parseWKTPolygon(final String searchCriterion) {
             fc = fjs.readFeatureCollection(new ByteArrayInputStream(
                     searchCriterion.getBytes("utf-8")));
             ReferencedEnvelope env = fc.getBounds();
+
+            CoordinateReferenceSystem sourceCRS = env.getCoordinateReferenceSystem();
+            if (sourceCRS == null) {
+                sourceCRS = CRS.decode("EPSG:4326", true); // default CRS in GeoJSON
```
jampukka: Since TARGET_SRS is also always "EPSG:4326", no CRS.decode nor ProjectionHelper.transformPoint methods are required. That said I think that doing it this way makes the code more readable and clear.

jampukka:
Looks good to me

service-csw/src/main/java/fi/nls/oskari/search/channel/MetadataCatalogueQueryHelper.java
data-ux: Yes, you're right. But if someone changes TARGET_SRS constant to something else, the above will still work without changes.

PR #106, opened 2017-10-26
DeleteUserIndicator route to write response

ZakarFin:
Deleting a user indicator didn't write any response when the operation is successful. Now writes the deleted indicator as JSON.


jampukka:
Looks good to me. Could've made PARAM_INDICATOR_ID field final and the UserIndicatorService userIndicatorService member not-static in DeleteUserIndicatorHandler.

ZakarFin:
Good points! I replaced the constant with another from ActionConstants class and removed static from service.

jampukka:
Great. Now the only thing that stands out is that the Logger log is static final so I'd name it LOG instead and since the userIndicatorService is no longer static variable I'd swap the order in which the two fields appear in the class :)

jampukka:
Thanks, it looks good to me.

PR #107, opened 2017-11-01
Basic TIFF reader for WCS module

jampukka:
Add basic (simple, far-from-complete yet extensible) support for reading (Geo)TIFF files.
Main reason for not using an existing TIFF library is that most of them don't support arbitrary float32 data (SampleFormat 3) or they don't allow tile based access.


service-wcs/src/main/java/org/oskari/wcs/geotiff/IFD.java
```diff
@@ -0,0 +1,383 @@
...
+        return arr;
+    }
+
+    private static double[] getDoubleArray(ByteBuffer bb, int count) {
+        double[] arr = new double[count];
+        bb.asDoubleBuffer().get(arr);
+        return arr;
+    }
+
+    private static final int getFieldTypeLength(int type) throws IllegalArgumentException {
```
ZakarFin: Some method comment on this would help figure out what's going on.

service-wcs/src/main/java/org/oskari/wcs/geotiff/IFD.java
```diff
@@ -0,0 +1,383 @@
+package org.oskari.wcs.geotiff;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+public class IFD {
```
ZakarFin: Add some comment what this class is used for. Maybe a link to describe Geotiff IFD? I found this one but not sure if there's a better one https://www.geospatialworld.net/article/geotiff-a-standard-image-file-format-for-gis-applications/

service-wcs/src/main/java/org/oskari/wcs/geotiff/IFD.java
jampukka: Added a comment for each of the types

PR #110, opened 2017-11-02
App environment configuration move from mapfull config to "core" functionality

ZakarFin:
Frontend configuration for the application has revolved around the "mapfull" bundle doing much of the stuff that should be part of the Oskari "core". This pull request adds information about the user and xhr endpoint (action routes) url  that has been part of the mapfull bundle config to a generic part of the "appsetup". Also adds information about the uuid and type of the application that's being launched.

There will be a follow-up pull request where the duplicated configuration is removed from mapfull bundles configuration. There's also a related frontend pull request for this change.


ZakarFin:
Related frontend PR is oskariorg/oskari-frontend#165

PR #119, opened 2017-11-23
ELF geolocator result sorting fix

data-ux:
Ensure exact search result is first in list

### DIFF
--- a/service-search-nls/src/main/java/fi/nls/oskari/search/channel/ELFGeoLocatorSearchChannel.java
+++ b/service-search-nls/src/main/java/fi/nls/oskari/search/channel/ELFGeoLocatorSearchChannel.java
@@ -448,12 +448,14 @@ public class ELFGeoLocatorSearchChannel extends SearchChannel implements SearchA
         try {
             JSONArray fuzzyHits = jsonObject.getJSONArray("fuzzy_search").getJSONObject(0).getJSONArray("options");
             for (int i = 0; i < fuzzyHits.length(); i++) {
-                combiner.addHit(fuzzyHits.getJSONObject(i), 0);
+                combiner.addHit(fuzzyHits.getJSONObject(i), false);
             }
 
-            JSONArray exactHits = jsonObject.getJSONArray("normal_search").getJSONObject(0).getJSONArray("options");
-            for (int i = 0; i < exactHits.length(); i++) {
-                combiner.addHit(exactHits.getJSONObject(i), 1000);
+            JSONArray normalHits = jsonObject.getJSONArray("normal_search").getJSONObject(0).getJSONArray("options");
+            for (int i = 0; i < normalHits.length(); i++) {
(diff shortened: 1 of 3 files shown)


service-search-nls/src/main/java/fi/nls/oskari/search/util/ScoredSearchHit.java
```diff
@@ -3,14 +3,20 @@
 class ScoredSearchHit implements Comparable<ScoredSearchHit> {
     public final String text;
     public final int score;
+    public final Boolean isExact;
```
jampukka: isExact could be primitive boolean. You can use Boolean.compare(boolean a, boolean b); for comparisons if that's the reason you chose the Object version.

service-search-nls/src/main/java/fi/nls/oskari/search/util/ScoredSearchHit.java
```diff
@@ -3,14 +3,20 @@
...
         score = s;
+        isExact = e;
     }
 
     @Override
     public int compareTo(ScoredSearchHit other) {
-        return score - other.score;
+        int comparison = this.isExact.compareTo(other.isExact);
+        if (comparison == 0) {
+            return score - other.score;
```
jampukka: Integer.compare(int a, int b); would handle edge cases better, most probably won't be a problem here, but for future reference.

service-search-nls/src/main/java/fi/nls/oskari/search/channel/ELFGeoLocatorSearchChannel.java
```diff
@@ -448,12 +448,14 @@ public void calculateCommonFields(final SearchResultItem item) {
...
+                combiner.addHit(fuzzyHits.getJSONObject(i), false);
             }
 
-            JSONArray exactHits = jsonObject.getJSONArray("normal_search").getJSONObject(0).getJSONArray("options");
-            for (int i = 0; i < exactHits.length(); i++) {
-                combiner.addHit(exactHits.getJSONObject(i), 1000);
+            JSONArray normalHits = jsonObject.getJSONArray("normal_search").getJSONObject(0).getJSONArray("options");
+            for (int i = 0; i < normalHits.length(); i++) {
+                JSONObject hit = normalHits.getJSONObject(i);
+                boolean isExact = searchString.trim().equalsIgnoreCase(hit.getString("text"));
```
jampukka: Could move the trim() outside of the loop, however that might get optimized by the compiler anyway.

data-ux: Yes, youre right, but I left this as is, beacause the performance impact is minimal with the amount of results we have here. I think it's more readable this way.

jampukka:
Looks good to me

PR #127, opened 2017-12-19
Database migration from old statistics functionality to the new one

ZakarFin:
- Migrates the state and config for statsgrid (checks path to only upgrade the old statsgrid, not the new ones) and publishedgrid bundles to the format the new statsgrid uses. 
- Updates the template for statsgrid (in portti_bundle db table) to use the new version
- Changes any publishedgrid bundle in appsetup to statsgrid bundle as the new one is usable in both geoportal and embedded maps.
- Adds divmanazer to any appsetups that have statsgrid, but not divmanazer as it is required by the new statsgrid. Places it after mapfull bundle.

Some of the functionalities are NOT migrated as the new implementation doesn't include all the functionality of the old one. These are written into the state as a "legacy" object leaving the possibility of migrating after the new implementation supports these. NOTE! If a map is edited in the publisher functionality these legacy blocks will be removed. This is ok since if the user modifies the map we don't want to change it again automatically.
- Filtered out regions are not supported -> migrated to the legacy block
- Manual breaks for classification are not supported -> migrated to the legacy block

Requires configuration in oskari-ext.properties based on the database contents:

    # datasource ids added for old indicators
    # [SELECT id FROM oskari_statistical_datasource where config like '%sotkanet.fi/rest%']
    flyway.1_45_4.sotkanet.ds.id=1
    # [SELECT id FROM oskari_statistical_datasource where plugin = 'UserStats']
    flyway.1_45_4.userindicators.ds.id=4

    # layer name mappings for old regionsets (layer name for matching statslayer)
    # [select name from oskari_maplayer where type = 'statslayer']
    flyway.1_45_4.layer.name.kunta=tilastointialueet:kunta4500k_2017
    flyway.1_45_4.layer.name.aluehallintovirasto=tilastointialueet:avi4500k
    flyway.1_45_4.layer.name.maakunta=tilastointialueet:maakunta4500k
    flyway.1_45_4.layer.name.seutukunta=tilastointialueet:seutukunta4500k
    flyway.1_45_4.layer.name.elykeskus=tilastointialueet:ely4500k
    flyway.1_45_4.layer.name.nuts1=dummy:nuts1
    flyway.1_45_4.layer.name.erva=dummy:erva
    flyway.1_45_4.layer.name.sairaanhoitopiiri=dummy:sairaanhoitopiiri

The code uses these configurations as late as possible to make sure you only need to configure things IF they are needed (if there's something to migrate). The migration will fail if configurations are needed, but not provided

### DIFF
--- a/MigrationGuide.md
+++ b/MigrationGuide.md
@@ -25,12 +25,13 @@ The configurations required for oskari-ext.properties are (based on the database
     flyway.1_45_4.layer.name.erva=dummy:erva
     flyway.1_45_4.layer.name.sairaanhoitopiiri=dummy:sairaanhoitopiiri
 
-If your instance wasn't using the thematic maps functionality you don't need to
+If your instance wasn't using the old thematic maps functionality you don't need to
 configure these. If you have the old thematic maps functionality in your application
- you need to add map-layers correspondind the regionsets before running the migration
+ you need to add map-layers corresponding the regionsets before running the migration
  and configure the layer-names in oskari-ext.properties.
 
 To configure the layers see: http://oskari.org/documentation/features/thematicmaps/config
+
 For more information about the migration see: https://github.com/oskariorg/oskari-server/pull/127
 
 ## 1.44.0
--- a/MigrationGuide.md
+++ b/MigrationGuide.md
@@ -1,10 +1,43 @@
 # Migration guide
 
+## 1.45.0
+
+### Thematic maps
+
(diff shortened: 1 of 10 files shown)


jampukka:
Looks good to me. Difficult to judge if it will cover all edge cases, however that might not even be necessary.

ZakarFin:
I'll add some notes to migration guide regarding this and merge it after that.

PR #128, opened 2017-12-20
Statistics data fetching by Sotkanet adapter improved

ZakarFin:
The Sotkanet API currently offers the indicator data as JSON as well as CSV. Previously the data was fetched using CSV and internally adapted to JSON format. After this change the data is fetched as JSON directly which results in less post processing on Oskari which means simpler code and less opportunities to introduce errors.


service-statistics-sotka/src/main/java/fi/nls/oskari/control/statistics/plugins/sotka/requests/IndicatorDataJSON.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public String getRequestSpecificParams() {
+        StringWriter writer = new StringWriter();
```
jampukka: Could use StringBuilder instead of StringWriter. StringWriter internally uses StringBuffer which is basically synchronized version of StringBuilder. StringBuilder also allows function chaining (e.g. `sb.append("foo).append("bar");`) which is nice.

service-statistics-sotka/src/main/java/fi/nls/oskari/control/statistics/plugins/sotka/SotkaIndicatorValuesFetcher.java
```diff
@@ -6,105 +6,131 @@
...
-                // Multiple years can be fetched over several requests.
-                years.add(selector.getValue());
-                break;
-            default:
+            switch (selector.getId()) {
+                case "sex":
+                    gender = selector.getValue();
+                    break;
+                case "year":
+                    // Even though SotkaNET API supports fetching multiple years, we don't do that here.
```
jampukka: What does this comment mean? You are collecting multiple year values and using them in `IndicatorDataJSON.getRequestSpecificParams())`

service-statistics-sotka/src/main/java/fi/nls/oskari/control/statistics/plugins/sotka/SotkaIndicatorValuesFetcher.java
```diff
@@ -6,105 +6,131 @@
...
+        Map<Integer, IndicatorValue> indicatorMap = new HashMap<>();
+        // The response is a String JSON array with JSONObjects
+        JSONArray responseArray = new JSONArray(response);
+        for (int i = 0; i < responseArray.length(); i++) {
+            // Example row: {"indicator" : 4,"region": 231,"year": 2012,"gender": "total","value": 3.4,"absValue": 9}
+            JSONObject valueRow = responseArray.getJSONObject(i);
+            Double doubleValue = valueRow.optDouble("value");
+            if (Double.NaN != doubleValue) {
+                IndicatorValue indicatorValue = new IndicatorValueFloat(doubleValue);
+                Integer id = Integer.valueOf(valueRow.getString("region"));
```
jampukka: If the "region" is always an integer could just call `valueRow.getInt("region");`

service-statistics-sotka/src/main/java/fi/nls/oskari/control/statistics/plugins/sotka/SotkaIndicatorValuesFetcher.java
```diff
@@ -6,105 +6,131 @@
...
+            if (regionId == null) {
+                // sometimes this is null with 0.0 as value.
+                // TODO: check if this is a bug in some other code or an empty row from sotkanet
+                continue;
+            }
+
+            IndicatorValue value = entry.getValue();
+            // SotkaNET gives "Kunta" in some places, "KUNTA" in others...
+            String category = regionParser.getCategoryById(regionId);
+            if (category != null && category.equalsIgnoreCase(regionCategoryId)) {
```
jampukka: Could avoid the null check by swapping order, e.g. `if (regionCategoryId.equalsIgnoreCase(category))`

service-statistics-sotka/src/main/java/fi/nls/oskari/control/statistics/plugins/sotka/SotkaIndicatorValuesFetcher.java
```diff
@@ -6,105 +6,131 @@
...
+     */
+    public Map<String, IndicatorValue> get(StatisticalIndicatorDataModel selectors, String indicator,
+                                           String regionCategoryId) {
+        Map<Integer, IndicatorValue> allValues = fetchData(selectors, indicator);
+        Map<String, IndicatorValue> filteredValues = new HashMap<>();
+        for (Entry<Integer, IndicatorValue> entry : allValues.entrySet()) {
+            Integer regionId = entry.getKey();
+            if (regionId == null) {
+                // sometimes this is null with 0.0 as value.
+                // TODO: check if this is a bug in some other code or an empty row from sotkanet
```
jampukka: Does this comment still hold true? I can't get `Integer.valueOf(valueRow.getString("region"));` (`Integer.valueOf(String)`) to return null.

ZakarFin: I added code to take out any null regions from result of fetchData (and renamed it to fetchDataForAllRegionsets).

PR #86, opened 2017-09-15
Feature/update-capabilities-automatically

jampukka:
This branch adds a new `service-capabilities-update` module.
The module consists of `UpdateCapabilitiesJob` which adds support for scheduled updating of information that is based on the GetCapabilties responses.
Refactor (rewrite) `CapabilitiesCacheService` and `CapabilitiesCacheServiceMybatisImpl`
Refactor (rewrite) `WMTSCapabilitiesParser` and the domain classes.
Add basic validation to the parsed WMTS Capabilities logic (some of the stuff that should throw an exception just logs a warning in this branch)
Remove few unused, mostly commented-out classes within `service-map/src/main/java/fi/nls/oskari/wfs package

PR #130, opened 2018-01-08
Add support for automatically updating capabilities of WMS layers

jampukka:
https://github.com/oskariorg/oskari-server/pull/86 added possibility to automatically update capabilities of WMTS layers. This PR improves the work done in https://github.com/oskariorg/oskari-server/pull/86 by adding support for WMS layers. Pulling the information out of WMS Capabilities documents (for both WMS versions 1.1.1 and 1.3.0) is refactored (rewritten).
Also adds possibility to set different update frequency for each and every layer. The update rate is configured via `oskari_maplayer.capabilities_update_rate_sec` which defaults to 0 (never) making this even more of an opt-in feature.


service-capabilities-update/src/main/java/org/oskari/capabilities/UpdateCapabilitiesJob.java
```diff
@@ -1,278 +1,187 @@
...
+        default:
+            return false;
+        }
     }
 
-    protected void updateLayersGroupAtATime(List<OskariLayer> layers) {
+    protected static boolean shouldUpdate(OskariLayer layer) {
+        Date lastUpdated = layer.getCapabilitiesLastUpdated();
+        if (lastUpdated == null) {
+            LOG.debug("Should update layer:", layer.getId(), "last updated unknown and update rate is positive");
```
ZakarFin: Update rate isn't checked here, but I guess if last update is null we should always update. Maybe change the message a bit?

jampukka: True, the message is a bit misleading, the check is done beforehand in layerService.findAllWithPositiveUpdateRateSec()

PR #132, opened 2018-01-15
Migration for thematic maps user generated indicators

ZakarFin:
## Migrates user generated indicators in database table oskari_user_indicator:
- changes references from Sotkanet "categories"/layer ids to new thematic regionset ids
- changes references from Sotkanet region ids (internal ids) to feature ids (like municipality code) found in regionset layers both in user indicators AND in previously migrated filtered regions
- removes any user indicator that has no data or links to no known region id

## Code refactoring:
- Fixes numerous small issues why user indicators didn't work with the new thematic maps functionality
- moved user indicator related services from service-map/base to service-statistics
- added initial mybatis implementation for database service (does not include inserting or updating user indicators)
- deprecated the ibatis implementation
- thematic maps adapter for EuroStat is no longer included by default in the oskari-server webapp as it's not production ready.

## Database changes:
- adds a constraint/primary key for user indicators and handles any duplicate ids by generating new ids for them (this might break some appsetups that reference these, but it's best to show nothing than a wrong indicator)
- changes the database schema for user indicators to enable multiple years/regionsets for user indicator


content-resources/src/main/java/flyway/oskari/V1_45_16__migrate_legacy_filtered_region_ids.java
```diff
@@ -0,0 +1,128 @@
...
+                    ind.bundle = rs.getInt("bundle_id");
+                    ind.state = rs.getString("state");
+                    list.add(ind);
+                }
+            }
+        }
+        return list;
+    }
+
+    private void populateRegionMapping() throws IOException, JSONException {
```
jampukka: Don't have to fix this, but this method could just return a Map<Integer, String>. Same thing in V1_45_14.

content-resources/src/main/java/flyway/oskari/V1_45_16__migrate_legacy_filtered_region_ids.java
```diff
@@ -0,0 +1,128 @@
...
+        final String json = IOHelper.readString(getClass().getResourceAsStream("NonFlywayResource_1_45_14_regions.json"));
+        JSONArray sotkaRegions = new JSONArray(json);
+        for (int i = 0; i < sotkaRegions.length(); ++i) {
+            JSONObject region = sotkaRegions.getJSONObject(i);
+            sotkaIdToRegionId.put(region.getInt("id"), region.getString("code"));
+        }
+    }
+
+    /**
+     * Returns true if indicator had content/migrated successfully and false if the indicator can be removed
```
jampukka: Nothing is being removed, is the doc out-of-date?

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/DeleteUserIndicatorHandler.java
```diff
@@ -16,23 +16,29 @@
...
-        if(ui == null) {
-            throw new ActionParamsException("Unknown indicator");
+        int id = params.getRequiredParamInt(ActionConstants.PARAM_ID);
+        if (!indicatorService.delete(id, params.getUser().getId())) {
+            throw new ActionParamsException("Unknown indicator/not the owner of the indicator.");
         }
-        if(params.getUser().getId() != ui.getUserId()) {
-            throw new ActionDeniedException("User has no right to delete indicator");
+        LOG.info("Deleting indicator", id);
+        if (!indicatorService.delete(id, params.getUser().getId())) {
```
jampukka: `indicatorService.delete(id, params.getUser().getId())` is being called twice.

ZakarFin: Good catch :+1:

PR #133, opened 2018-01-19
GFI response handling improvements

ZakarFin:
Continues the work done in PR #77.

Allows target attribute and data-urls for images in GFI responses by default (oskari.properties).

The allowed tags for HTML-sanitation and functionality specific sanitation can now be configured in oskari-ext.properties. Without any config the basic allowed tags are described in #77 as "Jsoup relaxed" setting. The configuration can be used to allow more tags/attributes/protocols, but NOT restrict it:

    # allowed tags
    gfi.html.whitelist=additional,tags,as,comma,separated,list
    # allowed attributes for a tag (like html.whitelist.attr.a=target)
    gfi.html.whitelist.attr.[tag]=attributes,for,tag,as,comma,separated,list
    # allowed protocols in an attribute value for a tag (like html.whitelist.attr.a.protocol.href=ftp)
    gfi.html.whitelist.attr.[tag].protocol.[attr]=ftp
    # true if  data urls in images are allowed
    gfi.html.whitelist.attr.img.dataurl=[true or false, defaults to false]

All links will get target="_blank" automatically for GFI responses (if target attribute is allowed for links). All relative links/images are transformed to absolute urls in GFI (/something to https://base.url/something).

Added a generic HTML-sanitation class and a Jsoup whitelist implementation to allow data-urls in images. This can be used in other places, but currently is not:

        String sanitizedHtml =  new org.oskari.util.HtmlDoc("htmlStr")
                .modifyLinks("https://my.base.url.for.relative.urls")
                .getFiltered([optional prefix for config as string]);

The config prefix can be omitted. Properties without any prefix like "html.whitelist" are used in that case. For example the gfi functionality uses prefix "gfi" which results in properties like "gfi.html.whitelist". The call to modifyLinks can also be omitted.


jampukka:
Looks good to me

PR #134, opened 2018-01-19
Myplaces implementation for ol3/myplaces3

okauppinen:
Backend implementation for myplaces3. Added MyPlacesFeatures ja MyPlacesLayers action routes and WFS-T request builders and response handlers/helpers.

Myplaces3 sends feature data in geojson format which is built to WFS-T request. WFS-T response is parsed to geojson and sent back to frontend. fid (myplaces.12345) is parsed to id (12345) in backend.

Default category (myplaces layer) is now created in backend if user doesn't have any (e.g. new user).


jampukka:
Looks good to me, but most of it is written by me. Should wait for someone else to doublecheck it.

service-myplaces/src/main/java/fi/nls/oskari/myplaces/handler/MyPlacesFeaturesHandler.java
```diff
@@ -0,0 +1,209 @@
...
+            } else {
+                LOG.debug("Get MyPlaces by user uuid, uuid:", user.getUuid(),
+                        "crs:", crs);
+                featureCollection = featureService.getFeaturesByUserId(user.getUuid(), crs);
+            }
+
+            ResponseHelper.writeResponse(params, featureCollection);
+        } catch (ServiceException e) {
+            LOG.warn(e);
+            throw new ActionException("Failed to get features");
```
ZakarFin: Wouldn't it make more sense to attach the cause exception to the ActionException? If you want to log the stacktrace with warning and the message as error you should use ActionParamsException to prevent an unnecessary stacktrace caused by ActionException.

service-myplaces/src/main/java/fi/nls/oskari/myplaces/service/wfst/CategoriesWFSTRequestBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+        final int n = ids.length;
+        String[] prefixed = new String[n];
+        for (int i = 0; i < n; i++) {
+            prefixed[i] = prefixId(ids[i]);
+        }
+        return prefixed;
+    }
+
+    public static long removePrefixFromId(String prefixed) {
+        return Long.parseLong(prefixed.substring("categories.".length()));
```
ZakarFin: Maybe check that prefixed.startsWith("categories.") and throw an exception about "unknown id" or something if not the expected prefix?

service-myplaces/src/main/java/fi/nls/oskari/myplaces/service/wfst/MyPlacesFeaturesWFSTRequestBuilder.java
```diff
@@ -0,0 +1,256 @@
...
+        final int n = ids.length;
+        String[] prefixed = new String[n];
+        for (int i = 0; i < n; i++) {
+            prefixed[i] = prefixId(ids[i]);
+        }
+        return prefixed;
+    }
+
+    public static long removePrefixFromId(String prefixed) {
+        return Long.parseLong(prefixed.substring("my_places.".length()));
```
ZakarFin: Same here "my_places." as constant and startsWith check.

ZakarFin:
A couple of minor issues that I commented on the code for exception handling. There are a couple of classes that I feel should be in other modules:

- I think everything inside service-myplaces/src/main/java/org/oskari/wfst are generic enough that they could be moved to service-map or service-wfs to make them easier to find.
- Then again service-base/src/main/java/fi/nls/oskari/domain/map/MyPlace.java and service-base/src/main/java/fi/nls/oskari/domain/map/MyPlaceCategory.java should be in service-myplaces, but they already were in service-base so let's leave them there for now.
- I would move service-base/src/main/java/fi/nls/oskari/util/GeoJSONReader.java, service-base/src/main/java/fi/nls/oskari/util/GML3Writer.java and service-base/src/main/java/fi/nls/oskari/util/GML2Writer.java to service-map OR service-wfs so service-base wouldn't need the jts dependency.

Other than that it's really nice code! Good job!

ZakarFin:
I'll just merge this. We can improve the minor stuff in future PRs.

PR #135, opened 2018-01-22
Transport reset history fix

okauppinen:
Fixes the bug when reset history tool is used with hidden WFS-layers. Reset history tool removes WFS-layers from store and then adds them back. addMapLayer adds layer with visible true default value. So hidden layers were added as visible layers. Now addMapLayer sets also layer's visibility.


jampukka:
Thanks

PR #142, opened 2018-01-29
Separating demo content from core to sample app

ZakarFin:
### Note! All application specific initial content scripts will most probably need to be updated because of this. See https://github.com/oskariorg/oskari-server-extension-template for updated setup instructions.

Initial database setup now always creates (when db doesn't have any Oskari related tables):
- a fixed set of tables (previously some could be omitted which results in a migration nightmare)
- default set of roles: 'User', 'Guest' and 'Admin' (assumed to be present by a number of functionalities - names can be changed later using the admin functionalities)
- default set of bundles (building blocks for applications)
- An "internal" dataprovider which modules like myplaces etc can use when registering baselayers even when there's no app-specific data providers

The "core" Flyway-module ('oskari') is responsible for:
- registering any additional non-app-specific bundles
- adding new database tables/removing old ones/migrating existing ones
- migrating data from a table to another or within a table when needed
- migrating any deprecated bundles to supported replacements

App-specific Flyway-module ('sample' is used by default) is responsible for:
- adding users like admin
- optionally adding credentials for default users to allow users to log in (log in functionality can be customized so credentials are not always needed in the Oskari database)
- adding any layer groups like Inspire themes and dataproviders that are needed by the application
- adding layers  (since all apps use different layersets)
- adding appsetup/view configurations (since mostly all apps are different)


ZakarFin:
I'll update https://github.com/oskariorg/oskari-server-extension-template to reflect these tomorrow.

content-resources/src/main/java/fi/nls/oskari/db/DBHandler.java
```diff
@@ -95,66 +93,93 @@ public static void createContentIfNotCreated() {
...
             datasource = ds;
             Connection conn = ds.getConnection();
             DatabaseMetaData dbmeta = conn.getMetaData();
-
             final String dbName = dbmeta.getDatabaseProductName().replace(' ', '_');
 
-            final ResultSet result = dbmeta.getTables(null, null, "portti_%", null);
-            final ResultSet result2 = dbmeta.getTables(null, null, "PORTTI_%", null);
-            final String propertyDropDB = System.getProperty("oskari.dropdb");
+            if (!hasOskariTables(dbmeta)) {
```
jampukka: Could be more readable if this `if` wasn't negative

content-resources/src/main/java/fi/nls/oskari/db/DBHandler.java
```diff
@@ -95,66 +93,93 @@ public static void createContentIfNotCreated() {
             createContentIfNotCreated(getDataSource());
         }
         catch (Exception ex) {
-            log.error(ex, "Couldn't get connection to db!");
+            getLog().error(ex, "Couldn't get connection to db!");
```
jampukka: Exception handling in this class is all over the place. However those lines aren't under review, perhaps we should leave them be.

jampukka: There's also few instances of System.out.println calls in this class :worried:

jampukka:
Looks good apart from the few comments I made.

content-resources/src/main/java/fi/nls/oskari/db/DBHandler.java
ZakarFin: Yep, it can be used as a standalone jar-file or as part of a webapp to setup the database content which makes error handling and logging a bit trickier and changes would require more testing. Perhaps some day this can have a better overhaul, but for now the main issue is to fix the initial database creation as it's not working at all without changes.

PR #145, opened 2018-02-09
Hierarchical layerlist

kuosman:
This PR register hierarchical-layerlist and admin-hierarchical-layerlist bundles.
Also this add support to order layers/groups and save their orders.
Some namings also included (inspireTheme changed).

Added OskariLayerWorker to get layer BBOX and saved to layer capabilities column. This geom param is used front-end code when zoomed map to layer extent.

This PR cannot be merged without front-end hierarchical-layerlist PR.

#### InspireThemeHandler
- InspireThemes action renamed to MapLayerGroups
- also handler class renamed to MapLayerGroupsHandler

#### inspireTheme key

Changed inpireTheme key to maplayerGroups in GetWmsServicesHandler and SaveLayerHandler.

#### New `GetHierarchicalMapLayerGroups` action

This new action handles hierarchical layerlist layers getting ordered by groups.

#### New `LayerAndGroupOrder` action

This new action handles layers and groups order savings.

#### Removed `GetMapLayersHandler`


control-base/src/main/java/fi/nls/oskari/control/data/LayerAndGroupOrderHandler.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * Handles updating the order and group of the given node i.e. layer or group.
+     * @param params
+     * @throws ActionException
+     */
+    @Override
+    public void handlePost(ActionParameters params) throws ActionException {
+        checkForAdminPermission(params);
```
jampukka: Admin check can be done with `params.requireAdminUser();`

control-base/src/main/java/fi/nls/oskari/control/data/LayerAndGroupOrderHandler.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    
+    /**
+     * Read JSON from request
+     * @param req
+     * @return
+     * @throws ActionException
+     */
+    protected JSONObject getOrderJSON(HttpServletRequest req) throws ActionException {
```
jampukka: This is fine as is, but we should probably add a helper method to `RequestHelper` for reading a JSONObject from the HttpServletRequest, the same way we have one for writing in `ResponseHelper`.

control-base/src/main/java/fi/nls/oskari/control/data/LayerAndGroupOrderHandler.java
```diff
@@ -0,0 +1,172 @@
...
+                        }
+					}
+				}
+			}
+
+
+
+        } catch (JSONException e) {
+        	log.warn(e);
+            throw new ActionException("Failed to read request!");
```
jampukka: This could be an `ActionParamsException` instead since the JSONException is thrown only if the JSON read from the request doesn't follow the required schema. The current exception message is also a bit misleading.

control-base/src/main/java/fi/nls/oskari/control/data/LayerAndGroupOrderHandler.java
```diff
@@ -0,0 +1,172 @@
...
+    private static final String KEY_OLD_PARENT = "oldParent";
+    private static final String TYPE_LAYER = "layer";
+	
+	private OskariMapLayerGroupService oskariMapLayerGroupService;
+	private OskariLayerService oskariLayerService;
+	
+	/**
+	 * Variable to keep track of the new order of the groups.
+	 */
+	private HashMap<Integer, Integer> newGroupOrderMap = new HashMap<>();
```
jampukka: Is `newGroupOrderMap` or `newLayerOrderMap` used somewhere? I can't find any usage.

control-base/src/main/java/fi/nls/oskari/control/data/LayerAndGroupOrderHandler.java
```diff
@@ -0,0 +1,172 @@
...
+        log.debug("Updating layer/group order");
+        JSONObject orderJSON = getOrderJSON(params.getRequest());
+
+
+		try {
+			int parentID = orderJSON.getInt(KEY_PARENT);
+			JSONArray orders = orderJSON.getJSONArray(KEY_ORDERS);
+
+			// change main groups orders
+			if(parentID == -1){
```
jampukka: This is a bit difficult to read. Could we separate the blocks in this if-else to separate functions? Like 
```
if (parentID == -1) {
    updateGroupOrder(orders);
} else {
    updateLayerGroups(orders);
}
```

control-base/src/main/java/fi/nls/oskari/control/data/LayerAndGroupOrderHandler.java
```diff
@@ -0,0 +1,172 @@
...
+            log.warn(e);
+            throw new ActionException("Invalid request!");
+        }
+    }
+    /**
+     * Commonly used with
+     * @param params
+     * @throws ActionException
+     */
+    private void checkForAdminPermission(ActionParameters params) throws ActionException {
```
jampukka: As noted previously, this functionality is available via `params.requireAdminUser();`

control-base/src/main/java/fi/nls/oskari/control/data/MapLayerGroupsHandler.java
```diff
@@ -121,11 +134,39 @@ private void checkForAdminPermission(ActionParameters params) throws ActionExcep
         }
     }
 
-    private void populateFromRequest(ActionParameters params, MaplayerGroup theme) throws ActionException {
-        final Map<String, String> parsed = RequestHelper.parsePrefixedParamsMap(params.getRequest(), PARAM_NAME_PREFIX);
-        for(Map.Entry<String, String> entry : parsed.entrySet()) {
-            // TODO: check against supported locales?
-            theme.setName(entry.getKey(), entry.getValue());
+    protected JSONObject getMapLayerGroup(HttpServletRequest req) throws ActionException {
```
jampukka: This is the same function as `LayerAndGroupOrderHandler#getOrderJSON(HttpServletRequest)` only exception is that the name of the function is different. We should add a `public static JSONObject readJsonFromRequestBody(ActionParameters params);` to `RequestHelper`.

control-base/src/main/java/fi/nls/oskari/control/data/MapLayerGroupsHandler.java
```diff
@@ -70,12 +83,12 @@ public void handleGet(ActionParameters params) throws ActionException {
      */
     public void handlePut(ActionParameters params) throws ActionException {
         checkForAdminPermission(params);
-        final MaplayerGroup theme = new MaplayerGroup();
-        populateFromRequest(params, theme);
-        final int id = oskariMapLayerGroupService.insert(theme);
+        final MaplayerGroup maplayerGroup = new MaplayerGroup();
+        populateFromRequest(params, maplayerGroup);
```
jampukka: Instead of passing the `maplayerGroup` as a parameter to `populateFromRequest` we could create one inside the function and then return it.

control-base/src/main/java/fi/nls/oskari/control/data/MapLayerGroupsHandler.java
```diff
@@ -121,11 +134,39 @@ private void checkForAdminPermission(ActionParameters params) throws ActionExcep
...
+            return maplayerGroup;
+        } catch (IOException e) {
+            log.warn(e);
+            throw new ActionException("Failed to read request!");
+        } catch (IllegalArgumentException | JSONException e) {
+            log.warn(e);
+            throw new ActionException("Invalid request!");
+        }
+    }
+    private void populateFromRequest(ActionParameters params, MaplayerGroup maplayerGroup) throws ActionException {
```
jampukka: This function doesn't really need the `ActionParameters params` or `MaplayerGroup maplayerGroup` as parameters. Instead of `ActionParameters params` we only need the  `JSONObject mapLayerGroup`. And instead getting an empty `MaplayerGroup maplayerGroup` as a parameter we could create one inside this function, modify it and then return it.
This would allow the function to focus more on what it's meant to do (transform/map the `JSONObject` the client sent into a `MaplayerGroup` Java object)

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
```diff
@@ -0,0 +1,117 @@
...
+
+	private static final String KEY_GROUPS = "groups";
+
+    @Override
+    public void handleAction(ActionParameters params) throws ActionException {
+
+        final String lang = params.getHttpParam(PARAM_LANGUAGE, params.getLocale().getLanguage());
+
+        log.debug("Getting layer groups");
+        OskariMapLayerGroupService groupService = new OskariMapLayerGroupServiceIbatisImpl();
```
jampukka: We shouldn't create new instances of any services we need for every request. This should be moved to a private field of this class, and it should be initialized inside the init() function of the ActionHandler, just like you've done in the `LayerAndGroupOrderHandler`.

control-base/src/main/java/fi/nls/oskari/control/data/LayerAndGroupOrderHandler.java
ZakarFin: Also the try/catch block is pretty big and formatting is all over the place :)

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
```diff
@@ -0,0 +1,117 @@
...
+        // Get main groups
+        List<MaplayerGroup> mainGroups = layerGroups.stream()
+                .filter(g -> g.getParentId() == -1)
+                .sorted(Comparator.comparing(MaplayerGroup::getOrderNumber, Comparator.nullsLast(Comparator.naturalOrder())))
+                .collect(Collectors.toList());
+        JSONArray json = new JSONArray();
+        try{
+            for(MaplayerGroup group : mainGroups) {
+                List<Integer> intLayerIds = groupService.findMaplayersByGroup(group.getId());
+                List<String> strLayerIds = new ArrayList<>();
```
jampukka: Perhaps we could add a helper function (could be private function inside this class) that converts a `List<Integer>` to a `List<String>`. This 4-liner is called enough times to justify that, and it makes the main functionality more difficult to read.

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
```diff
@@ -0,0 +1,117 @@
...
+ */
+@OskariActionRoute("GetHierarchicalMapLayerGroups")
+public class GetMapLayerGroupsHandler extends ActionHandler {
+	
+	private static Logger log = LogFactory.getLogger(GetMapLayerGroupsHandler.class);
+
+	private static final String KEY_GROUPS = "groups";
+
+    @Override
+    public void handleAction(ActionParameters params) throws ActionException {
```
jampukka: This function is overly complex, we should split it up into smaller functions.

service-base/src/main/java/fi/nls/oskari/domain/map/MaplayerGroup.java
```diff
@@ -1,13 +1,24 @@
...
+import org.json.JSONArray;
 import org.json.JSONObject;
 
+import java.util.List;
 import java.util.Map;
 
 public class MaplayerGroup extends JSONLocalizedName {
 
 	private int id;
+
```
jampukka: Empty lines here are making it more difficult to spot the fields. Using them for grouping purposes is fine but in this case they aren't helping.

control-base/src/main/java/fi/nls/oskari/control/data/LayerAndGroupOrderHandler.java
ZakarFin: Or something like ActionParams.getPayloadJSON() could be nice.

jampukka: Yea or that. Feel free to add one in this PR :)

control-base/src/main/java/fi/nls/oskari/control/layer/SaveLayerHandler.java
```diff
@@ -270,9 +268,18 @@ private boolean handleRequestToMapLayer(final ActionParameters params, OskariLay
             }
         }
 
-        // FIXME: Update JSON key when frontend implementation was updated
-        MaplayerGroup maplayerGroup = oskariMapLayerGroupService.find(params.getHttpParam("inspireTheme", -1));
-        ml.addGroup(maplayerGroup);
+        String groupId = params.getHttpParam("maplayerGroups", "-1");
```
jampukka: Could consider placing the `"maplayerGroups"` into a static final String PARAM_MAPLAYER_GROUPS. However SaveLayerHandler seems to use "undocumented" keys all over the place. On the other hand that's a poor excuse to not do something "the right way" :)

ZakarFin:
This cannot be merged without any frontend changes since action routes that the current bundles use have been renamed.

service-map/src/main/java/fi/mml/map/mapwindow/service/db/OskariMapLayerGroupServiceIbatisImpl.java
```diff
@@ -134,6 +139,27 @@ public void delete(int id) {
...
         return queryForList(getNameSpace() + ".findMaplayersByGroup", id);
     }
+    
+    /**
+     * Returns the map layer groups which belong to the given parent.
+     * FIXME: Quick and dirty
+     * @param groupId
+     * @return child groups of the given group
+     */
+    public List<MaplayerGroup> findByParentId(final int groupId) {
```
jampukka: All of this logic should probably be written into the SQL statement.

service-map/src/main/java/fi/mml/map/mapwindow/util/OskariLayerWorker.java
```diff
@@ -125,16 +150,35 @@ public static JSONObject getListOfMapLayers(final List<OskariLayer> layers, fina
...
         for (OskariLayer layer : layers) {
+        	
             final String permissionKey = layer.getType()+ "+" + layer.getUrl() + "+" + layer.getName();
             if (layer.getParentId() == -1 && !resources.contains(permissionKey)) {
                 // not permitted if resource NOT found in permissions!
                 // sublayers can pass through since their parentId != -1
                 continue;
             }
             try {
+            	String boundingBoxWKT = layer.getCapabilities().optString("geom");
```
jampukka: These bounding box changes seem a bit out of place inside this PR. There seems to be no notes of changing any of this in the PR comment or in the change log. What's the reasoning for this changes? Are they necessary for the Hierarhical layerlist somehow?

servlet-map/src/main/resources/META-INF/resources/spring-map-jsp/index.jsp
```diff
@@ -6,9 +6,9 @@
 <html>
 <head>
     <title>Oskari - ${viewName}</title>
-    <script type="text/javascript" src="//code.jquery.com/jquery-1.7.2.min.js">
+    <script type="text/javascript" src="//code.jquery.com/jquery-1.10.2.min.js">
```
jampukka: Hope this doesn't break anything.

service-map/src/main/resources/META-INF/OskariLayer.xml
```diff
@@ -616,6 +641,30 @@
...
+                SELECT maplayerid
+                FROM   oskari_maplayer_group_link
+                WHERE  maplayerid=#layerId#
+                       AND groupid=#groupId#
+                LIMIT  1
+                FOR    UPDATE
+        )
+        UPDATE oskari_maplayer_group_link omgl
+        SET    order_number = #orderNumber#
+        FROM   cte
```
jampukka: I'd argue the cte makes this more difficult to read. We could just replace the
```
FROM cte
WHERE oml.maplayerid = cte.maplayerid
```
with
```
WHERE maplayerid = #layerId#
AND groupid = #groupId#
```
right? That way we can also drop the table aliases as well as the whole WITH statement.

service-map/src/main/resources/META-INF/OskariLayer.xml
```diff
@@ -551,16 +567,25 @@
...
     l.capabilities_last_updated,
     l.capabilities_update_rate_sec
 
     FROM
-    oskari_maplayer l LEFT OUTER JOIN oskari_maplayer_metadata m
-    ON  m.metadataid = l.metadataId
+    oskari_maplayer l
+    LEFT OUTER JOIN oskari_maplayer_metadata m ON  m.metadataid = l.metadataId
+    LEFT OUTER JOIN oskari_maplayer_group_link gl ON gl.maplayerid = l.id
+    <dynamic>
```
jampukka: This `<dynamic>` part should be removed. We just got rid of this join in #137. Nothing gets SELECTed from the oskari_maplayer_projections p table so it's easy to see that this isn't needed.

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
jampukka: Couldn't check if this functionality was okay or not, will do after it's split up.

servlet-map/src/main/resources/locale/messages_fi.properties
```diff
@@ -1,61 +1,61 @@
 logout=Kirjaudu ulos
-invalid_password_or_username=Virheellinen käyttäjätunnus tai salasana!!
-username=Käyttäjätunnus
+invalid_password_or_username=Virheellinen k�ytt�j�tunnus tai salasana!!
```
jampukka: There seems to be some line-ending or whitespace differences in this file. @ZakarFin could you check which one's correct

service-map/src/main/java/fi/nls/oskari/map/layer/OskariLayerServiceIbatisImpl.java
```diff
@@ -301,7 +303,71 @@ public OskariLayer find(int id) {
         LOG.debug("Parse layers with positive update rate sec took:", t2-t1, "ms");
         return layers;
     }
+    /**
+     * Returns the map layers which belong to the given parent.
+     * FIXME: Quick and dirty
+     * @param groupId
+     * @return layers of the given group
+     */
+    public List<OskariLayer> findAllByGroupId(final int groupId) {
```
jampukka: This is correctly marked as quick and dirty. :) If it's causing performance issues we should replace it.

servlet-map/src/main/resources/locale/messages_fi.properties
ZakarFin: It's a character encoding change. I wonder if the swedish version has the same problem

ZakarFin:
Implementation for oskariorg/oskari-docs#61

service-map/src/main/java/fi/mml/map/mapwindow/util/OskariLayerWorker.java
kuosman: Yes it is outside this PR. But also one requirement when creating hierarchical layerlist is at layer tools can zoom to layer extent. I add description to PR.

servlet-map/src/main/resources/META-INF/resources/spring-map-jsp/index.jsp
kuosman: Tested with default Oskari setup and all seems to works. In front-end code Jstree library needs newer than 1.7.2 version of jQuery.

control-base/src/main/java/fi/nls/oskari/control/data/LayerAndGroupOrderHandler.java
```diff
@@ -0,0 +1,137 @@
...
+				updateGroupOrder(orders);
+			} else {
+			    updateLayerAndGroupOrders(orders, parentID);
+			}
+        } catch (JSONException e) {
+        	log.warn(e);
+            throw new ActionParamsException("Cannot save orders!");
+        }
+
+        ResponseHelper.writeResponse(params, orderJSON);
```
jampukka: This method returns the payload the client sent us. What's the use case here?

service-control/src/main/java/fi/nls/oskari/control/ActionParameters.java
```diff
@@ -285,4 +292,19 @@ public String getAPIkey() {
...
+     */
+    public JSONObject getPayLoadJSON() {
+        JSONObject payloadJSON = null;
+        HttpServletRequest req = this.getRequest();
+        try (InputStream in = req.getInputStream()) {
+            final byte[] json = IOHelper.readBytes(in);
+            final String jsonString = new String(json, StandardCharsets.UTF_8);
+            return new JSONObject(jsonString);
+        } catch (Exception ignored) {}
+        return payloadJSON;
```
jampukka: Instead of silently swallowing the exception this method should probably throw it so that the caller can decide how he/she wants to handle it.

control-base/src/main/java/fi/nls/oskari/control/layer/SaveLayerHandler.java
```diff
@@ -270,13 +323,21 @@ private boolean handleRequestToMapLayer(final ActionParameters params, OskariLay
             }
         }
 
-        // FIXME: Update JSON key when frontend implementation was updated
-        MaplayerGroup maplayerGroup = oskariMapLayerGroupService.find(params.getHttpParam("inspireTheme", -1));
-        ml.addGroup(maplayerGroup);
-
-        ml.setVersion(params.getHttpParam("version", ""));
+        String groupId = params.getHttpParam(PARAM_MAPLAYER_GROUPS, "-1");
+        ml.emptyMaplayerGroups();
```
ZakarFin: Dataprovider is added as a group before this so you need to empty the array here? You should refactor this: 
 +        final DataProvider group = dataProviderService.find(params.getHttpParam(PARAM_GROUP_ID, -1));
          ml.addGroup(group);

Is dataprovider saved at all?

jampukka: This is confusing because of the naming. `DataProvider group` should be renamed to `DataProvider dataProvider`. In addition, `private Set<DataProvider> groups` inside OskariLayer should be renamed to `private Set<DataProvider> dataProviders` and `OskariLayer#addGroup(final DataProvider group)` should be renamed `OskariLayer#addDataProvider(final DataProvider dataProvider)`

ReleaseNotes.md
```diff
@@ -1,5 +1,27 @@
...
+
+### control-base
+
+#### New `GetHierarchicalMapLayerGroups` action
+
+This new action handles hierarchical layerlist layers getting ordered by groups.
+
+#### New `LayerAndGroupOrder` action
+
+This new action handles layers and groups order savings.
```
ZakarFin: Releasenotes are gathered from the PR descriptions on releases so you don't need to modify this file. You should add this information in the description though.

control-base/src/main/java/fi/nls/oskari/control/data/MapLayerGroupsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+     * @throws ActionException
+     */
+    public void handleDelete(ActionParameters params) throws ActionException {
+        params.requireAdminUser();
+        final int id = params.getRequiredParamInt(PARAM_ID);
+        final MaplayerGroup theme = oskariMapLayerGroupService.find(id);
+        final List<Integer> maplayerIds = oskariMapLayerGroupService.findMaplayersByGroup(id);
+        if(!maplayerIds.isEmpty()) {
+            // theme with maplayers under it can't be removed
+            throw new ActionParamsException("Maplayers linked to theme", JSONHelper.createJSONObject("code", "not_empty"));
```
ZakarFin: I think you should talk about groups and not themes on the comment/exception message.

ZakarFin: Rename the variable also (theme -> group)

control-base/src/main/java/fi/nls/oskari/control/data/MapLayerGroupsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+            while( keys.hasNext() ) {
+                String locale = (String)keys.next();
+                String name = locales.getString(locale);
+                maplayerGroup.setName(locale, name);
+            }
+            maplayerGroup.setParentId(parentId);
+            maplayerGroup.setSelectable(selectable);
+        } catch(JSONException ex) {
+            log.error("Cannot populate maplayergroup from request", ex);
+            throw new ActionException("Cannot populate maplayergroup from request");
```
ZakarFin: Don't log and throw as it leads to excessive logging. So you can remove the log.error here and add the "cause exception" to ActionException.

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Get all map layer groups registered in Oskari database
+ */
+@OskariActionRoute("GetHierarchicalMapLayerGroups")
+public class GetMapLayerGroupsHandler extends ActionHandler {
+	
+	private static Logger log = LogFactory.getLogger(GetMapLayerGroupsHandler.class);
+
+	private static final String KEY_GROUPS = "groups";
+    private OskariMapLayerGroupService oskariMapLayerGroupService;
```
ZakarFin: Some funny indentation here as well.

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
```diff
@@ -0,0 +1,96 @@
...
+                if(depth<=3) {
+                    JSONObject groupJson = group.getAsJSON();
+                    JSONArray subGroupsJSON = getGroupJSON(group.getId(), params, depth);
+                    groupJson.put(KEY_GROUPS, subGroupsJSON);
+                    json.put(groupJson);
+                }
+            }
+        } catch(JSONException ex) {
+            log.error("Group layerlist error", ex);
+            throw new ActionException("Cannot get groupped layerlist");
```
ZakarFin: Same here: don't log and throw. Only throw and add the cause.

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
```diff
@@ -0,0 +1,96 @@
...
+        try{
+            // Loop groups and their subgroups (max depth is 3)
+            for(MaplayerGroup group : layerGroups) {
+                List<String> strLayerIds = ConversionHelper.getStringListFromIntegers(oskariMapLayerGroupService.findMaplayersByGroup(group.getId()));
+
+                final JSONObject layers = OskariLayerWorker.getListOfMapLayersById(strLayerIds, params.getUser(), lang, params.getHttpParam(PARAM_SRS));
+                JSONArray layerList = layers.optJSONArray(OskariLayerWorker.KEY_LAYERS);
+                group.setLayers(layerList);
+
+                if(depth<=3) {
```
ZakarFin: I don't think this is the proper place to handle depth. Wouldn't it make more sense to restrict the depth when saving a group structure?

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
```diff
@@ -0,0 +1,96 @@
...
+        } catch(JSONException ex) {
+            log.error("Group layerlist error", ex);
+            throw new ActionException("Cannot get groupped layerlist");
+        }
+        return json;
+    }
+
+    @Override
+    public void handleAction(ActionParameters params) throws ActionException {
+
```
ZakarFin: This file would read better if you have handleAction before getGroupJSON() since handleAction is the access point that uses getGroupJSON().

control-base/src/main/java/fi/nls/oskari/control/layer/SaveLayerHandler.java
```diff
@@ -193,15 +248,13 @@ private SaveResult saveLayer(final ActionParameters params) throws ActionExcepti
...
 
                 addPermissionsForRoles(ml,
-                        getPermissionSet(params.getHttpParam("viewPermissions")),
-                        getPermissionSet(params.getHttpParam("publishPermissions")),
-                        getPermissionSet(params.getHttpParam("downloadPermissions")),
-                        getPermissionSet(params.getHttpParam("enbeddedPermissions")));
+                        getPermissionSet(params.getHttpParam(PARAM_VIEW_PERMISSIONS)),
+                        getPermissionSet(params.getHttpParam(PARAM_PUBLISH_PERMISSIONS)),
+                        getPermissionSet(params.getHttpParam(PARAM_DOWNLOAD_PERMISSIONS)),
+                        getPermissionSet(params.getHttpParam(PARAM_ENBEDDED_PERMISSIONS)));
```
ZakarFin: There's a typo there and now it's part of the contant :)

control-base/src/main/java/fi/nls/oskari/control/layer/SaveLayerHandler.java
ZakarFin: Please do this renaming ^

control-base/src/main/java/fi/nls/oskari/control/layer/SaveLayerHandler.java
```diff
@@ -270,13 +323,21 @@ private boolean handleRequestToMapLayer(final ActionParameters params, OskariLay
...
+        if(groupId.contains(",")) {
+            String[] groupIds = groupId.split(",");
+            for (String groupid: groupIds) {
+                MaplayerGroup maplayerGroup = oskariMapLayerGroupService.find(ConversionHelper.getInt(groupid, -1));
+                ml.addGroup(maplayerGroup);
+            }
+        } else {
+            MaplayerGroup maplayerGroup = oskariMapLayerGroupService.find(ConversionHelper.getInt(groupId, -1));
+            ml.addGroup(maplayerGroup);
+        }
```
ZakarFin: You can remove the if-statement and just use split for any input since it returns an array of one element if comma isn't present. So instead of 10 lines you can just do:

    for (String id: groupId.split(",")) {
        MaplayerGroup maplayerGroup = oskariMapLayerGroupService.find(ConversionHelper.getInt(id, -1));
        ml.addGroup(maplayerGroup);
    }

service-base/src/main/java/fi/nls/oskari/domain/map/MaplayerGroup.java
```diff
@@ -28,7 +48,26 @@ public JSONObject getAsJSON() {
             JSONHelper.putValue(names, localization.getKey(), localization.getValue());
         }
         JSONHelper.putValue(me, "name", names);
-
+        JSONHelper.putValue(me, "layers", this.getLayers());
+		JSONHelper.putValue(me, "selectable", this.isSelectable());
+		JSONHelper.putValue(me, "parentId", this.getParentId());
+		JSONHelper.putValue(me, "orderNumber", this.getOrderNumber());
```
ZakarFin: Fix intendation.

service-base/src/main/java/fi/nls/oskari/domain/map/MaplayerGroup.java
```diff
@@ -17,6 +23,20 @@ public void setId(int id) {
 		this.id = id;
 	}
 
+	public JSONArray getLayers() {
+		return layers;
+	}
+
+	public void setLayers(JSONArray layers) {
+		this.layers = layers;
```
ZakarFin: Add a comment what is expected to be in the array here

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    private JSONArray getGroupJSON(int parentId, ActionParameters params, int depth) throws ActionException {
+        final String lang = params.getHttpParam(PARAM_LANGUAGE, params.getLocale().getLanguage());
+        List<MaplayerGroup> layerGroups = oskariMapLayerGroupService.findByParentId(parentId);
+        JSONArray json = new JSONArray();
+        depth++;
+        try{
+            // Loop groups and their subgroups (max depth is 3)
+            for(MaplayerGroup group : layerGroups) {
+                List<String> strLayerIds = ConversionHelper.getStringListFromIntegers(oskariMapLayerGroupService.findMaplayersByGroup(group.getId()));
```
ZakarFin: A better solution would be to implement a method that accepts integer list in OskariLayerWorker and OskariLayerService than converting ints to string just to have OskariLayerService convert them back to integers.

service-map/src/main/java/fi/nls/oskari/map/layer/OskariLayerServiceIbatisImpl.java
```diff
@@ -240,6 +242,22 @@ public OskariLayer find(final String idStr) {
 
     }
 
+    public List<OskariLayer> findByIdList(final List<Integer> intList) {
+        final List<String> strList =  ConversionHelper.getStringListFromIntegers(intList);
```
ZakarFin: This is not needed as params are integers. You can also remove the method from ConversionHelper.

ZakarFin:
Tried running this with paikkatietoikkuna.fi db and loading layers fails on:
```
2018-03-02 16:52:34,030 DEBUG fi.mml.map.mapwindow.util.OskariLayerWorker.getListOfMapLayers(OskariLayerWorker.java:216) - Returning 17 / 17 layers
java.lang.NumberFormatException: For input string: "base_36"
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.lang.Integer.parseInt(Integer.java:580)
        at java.lang.Integer.parseInt(Integer.java:615)
        at fi.mml.map.mapwindow.util.OskariLayerWorker.reorderLayersByIntList(OskariLayerWorker.java:499)
```

ZakarFin:
Layers that have collection=true and/or layers that have parentId other than -1 in the database don't work correctly now. Those "sublayers" are listed as normal layers instead of gathered as the collection layers sublayers. In the user interface this looks like there's layers without any groups as sublayers don't have any groups. In such case only the parent layer should be shown to the user as it has the information about groups etc.

service-map/src/main/resources/META-INF/OskariLayer.xml
```diff
@@ -509,25 +509,28 @@
...
                 </iterate>
             </isNotEmpty>
+            <isNotEmpty property="parentIntList" prepend="OR">
+                l.parentId in
+                <iterate property="parentIntList" open="(" close=")" conjunction=",">
+                    #parentIntList[]#
+                </iterate>
+            </isNotEmpty>
         </dynamic>
-
```
ZakarFin: This needs   ORDER BY l.parentId ASC; at the end. Otherwise sublayers are not mapped correctly. The code assumes "main level" layers are in the list first and sublayers to them after all the other layers have been parsed: https://github.com/dimenteq/oskari-server/blob/hierarchical-layerlist/service-map/src/main/java/fi/nls/oskari/map/layer/OskariLayerServiceIbatisImpl.java#L181

service-map/src/main/java/fi/mml/map/mapwindow/util/OskariLayerWorker.java
```diff
@@ -145,7 +204,103 @@ public static JSONObject getListOfMapLayers(final List<OskariLayer> layers, fina
...
+     *
+     * @param layers      list of layers to render
+     * @param user        User
+     * @param lang        Language
+     * @param isPublished Determines the permission type used for the layers (view/published view)
+     * @param isSecure    true to modify urls for easier proxy forwarding/false to keep as is
+     * @return JSONObject of map layers
+     */
+    public static JSONObject getListOfMapLayers(final List<OskariLayer> layers, final User user,
+                                                final String lang, final boolean isPublished, final boolean isSecure) {
```
ZakarFin: Is this a copy of the method above, but without the crs param? Just call the above method instead and handle null crs instead.

service-map/src/main/java/fi/mml/map/mapwindow/util/OskariLayerWorker.java
ZakarFin: This doesn't seem to work. There's a transformWKTGeom() method available in this class that already does this thing correctly. You can remove this code and add a call to transformWKTGeom() below.

service-map/src/main/java/fi/mml/map/mapwindow/util/OskariLayerWorker.java
```diff
@@ -125,16 +165,35 @@ public static JSONObject getListOfMapLayers(final List<OskariLayer> layers, fina
...
+                    CoordinateReferenceSystem targetCRS = CRS.decode(strTargetCRS, true);
+                    MathTransform transform = CRS.findMathTransform(sourceCRS, targetCRS, true);
+                    Geometry targetGeometry = JTS.transform(jtsGeometry, transform);
+                    layer.setGeometry(targetGeometry.toText());
+            	}
                 final JSONObject layerJson = FORMATTER.getJSON(layer, lang, isSecure);
                 //log.debug("Generated JSON");
                 if (layerJson != null) {
+                    if(layer.getGeometry() != null && layer.getGeometry().length() > 0) {
+                    	JSONHelper.putValue(layerJson, "geom", layer.getGeometry());
```
ZakarFin: Instead of this if-statement call transformWKTGeom() here. Remove call to transform from GetMapLayersHandler so it's not done twice: https://github.com/oskariorg/oskari-server/blob/develop/control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayersHandler.java#L33-L36

service-map/src/main/java/fi/mml/map/mapwindow/util/OskariLayerWorker.java
ZakarFin: So instead of copy/pasting the implementation call getListOfMapLayers(layers, user, lang, null, isPublished, isSecure) implemented above.

service-map/src/main/java/fi/mml/map/mapwindow/util/OskariLayerWorker.java
ZakarFin: Actually call from GetMapLayersHandler is still required as crs is not passed from there. Do we still need GetMapLayersHandler with this PR? Maybe it can be removed?

control-base/src/main/java/fi/nls/oskari/control/view/modifier/bundle/MapfullHandler.java
```diff
@@ -109,7 +109,8 @@ public boolean modifyBundle(final ModifierParams params) throws ModifierExceptio
                 params.getViewType(),
                 bundleIds,
                 useDirectURLForMyplaces,
-                params.isModifyURLs());
+                params.isModifyURLs(),
+                mapSRS);
 
 
         // transform WKT for layers now that we know SRS
```
ZakarFin: Now that getFullLayerConfig() already adds the coverage information and transforms it the loop below will do it again and accumulate a massive amount of JSON:
transforms_gone_wrong

ZakarFin: I'll fix this ^

PR #146, opened 2018-02-12
Available appsetups for multiprojection support

ZakarFin:
Adds listing of the default views in the Oskari instance to the 'GetAppSetup' action route response as part of the information under env-key. The information was previously made available as a new action route 'GetSystemDefaultViews', but the information is so crucial for the multiprojection support that it should be always available.


control-base/src/main/java/fi/nls/oskari/util/EnvHelper.java
```diff
@@ -31,13 +37,35 @@
...
 
     public static final String SVG_MARKERS_JSON = "svg-markers.json";
     public static final String PROPERTY_AJAXURL = "oskari.ajax.url.prefix";
 
+    public static void setupViews(List<View> views) {
+        DEFAULT_VIEWS.clear();
+        for(View view: views) {
+            try {
+                JSONObject json = new JSONObject();
+                String srsName = view.getBundleByName("mapfull")
```
jampukka: This could be replaced with View#getSrsName() added in #140

jampukka:
Looks good

PR #147, opened 2018-02-12
Get map options for embedded maps from publisher appsetup

ZakarFin:
The template appsetup used for publishing embedded maps is a shared resource between appsetups in an Oskari instance. It includes for example the projection config for the embedded maps. This is problematic for the publish functionality when appsetups can have different projections. Previously the publish functionality needed to be disabled from the appsetups with different projection as the template. With this change the map options like zoom-levels, projection and units (degree/meters) are copied from the appsetup that was used to publish the embedded map instead of from a "hard coded" template. This removes the possibility of misconfiguration and the publish template being "out of sync" from the geoportal view (like different set of zoom levels) and enables geoportal appsetups with different projections share the base template for publishing.


jampukka:
Looks good

PR #150, opened 2018-02-14
Print improvements

jampukka:
Fixes multiple issues with service-print:
- Pages in landscape mode were [height,height] of their portrait counterparts, now correctly [height,width]
- Used OskariLayer.getExternalId() as the id, which lead to NPEs with Layers that didn't specify externalId, fixed by using the id property instead
- Disable scaleLine drawing. Current implementation doesn't work correctly. Needs to be revisited in the future before it can be accepted as working.
- WMTS layers used the zoomLevel we received from the frontend which lead to issues when the zoomLevel in OpenLayers wasn't the same zoomLevel as in the TileMatrixSet of the service. New implementation uses resolution from OpenLayers and scaleDenominator from WMTS Capabilities to determine the correct TileMatrixSet.
- Remove LRUCache, use Oskari's generic `Cache<T>` instead
- Instead of caching the TileMatrixSets from the WMTS Capabilities document, cache the whole Capabilities document. This way we can determine whether the service supports Tile URLs (WMTS/REST) or KVP variant of the spec. Also adds the possibility to request the correct image type (png/jpeg).
- Improve Hystrix usage. Current implementation is still far from perfect. Difficult to determine what to do as fallback, use a transparent image, or just fail the process? Nobody wants to wait forever but nobody wants PDFs that are missing half of the images.
- Fix drawing of WFS layers, they were upside down.


service-print/src/main/java/org/oskari/print/loader/CommandLoadImageFromURL.java
```diff
@@ -6,58 +6,42 @@
...
 
     @Override
     public BufferedImage run() throws Exception {
-        return loadImageFromURL(new URL(uri));
-    }
-
-    protected static BufferedImage loadImageFromURL(URL url) {
-        LOG.debug(url.toString());
-        for (int i = 0; i < RETRY_COUNT; i++) {
+        LOG.info(uri);
```
ZakarFin: Add some message like LOG.info("Loading image from", uri);

service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMS.java
```diff
@@ -43,7 +52,22 @@ public BufferedImage run() throws Exception {
...
                 .toKVP();
-        return CommandLoadImageFromURL.loadImageFromURL(new URL(request));
+        
+        LOG.info(request);
+        URL url = new URL(request);
+        for (int i = 0; i < RETRY_COUNT - 1; i++) {
+            try {
+                return ImageIO.read(url);
+            } catch (IOException e) {
+                LOG.warn(e);
```
ZakarFin: Perhaps this could wait a while after failure before retrying? Not too long since the user is waiting, but maybe ~50-100ms?

ZakarFin: Add some message for the warning and perhaps log only the last retry to prevent spam on the logs.

PR #151, opened 2018-02-28
GetViews add srsName plus refactoring

jampukka:
GetViews now adds srs information for each View in 'srsName' key. If the information is not available the value will be JSON null.
Refactored the class to use Java Streams.

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/view/GetViewsHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/view/GetViewsHandler.java
@@ -55,8 +55,13 @@ public class GetViewsHandler extends ActionHandler {
 
         final String type = params.getHttpParam(ViewTypes.VIEW_TYPE, ViewTypes.USER);
         final List<View> views = viewService.getViewsForUser(userId);
+        for (View view : views) {
+            if (view.getType() == null) {
+                view.setType(ViewTypes.USER);
+            }
+        }
         final List<JSONObject> viewsAsJsonObjects = views.stream()
-                .filter(v -> isTypeCorrect(type, v.getType()))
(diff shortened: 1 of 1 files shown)


control-base/src/main/java/fi/nls/oskari/control/view/GetViewsHandler.java
```diff
@@ -28,88 +31,81 @@
...
+            final JSONObject stateAccu = new JSONObject();
+            for (Bundle bundle : view.getBundles()) {
+                final JSONObject bundleNode = new JSONObject();
+                try {
+                    bundleNode.put(KEY_STATE, new JSONObject(bundle.getState()));
+                    bundleNode.put(KEY_CONFIG, new JSONObject(bundle.getConfig()));
+                    stateAccu.put(bundle.getBundleinstance(), bundleNode);
+                } catch (Exception e) {
+                    log.debug("Status " + bundle.getStartup());
+                    log.debug("Config " + bundle.getConfig());
```
ZakarFin: I would change the logging to something like:

    log.warn("View", view.getId(), "has invalid bundle", bundle.getName())

jampukka: Will check, the functionality was copied from current implementation.

jampukka: Didn't change the `Bundle#getState()` or  `Bundle#getConfig()` methods to return `{}` if null. Instead decided to use the JSONHelper methods. They'll .info() the empty values. Checked out how `Bundle` manages the state and config as both String and JSONObject, was a bit confusing so didn't want to touch that here.

PR #152, opened 2018-03-02
Download basket critical backend fixes

ollijakobsson:
GetFeatureForCropping.java

- Added support for using authenticated cropping layers by using IOHelper basic auth (credentials will be added in ext-properties)

OGCServices.java

- Fixed ogc-prefixes, which now work like a charm

- Added geometryColumnName which can be used to determine a custom name for geometry column of a downloadable dataset in case it is not the default one (SHAPE). As cropGeomColumn is configured in the layer attributes of a cropping layer, geometryColumnName is configured in the layer attributes of a downloadable layer.

SendDownloadDetailsToEmailThread.java

- just formation fixes


download-basket/src/main/java/downloadbasket/helpers/OGCServices.java
```diff
@@ -76,60 +87,69 @@ public static String getFilter(JSONObject downloadDetails, Boolean writeParam)
...
-	
-	public static String getPluginFilter(JSONObject download)
-			throws JSONException {
-		JSONArray identifiers = new JSONArray(download.getString(PARAM_IDENTIFIERS));
+
+	public static String getPluginFilter(JSONObject download) throws JSONException {
+		JSONArray identifiers = new JSONArray(download.getString(KEY_IDENTIFIERS));
 		String xml = "";
 		String croppingNameSpace = PropertyUtil.get("oskari.wfs.cropping.namespace");
+		final OskariLayerService mapLayerService = new OskariLayerServiceIbatisImpl();
```
jampukka: Creating a new OskariLayerService inside this function is wrong. Modify the function so that OskaryLayerService instance gets passed as a parameter.

download-basket/src/main/java/downloadbasket/actions/GetFeatureForCropping.java
```diff
@@ -67,6 +72,9 @@ public void handleAction(final ActionParameters params) throws ActionException {
 			try {
 
 				HttpURLConnection con = IOHelper.getConnection(wmsUrl);
+				if(hasAuthorization()){
```
jampukka: The same username/password combination is used for all layers. Is this intentional?

download-basket/src/main/java/downloadbasket/actions/GetFeatureForCropping.java
```diff
@@ -67,6 +72,9 @@ public void handleAction(final ActionParameters params) throws ActionException {
...
+					IOHelper.setupBasicAuth(con,croppingUsername, croppingPassword);
+				}
```
ZakarFin: You just pass user/pw to IOHelper.getConnection(url, user, pw). It won't setup basic auth if username is null or empty.

download-basket/src/main/java/downloadbasket/actions/GetFeatureForCropping.java
ollijakobsson: Yes, it is intentional. Example case: a GeoServer workspace is restricted via WFS, so all the layers inside the workspace can be accessed with the same authentication credentials.

In ext-properties, the workspace is defined by its namespace: oskari.wfs.cropping.namespace = cropping_data

ZakarFin: You should use oskariLayer.getUser/passwd instead as you are doing a GFI response.

download-basket/src/main/java/downloadbasket/actions/GetFeatureForCropping.java
```diff
@@ -48,6 +50,9 @@
 
 	private OskariLayerService mapLayerService;
 
+	private String croppingUsername;
+	private String croppingPassword;
```
ZakarFin: These are not used, remove the variables.

download-basket/src/main/java/downloadbasket/actions/GetFeatureForCropping.java
```diff
@@ -25,11 +25,13 @@
 import fi.nls.oskari.map.layer.OskariLayerServiceIbatisImpl;
 import fi.nls.oskari.util.ResponseHelper;
 import fi.nls.oskari.util.IOHelper;
+import fi.nls.oskari.util.PropertyUtil;
```
ZakarFin: Not used, remove.

download-basket/src/main/java/downloadbasket/helpers/OGCServices.java
```diff
@@ -76,60 +87,68 @@ public static String getFilter(JSONObject downloadDetails, Boolean writeParam)
...
+
+	public static String getPluginFilter(JSONObject download, OskariLayerService mapLayerService) throws JSONException {
+		JSONArray identifiers = new JSONArray(download.getString(KEY_IDENTIFIERS));
 		String xml = "";
 		String croppingNameSpace = PropertyUtil.get("oskari.wfs.cropping.namespace");
 
 		try {
 			ByteArrayOutputStream baos = new ByteArrayOutputStream();
 
+			OskariLayer oskariLayer = mapLayerService.find(download.getString(KEY_LAYER_ID));
```
ZakarFin: Why not just pass the layer as param if you don't need the service for anything else? It would make this function so much easier to test.

download-basket/src/main/java/downloadbasket/helpers/SendDownloadDetailsToEmailThread.java
```diff
@@ -92,11 +92,10 @@ public SendDownloadDetailsToEmailThread(OskariLayerService mapLayerService, JSON
 				}
 
 				if (ldz.isDownloadNormalWay()) {
-					ldz.setGetFeatureInfoRequest(OGCServices.getFilter(download, true));
+					ldz.setGetFeatureInfoRequest(OGCServices.getFilter(download, true, mapLayerService));
```
ZakarFin: You are loading the layer referenced by download.getString("id") in here. You are also passing mapLayerService "down the line" for no other reason than to load the same layer again a few methods later. You should remove the maplayerService parameter and just send the layer you already have available in this method (OskariLayer oskariLayer). This will make the methods cleaner and easier to test.

download-basket/src/main/java/downloadbasket/helpers/SendDownloadDetailsToEmailThread.java
```diff
@@ -46,18 +46,16 @@
...
-	 * @param downLoadDetails
-	 *            download details
-	 * @param Locale
-	 *            locale
-	 * 
+	 * @param locale locale
 	 */
 
-	public SendDownloadDetailsToEmailThread(OskariLayerService mapLayerService, JSONArray downLoadDetails,
+	public SendDownloadDetailsToEmailThread(OskariLayerService mapLayerService, JSONArray downloadDetails,
```
ZakarFin: You should just pass the OskariLayer here instead of the service.

PR #160, opened 2018-03-13
Refactoring RegionSet and the supporting classes

jampukka:
Move code around, add empty lines where they were missing, rename fields.
_Only_ thing that _could_ cause issues is that `RegionSet#getUrl()` no longer defaults to a hardcoded value.

### DIFF
--- a/service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/db/RegionSet.java
+++ b/service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/db/RegionSet.java
@@ -20,36 +20,28 @@ import org.json.JSONObject;
  */
 public class RegionSet {
 
-    private long oskariLayerId; // oskari_maplayer.id
-    private String oskariLayerName; // oskari_maplayer.name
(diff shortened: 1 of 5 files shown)


service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/db/RegionSet.java
```diff
@@ -94,32 +67,51 @@ public String getAttributes() {
...
+            if (stats == null) {
+                stats = new JSONObject();
+            }
+        }
+        return stats;
     }
 
-    public void setSrs(String srs) {
-        this.srs = srs;
+    public List<Region> getRegions(String requestedSRS) throws IOException, ServiceException {
```
ZakarFin: Not a big fan of this change since now Regionset is no longer a value object, but contains logic for fetching it's regions. But meh... let's roll with this.

jampukka: That's fair. Will re-evaluate this soon.

PR #163, opened 2018-03-14
Cleanup mapOptions from publish templates

ZakarFin:
Publishing functionality now uses the mapOptions from the appsetup where the publish was made.
This enables publishing using different projections and also makes sure the zoom levels etc map
configuration matches the ones on the preview.

Because of the above publish template no longer needs to have mapOptions at all. This migration removes it from all appsetups/views of type PUBLISH (== publish templates)

### DIFF
--- a/content-resources/src/main/java/flyway/oskari/V1_46_7__cleanup_publish_template.java
+++ b/content-resources/src/main/java/flyway/oskari/V1_46_7__cleanup_publish_template.java
@@ -1,0 +1,75 @@
+package flyway.oskari;
+
+import org.flywaydb.core.api.migration.jdbc.JdbcMigration;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Publishing functionality now uses the mapOptions from the appsetup where the publish was made.
+ * This enables publishing using different projections and also makes sure the zoomlayers etc map
+ * configuration matches the ones on the preview.
+ *
+ * Because of the above publish template no longer needs to have mapOptions at all. This migration removes it from all
+ * appsetups/views of type PUBLISH (== publish templates)
+ */
+public class V1_46_7__cleanup_publish_template implements JdbcMigration {
+
+    public void migrate(Connection connection) throws SQLException {
+        for(Bundle bundle: getMapFullConfigsForPublishTemplates(connection)) {
+            try {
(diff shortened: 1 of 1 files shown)


content-resources/src/main/java/flyway/oskari/V1_46_7__cleanup_publish_template.java
```diff
@@ -0,0 +1,75 @@
...
+                b.view = rs.getLong("view_id");
+                b.bundle = rs.getLong("bundle_id");
+                b.config = rs.getString("config");
+                bundles.add(b);
+            }
+        }
+        return bundles;
+    }
+
+    public static Bundle updateBundleConfig(Connection connection, Bundle bundle)
```
jampukka: Why does this claim to return Bundle? Doesn't really matter though :)

jampukka:
Looks good

content-resources/src/main/java/flyway/oskari/V1_46_7__cleanup_publish_template.java
ZakarFin: True true, copypasting ftw :D

content-resources/src/main/java/flyway/oskari/V1_46_7__cleanup_publish_template.java
```diff
@@ -0,0 +1,75 @@
...
+ * Because of the above publish template no longer needs to have mapOptions at all. This migration removes it from all
+ * appsetups/views of type PUBLISH (== publish templates)
+ */
+public class V1_46_7__cleanup_publish_template implements JdbcMigration {
+
+    public void migrate(Connection connection) throws SQLException {
+        for(Bundle bundle: getMapFullConfigsForPublishTemplates(connection)) {
+            try {
+                JSONObject config = new JSONObject(bundle.config);
+                config.remove("mapOptions");
```
jampukka: Since performance is not an issue here this is fine, but for those interested:
This update "phase" could be seriously improved by not updating the Bundles that do not have mapOptions object in there and by doing all updates inside a single prepared statement with JDBC Batch.
For example:
```
List<Bundle> bundlesToUpdate = new ArrayList<>();
for (Bundle b : ...) {
    try {
        JSONObject config = new JSONObject(bundle.confg);
        if (config.remove("mapOptions") != null) {
            bundle.config = config.toString();
            bundlesToUpdate.add(bundle);
        }
    } catch (JSONException ignore) {}
}
updateAll(connection, bundlesToUpdate);
```
Then in `updateAll()` one would make sure `Connection#setAutoCommit(false)` and then call `PreparedStatement#addBatch()` for each `Bundle` followed by a `PreparedStatement#executeBatch()`

PR #165, opened 2018-03-14
Remove org.json package from service-base

jampukka:
Remove org.json package from service-base, add it as a maven dependency instead.


jampukka:
Yea... this isn't not happening. Versions of org.json/json after 20090211 are missing the `JSONObject#sortedKeys()` method and have changed the way `JSONObject#getString(String key)` works so that the value actually has to be a String, whereas before `Object#toString()` was called for the value. At the same time version 20090211 and the ones before that are missing `JSONArray#remove(int index)`. Our version is some sort of hybrid version between versions 20090211 and 20131018, which I can't find in any maven repository.

PR #166, opened 2018-03-15
Fix metadata data quality parsing

okauppinen:
Data quality nodes' free text content was parsed from parent node, which caused the issue where localized contents were also included in data quality results. Now free text is parsed from CharacterString node and if localized content was found then from LocalisedCharacterString.

Lineage statements were parsed only if them were in the same dataqualityinfo node(s) with dataQuality information nodes. Now lineage statements are parsed separately and returned outside of DataQuality object (reports).
Response:
dataQualities: [DataQuality]
lineageStatements:[String]
(Previously: dataQualityObject.dataQualityNodes:[DataQuality])

Quantitative result's error statistic and value were parsed wrongly.
Removed some unused code and also some refactoring.

### DIFF
--- a/service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
+++ b/service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
@@ -169,7 +169,7 @@ public class CSWIsoRecord {
             arr.put(onlineResource.toJSON());
         }
         JSONHelper.putValue(ret, "onlineResources", arr);
-
+        //TODO: should we create toJSON() instead of using ObjectMapper
         ObjectMapper mapper = new ObjectMapper();
         String jsonInString = "";
         try {
@@ -202,7 +202,7 @@ public class CSWIsoRecord {
 
     public static class DataQualityNode {
         private String nodeName;
-        private String linageStatement;
+        private String lineageStatement;
         private String nameOfMeasure;
         private String measureIdentificationCode;
         private String measureIdentificationAuthorization;
@@ -221,12 +221,12 @@ public class CSWIsoRecord {
             this.nodeName = nodeName;
         }
 
-        public String getLinageStatement() {
-            return linageStatement;
+        public String getLineageStatement() {
+            return lineageStatement;
         }
 
-        public void setLinageStatement(String linageStatement) {
-            this.linageStatement = linageStatement;
+        public void setLineageStatement(String lineageStatement) {
+            this.lineageStatement = lineageStatement;
         }
 
         public String getNameOfMeasure() {
--- a/service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
+++ b/service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
@@ -171,15 +171,22 @@ public class CSWIsoRecord {
         JSONHelper.putValue(ret, "onlineResources", arr);
         //TODO: should we create toJSON() instead of using ObjectMapper
         ObjectMapper mapper = new ObjectMapper();
-        String jsonInString = "";
+        String json = "";
         try {
-            jsonInString = mapper.writeValueAsString(dataQualityObject);
+            arr = new JSONArray();
+            for (DataQuality dqNode: dataQualityObject.getDataQualities()){
+                json = mapper.writeValueAsString(dqNode);
+                arr.put(JSONHelper.createJSONObject(json));
+            }
+        } catch (Exception e) {
+            // TODO?
         }
-        catch (Exception e) {
-            //TODO?
+        JSONHelper.putValue(ret, "dataQualities", arr);
+        arr = new JSONArray();
+        for (String lineage : dataQualityObject.getLineageStatements()){
+            arr.put(lineage);
         }
-        JSONHelper.putValue(ret, "dataQualityObject", JSONHelper.createJSONObject(jsonInString));
-
+        JSONHelper.putValue(ret, "lineageStatements", arr);
         return ret;
     }
 
@@ -187,22 +194,28 @@ public class CSWIsoRecord {
         return referenceSystems;
     }
 
-
     public static class DataQualityObject {
-        private List<DataQualityNode> dataQualityNodes = new ArrayList<>();
+        private List<DataQuality> dataQualities = new ArrayList<>();
+        private List<String> lineageStatements = new ArrayList<>();
 
-        public List<DataQualityNode> getDataQualityNodes() {
-            return dataQualityNodes;
+        public List<DataQuality> getDataQualities() {
+            return dataQualities;
         }
 
-        public void setDataQualityNodes(List<DataQualityNode> dataQualityNodes) {
-            this.dataQualityNodes = dataQualityNodes;
+        public void setDataQualities(List<DataQuality> dataQualities) {
+            this.dataQualities = dataQualities;
+        }
+        public List<String> getLineageStatements() {
+            return lineageStatements;
+        }
+
+        public void setLineageStatements(List<String> lineageStatements) {
+            this.lineageStatements = lineageStatements;
         }
     }
 
-    public static class DataQualityNode {
+    public static class DataQuality {
         private String nodeName;
-        private String lineageStatement;
         private String nameOfMeasure;
         private String measureIdentificationCode;
         private String measureIdentificationAuthorization;
@@ -219,14 +232,6 @@ public class CSWIsoRecord {
 
         public void setNodeName(String nodeName) {
             this.nodeName = nodeName;
-        }
-
-        public String getLineageStatement() {
-            return lineageStatement;
-        }
-
-        public void setLineageStatement(String lineageStatement) {
-            this.lineageStatement = lineageStatement;
         }
 
         public String getNameOfMeasure() {
@@ -668,7 +673,6 @@ public class CSWIsoRecord {
                     this.code = code;
                 }
 
-
                 public JSONObject toJSON() {
                     JSONObject ret = new JSONObject();
                     JSONHelper.putValue(ret, "code", code);
@@ -696,7 +700,9 @@ public class CSWIsoRecord {
             private String dateType;
             private String xmlDate;
 
-            public String getXmlDate() { return xmlDate; }
+            public String getXmlDate() {
+                return xmlDate;
+            }
 
             public void setXmlDate(String xmlDate) {
                 this.xmlDate = xmlDate;
@@ -725,8 +731,7 @@ public class CSWIsoRecord {
                     try {
                         final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd", Locale.US);
                         formattedDate = sdf.format(date);
-                    }
-                    catch (Exception e){
+                    } catch (Exception e){
                         //do nothing
                     }
                 }
--- a/service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
+++ b/service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
@@ -10,7 +10,6 @@ import org.w3c.dom.NodeList;
 import javax.xml.xpath.*;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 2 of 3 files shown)


service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
```diff
@@ -169,40 +169,53 @@ public JSONObject toJSON() {
             arr.put(onlineResource.toJSON());
         }
         JSONHelper.putValue(ret, "onlineResources", arr);
-
+        //TODO: should we create toJSON() instead of using ObjectMapper
```
jampukka: Yes, definitely. Or we could make the object model correspond to the current "JSON model" and use Jackson for serialization. Either way current implementation of mixing the two is horrible. Plus creating a new ObjectMapper is just wrong, it should be re-used.

service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
```diff
@@ -169,40 +169,53 @@ public JSONObject toJSON() {
...
         ObjectMapper mapper = new ObjectMapper();
-        String jsonInString = "";
+        String json = "";
```
jampukka: Why is this outside of the loop?

service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
```diff
@@ -725,8 +731,7 @@ public JSONObject toJSON() {
                     try {
                         final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd", Locale.US);
                         formattedDate = sdf.format(date);
-                    }
-                    catch (Exception e){
+                    } catch (Exception e){
```
jampukka: SimpleDateFormat should be re-used

service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
```diff
@@ -10,16 +10,15 @@
...
 
 public class CSWISORecordDataQualityParser {
 
     private static final Logger log = LogFactory.getLogger(CSWISORecordDataQualityParser.class);
     private static XPathExpression pathToLocalizedValue = null;
 
-    //Linage statement
-    private static XPathExpression XPATH_LINAGE_STATEMENT = null;
+    //Lineage statement
+    private static XPathExpression XPATH_LINEAGE_STATEMENT = null;
```
jampukka: Why are these fields static? They are modified within the constructor. Either make them final and init them in a `static { }` block outside of the constructor or make them non-static final and init them in the constructor.

service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
```diff
@@ -169,40 +169,53 @@ public JSONObject toJSON() {
...
+                arr.put(JSONHelper.createJSONObject(json));
+            }
+        } catch (Exception e) {
+            // TODO?
         }
-        catch (Exception e) {
-            //TODO?
+        JSONHelper.putValue(ret, "dataQualities", arr);
+        arr = new JSONArray();
+        for (String lineage : dataQualityObject.getLineageStatements()){
```
jampukka: Could use `new JSONArray(dataQualityObject.getLineageStatements())` instead

service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
```diff
@@ -90,64 +91,74 @@ public CSWISORecordDataQualityParser() {
...
-                pathToLocalizedValue = xpath.compile(
-                        "../gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString[@locale='#"
-                                + locale.getLanguage().toUpperCase() + "']");
-        }
+    public CSWIsoRecord.DataQualityObject parseDataQualities(final NodeList dataQualityNodes, final XPathExpression pathToLoc)  throws XPathExpressionException {
+        pathToLocalizedValue = pathToLoc;
+
         CSWIsoRecord.DataQualityObject dataQualityObject = new CSWIsoRecord.DataQualityObject();
-        List<CSWIsoRecord.DataQualityNode> dataQualityObjectNodeList = dataQualityObject.getDataQualityNodes();
+        List<CSWIsoRecord.DataQuality> dataQualityList = dataQualityObject.getDataQualities();
```
jampukka: For readability I'd just created a `new ArrayList<T>` instead of calling the `#getDataQualities()`

service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
```diff
@@ -90,64 +91,74 @@ public CSWISORecordDataQualityParser() {
...
+        CSWIsoRecord.DataQuality dataQuality;
```
jampukka: Why are these outside of the loop? Not very Java'ish

service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
```diff
@@ -90,64 +91,74 @@ public CSWISORecordDataQualityParser() {
...
-        dataQualityObject.setDataQualityNodes(dataQualityObjectNodeList);
+        dataQualityObject.setDataQualities(dataQualityList);
         return dataQualityObject;
     }
-
-    private CSWIsoRecord.DataQualityNode GetDataQualityNodeInformation(Node parentNode, String listName)  throws XPathExpressionException {
-        CSWIsoRecord.DataQualityNode dataQualityObjectNode = new CSWIsoRecord.DataQualityNode();
-        dataQualityObjectNode.setNodeName(listName);
-
+    //
```
jampukka: Empty line comment here

service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
```diff
@@ -203,29 +212,33 @@ public CSWISORecordDataQualityParser() {
...
+        Node quantitativeResultErrorStatisticNode = (Node) XPATH_QUANTITATIVE_RESULT_ERROR_STATISTIC.evaluate(parentNode, XPathConstants.NODE);
+        dataQualityObjectQuantitativeResult.setErrorStatistic(localize(quantitativeResultErrorStatisticNode));
 
         return dataQualityObjectQuantitativeResult;
     }
 
     //Move to common utility class
     private String localize(final Node elem) {
-        String ret = getText(elem);
+        String ret = null;
```
jampukka: Could easier to read if instead of going `if ($a != null) {}` we'd just call `if ($a == null) { return null; }`

service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
```diff
@@ -203,29 +212,33 @@ public CSWISORecordDataQualityParser() {
...
-        if (elem != null && pathToLocalizedValue != null) {
+        if (elem != null) {
             try {
-                final Node localeNode = (Node) pathToLocalizedValue.evaluate(elem, XPathConstants.NODE);
-                localized  = getText(localeNode);
-                if (localized != null && !localized.isEmpty()) {
-                    ret = localized;
+                Node contentNode = (Node) XPATH_CHARACTER_STRING.evaluate(elem, XPathConstants.NODE);
+                ret = getText(contentNode);
+                if (pathToLocalizedValue != null){
```
jampukka: The logic here is to use the text content of `../gmd:LI_Lineage/gmd:statement` node as a fallback value the node doesn't have `../gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString[@locale='#foo']` doesn't exist? Is that correct? Obviously if it does have a <gmd:PT_FreeText> element as it's child it can not have a text content, right? Then I'd argue that it would be cleaner to separate the two options more clearly.

okauppinen: The logic is to use non-localized content (CharacterString) if localized content is not found or it's empty. I don't know if it's wanted, but also CSWISORecordParser works similarly. This feature has been noted in the future improvements (should we return empty string or non-localized if localized content isn't found).

         <gmd:statement xsi:type="gmd:PT_FreeText_PropertyType"> 
               <gco:CharacterString>Kuntajako...</gco:CharacterString> 
               <gmd:PT_FreeText> 
                    <gmd:textGroup> 
                         <gmd:LocalisedCharacterString locale="#EN" >Municipal...</gmd:LocalisedCharacterString> 
                     </gmd:textGroup> 
                </gmd:PT_FreeText> 
          </gmd:statement>

service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
```diff
@@ -695,8 +695,11 @@ public JSONObject toJSON() {
             private Date date;
             private String dateType;
             private String xmlDate;
+            final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd", Locale.US);
```
jampukka: This needs to be static so one isn't created for each instance of DateWithType.

jampukka:
Thanks. Could you still add a new Unit test to `CSWISORecordParserTest` that asserts that what is done here works? That would be perfect.

service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
jampukka: If you're not gonna implement this then atleast move the ObjectMapper to a static final field so that it gets atleast some reuse.

service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
jampukka: This is not addressed even though it shows up as outdated.

service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
jampukka: This is not addressed even though it shows up as outdated.

service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
jampukka: This is not addressed even though it shows up as outdated.

jampukka: This is not addressed even though it shows up as outdated.

service-csw/src/main/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParser.java
jampukka: This is not addressed even though it shows up as outdated.

jampukka:
Adding the unit text was good, still waiting for couple of other fixes.

service-csw/src/main/java/fi/nls/oskari/csw/domain/CSWIsoRecord.java
```diff
@@ -695,8 +704,11 @@ public JSONObject toJSON() {
             private Date date;
             private String dateType;
             private String xmlDate;
+            private static final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd", Locale.US);
```
ZakarFin: SimpleDateFormat is not threadsafe, but you could do:

     private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd");

And layer on in setXmlDate():

         date.toInstant().format(DATE_FORMATTER);

PR #167, opened 2018-03-15
Move UserLayer Handlers and Services to separate modules

jampukka:
Add two new modules, control-userlayer and service-userlayer. Move ActionHandlers that handle UserLayer s from control-base to control-userlayer. Move Services that handle UserLayers from service-map to control-userlayer.

If you want to use UserLayers in your application you'll have to add the following dependency to your webapp-map:
```
<dependency>
    <groupId>org.oskari</groupId>
    <artifactId>control-userlayer</artifactId>
</dependency>
```


ZakarFin:
Looks good. Just add something like what's on the PR description to https://github.com/oskariorg/oskari-server/blob/master/MigrationGuide.md under 1.46.0 heading.

ZakarFin:
Example for dependency change on server-extension: nls-oskari/kartta.paikkatietoikkuna.fi#23

PR #168, opened 2018-03-19
Add GeoJSON module

jampukka:
Working with GeoJSON can be troublesome. This module adds support for converting GeoTools `SimpleFeatureCollection`, `SimpleFeature` and JTS `Geometry` to and back from the JSONObject of oskari-base.


oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONReader.java
```diff
@@ -0,0 +1,272 @@
...
+
+        if (builder == null) {
+            builder = getBuilder(json);
+        }
+
+        builder.reset();
+
+        JSONObject geometry = json.optJSONObject(GeoJSON.GEOMETRY);
+        if (geometry != null) {
+            builder.set("geom", toGeometry(geometry));
```
ZakarFin: Add "geom" as a constant as it is repeated multiple times on the code. Maybe even use builder.getDefaultGeometry() instead of "geom" here?

ZakarFin:
Looks good. Just one note about using the "geom" string.

After merging: any app using fi.nls.oskari.util.GeoJSONReader should use org.oskari.geojson.GeoJSONReader and add the dependency:

         <dependency>
             <groupId>org.oskari</groupId>
             <artifactId>oskari-geojson</artifactId>
         </dependency>

oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONReader.java
jampukka: Changed it to a public static final field. Also changed the value from `"geom"` to `"geometry"` since that's what the GeoJSON plugin in GeoTools calls it.

PR #170, opened 2018-03-20
Add support for using GeoJSON resource file as a source for RegionSet

jampukka:
Currently `Regions` of a `RegionSet` are always retrieved via WFS. This PR adds support for using GeoJSON files as the source as well. This will activate for RegionSets that return `RegionSet#getFeaturesUrl()` that starts with `resources://` (instead of `http://`).


service-statistics-common/src/test/java/fi/nls/oskari/control/statistics/RegionSetHelperTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.opengis.referencing.operation.TransformException;
+import org.oskari.geojson.GeoJSON;
+
+import fi.nls.oskari.control.statistics.db.RegionSet;
+import fi.nls.oskari.service.ServiceException;
+
+public class RegionSetHelperTest {
+
+    @Test
+    @Ignore("Requires HTTP connection to a WFS server serving specific FeatureType")
```
ZakarFin: I would have liked to keep the input as a resource file as an example of what the regionset config should be for a given WFS-response. But there was some problems with linked resources using the new parser?

jampukka: The used parser (GeoTools GML) is the same as before, but the it won't work when the schemaLocation links etc. in the document don't work. For example current resource files contain links to `dev.paikkatietoikkuna.fi:80`, which won't resolve. This test is also more of an "integration" test in the sense that it's asserting more the fact that the request generated is valid WFS request and that the WFS service responses with a `<wfs:FeatureCollection>` GML document for the request. Parsing the GML is tested more thoroughly within GeoTools already, so that's not the main thing under inspection here.

PR #171, opened 2018-03-20
Add WGS84 bounds check to WKTHelper.transformLayerCoverage(String wkt, String targetSRS)

jampukka:
`WKTHelper.transformLayerCoverage(String wkt, String targetSRS)` clearly states that it should only be called with a WKT that uses WGS84 coordinates, but it does not really enforce it. Interpolating with a tiny threshhold value of 1.0 (meant to be used only with deegrees) can easily lead to OutOfMemoryExceptions when executed on coordinates that are in a metric projection.

The logic is to check that for all coordinates in the WKT extent:
- the `x` value (first number) must be within [-180.0, 180.0]
- the `y` value (second) must be within [-90.0, 90.0].

### DIFF
--- a/service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
+++ b/service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
@@ -24,7 +24,7 @@ public class WKTHelper {
     private static final Logger log = LogFactory.getLogger(WKTHelper.class);
     public final static CoordinateReferenceSystem CRS_EPSG_4326 = getCRS(PROJ_EPSG_4326);
 
-    private static final double INTEPOLATE_THRESHHOLD = 1.0;
+    private static final double INTEPOLATE_THRESHOLD = 1.0;
     private static final double WGS84_LON_MIN = -180.0;
     private static final double WGS84_LON_MAX=   180.0;
     private static final double WGS84_LAT_MIN =  -90.0;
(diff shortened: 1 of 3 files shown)


service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
```diff
@@ -23,6 +24,12 @@
     private static final Logger log = LogFactory.getLogger(WKTHelper.class);
     public final static CoordinateReferenceSystem CRS_EPSG_4326 = getCRS(PROJ_EPSG_4326);
 
+    private static final double INTEPOLATE_THRESHOLD = 1.0;
```
ZakarFin: Missing an R here as well ;)

service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
```diff
@@ -78,16 +85,29 @@ public static String transform(final String wkt, final String sourceSRS, final S
...
         }
         Polygon polygon = (Polygon) geom;
-        CoordinateSequence cs = GeometryHelper.interpolateLinear(polygon.getExteriorRing(), 1.0, gf);
+        LineString exterior = polygon.getExteriorRing();
+        boolean withinWGS84Bounds = GeometryHelper.isWithin(exterior.getCoordinateSequence(),
+                WGS84_LON_MIN, WGS84_LAT_MIN,
+                WGS84_LON_MAX, WGS84_LAT_MAX);
+        if (!withinWGS84Bounds) {
+            log.info("Layer coverage not within WGS84 bounds, not interpolating or transforming extent");
+            return null;
```
ZakarFin: Returning empty instead of unmodified geometry is better in this case since having the wrong information for the frontend will most likely make the layer be invisible to the user (not shown on the map properly) while having no information will just prevent the user from zooming to the extent (since the extent is not known).

PR #172, opened 2018-03-20
Downgrade jQuery to fix myplaces (and break hierarchical layerselector)

ZakarFin:
jQuery was updated in PR #145 to accommodate a new library jstree used in the new bundle for hierarchical layerselector. Jstree requires the minimum jQuery version of 1.9.0, but the same version does no longer include some functions like jQuery.live() that is used in for example myplaces functionality. Therefore a quick fix for this is to downgrade jQuery while continuing the work on making the upgrade possible.

### DIFF
--- a/servlet-map/src/main/resources/META-INF/resources/spring-map-jsp/index.jsp
+++ b/servlet-map/src/main/resources/META-INF/resources/spring-map-jsp/index.jsp
@@ -6,7 +6,7 @@
 <html>
 <head>
     <title>Oskari - ${viewName}</title>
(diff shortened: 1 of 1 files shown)


ZakarFin:
JStree used in oskariorg/oskari-frontend#263

jampukka:
As documented in https://github.com/oskariorg/oskari-server/pull/145/files#r167166813 
This change seems rather straightforward.

ZakarFin:
No need for downgrade now that myplaces have been upgraded

PR #174, opened 2018-03-22
Add myplaces2 handling for myplaces3

ZakarFin:
The "my places" functionality has some special handling if the layers are used in the geoportal view or a published map. Since the OpenLayers 4 version of my places functionality is added as a new bundle named myplaces3 it requires the same handling as myplaces2 bundle.

For users this fixes an issue where myplaces layers are referenced in a link or in a saved view. Without this handling layers are not listed properly and the frontend console has warnings about duplicate layer ids. The layer that is part of the initial "selected layers" will prevent any further my places layers to be added as it's already registered as a WMS-layer when the frontend tries to add it as a "my places" layer.

This is a quick fix for the issue. A better one would be to rename myplaces3 -> myplaces2 as they both share the same config and other characteristics. The other is an OL4 implementation of the previous functionality.


jampukka:
OK. Clean up all of in the future in the same update that removes OL2 stuff for good.

PR #179, opened 2018-03-28
Rewrite CreateUserLayerHandler

jampukka:
Rewrite CreateUserLayerHandler.
* Smaller functions
* Always clean up temp files we create
* Improved error reporting (far from perfect still)

Rewrite all `*GeoJSONCollection` parser classes to not return GeoJSON `String` but to return `SimpleFeatureCollection` instead.
* Make future changes easier (for example calculate the bounds of the Layer when inserting)
* Use the same code for projection transformation
* Add unit tests for some of the parsers (the ones that do not require GDAL installation).

Use oskari-geojson module for GeoJSON serialization instead of the GeoTools module.
* Remove gt-geojson dependency
* Remove unnecessary back and forth `String` <-> `JSONObject` conversions.


control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
```diff
@@ -1,222 +1,250 @@
...
-    private static final String IMPORT_SHP = ".SHP";
-    private static final String IMPORT_GPX = ".GPX";
-    private static final String IMPORT_MIF = ".MIF";
-    private static final String IMPORT_KML = ".KML";
+    private static final Logger log = LogFactory.getLogger(CreateUserLayerHandler.class);
     private static final String PARAM_EPSG_KEY = "epsg";
     private static final String PARAM_SOURCE_EPSG_KEY = "sourceEpsg";
     private static final String USERLAYER_MAX_FILE_SIZE_MB = "userlayer.max.filesize.mb";
-    private static final String USERLAYER_DEFAULT_SOURCE_EPSG = "userlayer.default.source.epsg";
+    private static final String USERLAYER_DEFAULT_TARGET_EPSG = "userlayer.default.target.epsg";
```
ZakarFin: Change in the properties should be mentioned in MigrationGuide.md

jampukka: Added note to migration guide about properties

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
```diff
@@ -1,222 +1,251 @@
...
-
-            // Only 1st file item is handled
-            RawUpLoadItem loadItem = getZipFiles(params);
-
-            // Checks file size
-
-            File file = unZip(loadItem.getFileitem());
+    private String getFileExt(String name) {
+        int i = name.lastIndexOf('.');
+        if (i < 0 || i > name.length() - 1) {
```
ZakarFin: This could be just i == -1?

jampukka: The first part yes, but the or is there to avoid files named "foo."

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
```diff
@@ -1,222 +1,251 @@
...
+                    break;
+                }
+                if (ze.isDirectory()) {
+                    continue;
+                }
+                String ext = getFileExt(ze.getName());
+                if (ext == null) {
+                    continue;
+                }
+                String name = baseName + '.' + ext;
```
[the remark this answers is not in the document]
jampukka: That's true. Also what we could do is instead is to create a temp directory with the random uuid name instead and unzip the files with their "actual" names to that directory.

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
```diff
@@ -1,222 +1,251 @@
...
+        }
+        return name.substring(i + 1);
+    }
 
-            if (file == null) {
-                log.error("Couldn't find valid import file in zip file");
-                throw new ActionParamsException("invalid_file");
-            }
+    private File writeToTempFile(InputStream in, String name) throws IOException {
+        File file = new File(getTempDir(), name);
```
ZakarFin: Each file will get it's own temp directory? So having files with same filename work instead of throwing an exception about existing file?

service-userlayer/src/main/java/org/oskari/map/userlayer/input/FeatureCollectionParsers.java
```diff
@@ -0,0 +1,86 @@
...
+     * Read Features from FeatureSource to memory (ensure they won't
+     * disappear when DataStore#dispose() is called) while transforming
+     * their geometries from source projection to target projection
+     * @throws Exception lots can go wrong
+     */
+    public static SimpleFeatureCollection read(SimpleFeatureSource src,
+            CoordinateReferenceSystem sourceCRS, CoordinateReferenceSystem targetCRS) throws Exception {
+        DefaultFeatureCollection fc = new DefaultFeatureCollection();
+        SimpleFeatureCollection sfc = src.getFeatures();
+        MathTransform transform = getTransform(sourceCRS, targetCRS);
```
ZakarFin: Does transform need a null check here?

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
jampukka: Nope, `getTempDir()` currently always returns the same dir (systems temp dir). However having one directory for each upload would probably be better. The way this is supposed to work right now is that the basename should be unique enough (the first 24 letters of a random UUID) so that the file names won't collide within the temp directory (and they get deleted rather quickly)

service-userlayer/src/main/java/org/oskari/map/userlayer/input/FeatureCollectionParsers.java
jampukka: If either sourceCRS or targetCRS is null getTransform will return null. The `transform` is scoped to this function, and it's null-ness is checked on line 64.

service-userlayer/src/main/java/org/oskari/map/userlayer/input/FeatureCollectionParsers.java
```diff
@@ -0,0 +1,86 @@
...
+        try (SimpleFeatureIterator it = sfc.features()) {
+            while (it.hasNext()) {
+                SimpleFeature f = it.next();
+                if (transform != null) {
```
ZakarFin: You have the null check here since you want to read the features even if transform isn't done? I think not doing the transform should be handled in some other way than just passing features through. Not sure if not having default geometry is handled somewhere down the line, but that's kind of a deal breaker as well.

service-userlayer/src/main/java/org/oskari/map/userlayer/input/GPXParser.java
```diff
@@ -0,0 +1,59 @@
...
+        DataStore store = null;
+        try {
+            store = factory.createDataStore(connectionParams);
+            for (String typeName : store.getTypeNames()) {
+                // Skip track points
+                if ("track_points".equals(typeName)) {
+                    continue;
+                }
+                // GPX always lon,lat 4326
+                sourceCRS = CRS.decode("EPSG:4326", true);
```
ZakarFin: Do this outside of the loop before the connectionParams.

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
ZakarFin: Or is createTempFile(),getParentFile() the java.io,tmp folder so everything is saved to the tmp root?

service-userlayer/src/main/java/org/oskari/map/userlayer/input/FeatureCollectionParsers.java
jampukka: Having no geometry is currently checked and tracked at the `UserLayerDataService` level. The number of features that get declined because they had no geometry is currently reported back to the client. However I'm not certain how valuable that information truly is.

jampukka: Agree with the transform check, we must always know both the sourceCRS and the targetCRS, otherwise none of this makes any sense.

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
```diff
@@ -235,115 +264,9 @@ public void handleAction(ActionParameters params) throws ActionException {
                 JSONHelper.putValue(userLayer, "warning", featuresSkipped);
             }
             response.getWriter().print(userLayer);
```
ZakarFin: You could just do ResponseHelper.writeResponse(params, userLayer)?

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
jampukka: Yes it is. Let's discuss tomorrow how we want to handle the temp files and I'll make the required changes.

service-userlayer/src/main/java/org/oskari/map/userlayer/service/UserLayerDataService.java
```diff
@@ -44,24 +49,13 @@
...
-
-
-        final UserLayer userLayer = new UserLayer();
-        final UserLayerStyle style = new UserLayerStyle();
-        List <UserLayerData> userLayerDataList = new ArrayList<UserLayerData>();
-
-        log.info("user data store start: ", fparams);
-
-        //TODO: Style insert
+    public UserLayer storeUserData(SimpleFeatureCollection fc, User user, Map<String, String> fparams) throws ServiceException {
```
ZakarFin: I would rather replace fparams with UserLayer and parse the form fields on the ActionHandler level.

ZakarFin: Or create a bunch of parseUserLayer(), parseStyle() methods that you use to get a class out of params and call userLayerService.insertUserLayer(userLayer, style, userLayerDataList) outside of those params to classes methods.

service-userlayer/src/main/java/org/oskari/map/userlayer/service/UserLayerDataService.java
```diff
@@ -70,85 +64,88 @@ public UserLayer storeUserData(GeoJsonWorker gjsWorker, User user, Map<String, S
...
-            for (int i = 0; i < geofeas.length(); i++) {
-                JSONObject geofea = geofeas.optJSONObject(i);
-                if (geofea == null){
-                    continue;
-                }
-
-                if (!geofea.has("geometry") || geofea.optJSONObject("geometry")== null) {
+            while (it.hasNext()) {
+                SimpleFeature f = it.next();
+                if (f.getDefaultGeometry() == null) {
```
ZakarFin: Here's the null defaultGeometry handling that I was requesting earlier. It's fine to handle it here.

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
jampukka: Yea didn't want to touch this beast, no idea if the IE iframe workaround thingy is still valid (seems that it requires "Content-Type: text/plain").

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
ZakarFin: True, I guess that would return an empty string instead of the null we now get on such a filename? Maybe i +1 == name.length() or name.endsWith(".") would be clearer? With "foo." isn't the i=3 and name.length() - 1 = 3 so the check wouldn't work properly now? Should check this in the morning, it's way too late for thinking this through :D

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
ZakarFin: The frontend doesn't use iframe for the upload anymore so the hack shouldn't be needed anymore.

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
jampukka: Yea it the comparison should definitely be equal to not greater than for this to work, hehe.

ZakarFin:
Nice work! It's much cleaner than it was.

service-userlayer/src/main/java/org/oskari/map/userlayer/input/FeatureCollectionParsers.java
jampukka: Changed the way this works. Now if either sourceCRS or targetCRS is null we will throw an exception instead.

PR #181, opened 2018-03-28
Edit userlayer style

okauppinen:
Backend implementation for editing userlayer's name, description, data source and style.

New action routes for getting style and setting updated values.


control-userlayer/src/main/java/org/oskari/control/userlayer/EditUserLayerHandler.java
```diff
@@ -0,0 +1,91 @@
...
+        if(userLayerDbService == null) {
+            setUserLayerDbService(OskariComponentManager.getComponentOfType(UserLayerDbService.class));
+        }
+    }
+
+    @Override
+    public void handleAction(ActionParameters params) throws ActionException {
+        params.requireLoggedInUser();
+
+        final long id = ConversionHelper.getLong(params.getHttpParam(PARAM_ID), -1);
```
jampukka: Prefer `ActionParameters#getRequiredParamLong(String key)`

control-userlayer/src/main/java/org/oskari/control/userlayer/EditUserLayerHandler.java
```diff
@@ -0,0 +1,91 @@
...
+        final UserLayerStyle style = new UserLayerStyle();
+
+        if(userLayer == null) {
+            throw new ActionParamsException("Userlayer id doesn't exist: " + id);
+        }
+        if(!userLayer.isOwnedBy(params.getUser().getUuid())) {
+            throw new ActionDeniedException("Userlayer belongs to another user");
+        }
+               
+        userLayer.setLayer_name(params.getHttpParam(PARAM_NAME));
```
jampukka: If these parameters are not set do we want to replace them? If not could call `userLayer.setLayer_name(params.getHttpParam(PARAM_NAME, userLayer.getLayer_name());`

jampukka: Maybe this is better, since then it would be impossible for the frontend to set the value to null

control-userlayer/src/main/java/org/oskari/control/userlayer/GetUserLayerStyleHandler.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public void init() {
+        super.init();
+        if(userLayerDbService == null) {
+            userLayerDbService = OskariComponentManager.getComponentOfType(UserLayerDbService.class);
+        }
+    }
+    
+    public void handleAction(ActionParameters params) throws ActionException {
+        if(params.getUser().isGuest()) {
```
jampukka: Prefer `ActionParameters#requireLoggedInUser()`

control-userlayer/src/main/java/org/oskari/control/userlayer/GetUserLayerStyleHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+@OskariActionRoute("GetUserLayerStyle")
+public class GetUserLayerStyleHandler extends ActionHandler {
+    private final static String PARAM_ID = "id";
+    private UserLayerDbService userLayerDbService = null;
+    
+    @Override
+    public void init() {
+        super.init();
+        if(userLayerDbService == null) {
```
jampukka: No way to set the userLayerDbService so this will always be true

control-userlayer/src/main/java/org/oskari/control/userlayer/EditUserLayerHandler.java
```diff
@@ -0,0 +1,91 @@
...
+import fi.nls.oskari.util.ResponseHelper;
+import org.oskari.map.userlayer.service.UserLayerDataService;
+import org.oskari.map.userlayer.service.UserLayerDbService;
+
+/**
+ * Expects to get layer id as http parameter "id".
+ */
+@OskariActionRoute("EditUserLayer")
+public class EditUserLayerHandler extends ActionHandler {
+    private final static String PARAM_ID = "id";
```
jampukka: This could be static final for consistency with the rest of them and the rest of the codebase

control-userlayer/src/main/java/org/oskari/control/userlayer/GetUserLayerStyleHandler.java
```diff
@@ -0,0 +1,51 @@
...
+import fi.nls.oskari.util.ResponseHelper;
+import org.oskari.map.userlayer.service.UserLayerDbService;
+
+
+/**
+ * Expects to get layer id as http parameter "id".
+ */
+@OskariActionRoute("GetUserLayerStyle")
+public class GetUserLayerStyleHandler extends ActionHandler {
+    private final static String PARAM_ID = "id";
```
jampukka: static final is more used

jampukka: Prefer ActionParameters#getRequiredParamLong(String key)

control-userlayer/src/main/java/org/oskari/control/userlayer/GetUserLayerStyleHandler.java
```diff
@@ -0,0 +1,51 @@
...
+        if(userLayerDbService == null) {
+            userLayerDbService = OskariComponentManager.getComponentOfType(UserLayerDbService.class);
+        }
+    }
+    
+    public void handleAction(ActionParameters params) throws ActionException {
+        if(params.getUser().isGuest()) {
+            throw new ActionDeniedException("Session expired");
+        }
+        final long id = ConversionHelper.getLong(params.getHttpParam(PARAM_ID), -1);
```
jampukka: Prefer `ActionParameters#getRequiredParamLong(String key)`

service-userlayer/src/main/resources/org/oskari/map/userlayer/service/UserLayerMapper.xml
```diff
@@ -174,6 +174,7 @@
         stroke_dasharray,
         border_linejoin,
         border_dasharray
+		FROM user_layer_style
```
jampukka: So this wasn't working at all? Nice catch

service-base/src/main/java/fi/nls/oskari/domain/map/userlayer/UserLayerStyle.java
```diff
@@ -47,6 +49,35 @@ public void populateFromJSON(final JSONObject stylejs) throws JSONException {
 
     }
 
+    public JSONObject parseUserLayerStyle2JSON(){
```
jampukka: Maybe we should consider using Jackson and Json annotations for this. This quickly gets very complicated when the POJO model corresponds to the  doesn't correspond to the JSON model.

control-userlayer/src/main/java/org/oskari/control/userlayer/GetUserLayerStyleHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+        if(id == -1) {
+            throw new ActionParamsException("Parameter missing or non-numeric: " + PARAM_ID + "=" + params.getHttpParam(PARAM_ID));
+        }
+        final UserLayerStyle style = userLayerDbService.getUserLayerStyleById(id);
+        if (style != null){
```
jampukka: There is no check whether or not the style belongs to the current user. Should we check it? If not, I wonder if the logged-in requirement is necessary?

service-base/src/main/java/fi/nls/oskari/domain/map/userlayer/UserLayerStyle.java
jampukka: Anyway we should probably add a unit test that creates a UserLayerStyle object by hand, sets some value to all the fields and then calls `parseUserLayerStyle2JSON()` and `populateFromJSON(JSONObject json)` and checks that the all the fields match the original UserLayerStyle.

control-userlayer/src/main/java/org/oskari/control/userlayer/EditUserLayerHandler.java
```diff
@@ -0,0 +1,91 @@
...
+    private UserLayerDbService userLayerDbService = null;
+    private final UserLayerDataService userlayerService = new UserLayerDataService();
+    
+    public void setUserLayerDbService(final UserLayerDbService service) {
+        userLayerDbService = service;
+    }
+
+    @Override
+    public void init() {
+        super.init();
```
jampukka: Calling `super.init();` is unnecessary, since `ActionHandlers` implementation is empty.

control-userlayer/src/main/java/org/oskari/control/userlayer/GetUserLayerStyleHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ * Expects to get layer id as http parameter "id".
+ */
+@OskariActionRoute("GetUserLayerStyle")
+public class GetUserLayerStyleHandler extends ActionHandler {
+    private final static String PARAM_ID = "id";
+    private UserLayerDbService userLayerDbService = null;
+    
+    @Override
+    public void init() {
+        super.init();
```
jampukka: Calling `super.init();` is unnecessary, since `ActionHandlers` implementation is empty.

control-userlayer/src/main/java/org/oskari/control/userlayer/GetUserLayerStyleHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+        if(id == -1) {
+            throw new ActionParamsException("Parameter missing or non-numeric: " + PARAM_ID + "=" + params.getHttpParam(PARAM_ID));
+        }
+        final UserLayerStyle style = userLayerDbService.getUserLayerStyleById(id);
+        if (style != null){
+            JSONObject response = style.parseUserLayerStyle2JSON();
+            ResponseHelper.writeResponse(params, response);
+        }else{
+            throw new ActionException("Unable to get style for id" + id);
```
jampukka: Should this be ActionParamsException? The same case in `EditUserLayerHandler` is handled as one.

jampukka:
Added some easy-to-fix comments, mainly about consistency and about using existing helper functions.
The most important one is the one about adding a unit test for the "function pair" `UserLayerStyle#populateFromJSON` and `UserLayerStyle#parseUserLayerStyle2JSON`

control-userlayer/src/main/java/org/oskari/control/userlayer/EditUserLayerHandler.java
ZakarFin: Things will probably break with null so prefer getRequiredParam() or default to previous name.

control-userlayer/src/main/java/org/oskari/control/userlayer/GetUserLayerStyleHandler.java
ZakarFin: Yep, ActionParamsException is better since we don't need the stacktrace

PR #182, opened 2018-03-28
Decrease time spent in GetHierarchicalMapLayerGroups

jampukka:
GetHierarchicalMapLayerGroups reloads all the permissions from the database for each and every group and their subgroup. This PR changes the logic so that the permissions are only loaded once per Action call. The price to pay is that the function header for the recursive function
`GetMapLayerGroupsHandler#getGroupJSON()` got really ugly. `OskariLayerWorker` was already passing Sets of permissions around functions. A `PermissionCollection` class is introduced here, with the sole purpose of wrapping those permission Sets into a single object that is easier to pass around.

The effect is quite noticeable (these are from my dev environment with no js minifying or gzip enabled):
Current version (Waiting 21.22s):
getlayers
After (Waiting 4.05s):
getlayers2


ZakarFin:
Nice work again. The OskariLayerWorker class still needs some more love, but this update is definitely something that we want to include in the upcoming release.

PR #185, opened 2018-04-05
Calculate UserLayer extent one time when UserLayer is created

jampukka:
Fixes the issues created by #104. The functionality fixed here was disabled in #126.
Previously UserLayer extent was calculated from the data every time GetMapLayers was requested and it was also returned in the wrong projection to the functions requesting it (in the native srs where as it should have been in EPSG:4326).
Now we store the extent of UserLayer geometries in WGS84 as WKT directly at the user_layer level in the same INSERT statement that creates the user_layer row.

### DIFF
--- a/service-userlayer/src/main/java/org/oskari/map/userlayer/service/UserLayerDataService.java
+++ b/service-userlayer/src/main/java/org/oskari/map/userlayer/service/UserLayerDataService.java
@@ -73,7 +73,7 @@ public class UserLayerDataService {
                     extentWGS84.getMaxY());
         } catch (FactoryException | TransformException e) {
             // This shouldn't really happen since EPSG:4326 shouldn't be problematic
-            // and converting to it should always work. But if it does happen
+            // and transforming into it should always work. But if it does happen
             // there's probably something wrong with the geometries of the features
             throw new ServiceRuntimeException("Failed to transform bounding extent", e);
         }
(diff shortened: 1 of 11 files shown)


service-userlayer/src/main/java/org/oskari/map/userlayer/service/UserLayerDataService.java
```diff
@@ -44,19 +50,29 @@
     private static final int USERLAYER_BASE_LAYER_ID = PropertyUtil.getOptional(USERLAYER_BASELAYER_ID, -1);
 
     public static UserLayer createUserLayer(SimpleFeatureCollection fc,
-            String uuid, String name, String desc, String source) {
+            String uuid, String name, String desc, String source)
+                    throws FactoryException, TransformException {
```
ZakarFin: I would wrap the exceptions to a ServiceException since (in my opinion) the details are not that interesting that a list of exceptions is required. A bunch of possible exceptions a method can throw tends to lead to a catch(Exception e) error handling.

I'm also not sure if problems when calculating the coverage/bounds for the data should be treated as a failure or should we just leave the WKT empty. Then again if the transform fails it could indicate a problem with the data and as such should be treated as a failure so we don't store invalid data.

Any thoughts on these?

jampukka: Yea this isn't nice. Previously this method didn't even throw any exceptions, so even an ServiceException seems like a downgrade. I'm not even sure if the FactoryException _can_ occur, since the throwing call is fixed CRS.decode("EPSG:4326", true) which shouldn't fail. For the TransformException I agree that _if_ for some reason it does fail there's probably something wrong with the data (don't know when this could even happen, maybe when coordinates are outside of the extent of the source projection? They won't be outside of WGS84 anyway, so that one probably exists mainly on the source projection side. 3D -> 2D maybe?)

jampukka: I'll make the changes

PR #187, opened 2018-04-09
Fix secure flag so unsecure services are proxied like before

ZakarFin:
The isSecure flag is used to detect if services running in http-urls should be proxied through oskari-server. It was hardcoded to false in #145 and with the change in this PR the flag is detected from the request as before. Fixes an issue where Oskari instance running as https-service would try to reference wms-services using http-requests causing mixed content warnings in the browser.


jampukka:
Looks good

PR #189, opened 2018-04-13
Appsetup mapOptions are optional

ZakarFin:
Previously publishing a map from an appsetup that didn't have mapOptions configured resulted in an error. While missing mapOptions is an issue when there's multiple projections used in the system it isn't an issue when the system is relying on the defaults used by the frontend code.

Now detecting missing mapOptions is just logged as a possible misconfiguration issue instead of treating it as a show-blocker preventing creation of published maps.


jampukka:
OK

PR #190, opened 2018-04-16
Compatibility fix for maplayer dataprovider/group admin

ZakarFin:
The frontend for administrating layer dataproviders/groups were changed with oskariorg/oskari-frontend#270, oskariorg/oskari-frontend#263 and  #145 which broke the "classic" user interface for these features. This change makes the server handle both payloads. In the future the frontend should be harmonized to use the same input from each UI.

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/layer/SaveLayerHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/layer/SaveLayerHandler.java
@@ -324,11 +324,13 @@ public class SaveLayerHandler extends ActionHandler {
             }
         }
 
-        String groupId = params.getHttpParam(PARAM_MAPLAYER_GROUPS, "-1");
-        ml.emptyMaplayerGroups();
-        for (String id: groupId.split(",")) {
-            MaplayerGroup maplayerGroup = oskariMapLayerGroupService.find(ConversionHelper.getInt(id, -1));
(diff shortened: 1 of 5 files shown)


control-base/src/main/java/fi/nls/oskari/control/layer/SaveLayerHandler.java
```diff
@@ -324,11 +324,13 @@ private boolean handleRequestToMapLayer(final ActionParameters params, OskariLay
...
-        String groupId = params.getHttpParam(PARAM_MAPLAYER_GROUPS, "-1");
-        ml.emptyMaplayerGroups();
-        for (String id: groupId.split(",")) {
-            MaplayerGroup maplayerGroup = oskariMapLayerGroupService.find(ConversionHelper.getInt(id, -1));
-            ml.addGroup(maplayerGroup);
+        String groupId = params.getHttpParam(PARAM_MAPLAYER_GROUPS);
+        if(groupId != null) {
+            ml.emptyMaplayerGroups();
+            for (String id: groupId.split(",")) {
+                MaplayerGroup maplayerGroup = oskariMapLayerGroupService.find(ConversionHelper.getInt(id, -1));
```
jampukka: Probably no point in calling the find() if the getInt() returns -1, but this code should get replaced pretty soon anyways.

PR #191, opened 2018-04-17
In preparation for removing oskari_layer.externalId

jampukka:
Having two different columns used as the id of a maplayer is confusing and unnecessary. Currently there are multiple places in the codebase where a special logic is required to determine whether  `getExternalId()` or `getId()` should be used.

This is not a small task so it's best done in small parts. This PR makes the following changes in order to get us closer to the goal:
- Copy oskari_maplayer.externalid to a separate table.
- Add OskariLayerExternalIdService and a Mybatis implementation to access the table.
- `externalId` is no longer written to the client in `LayerJSONFormatter` as the id of the Layer (always use `id`).
- Add functionality to LayersParamHandler that replaces externalIds of oskari_maplayers with their id using the newly created service (to allow pre-existing links with e.g. `mapLayers` to work)
- Migrate externalIds in `BackgroundLayerSelectionPlugin` and `selectedLayers` in Mapfull Bundles `config` and `state` (because the client no longer knows the old externalIds of the layers)


content-resources/src/main/resources/flyway/oskari/V1_46_8__move_externalid_to_separate_table.sql
```diff
@@ -0,0 +1,14 @@
+BEGIN;
+
+SELECT id AS maplayerid, externalid
+INTO oskari_maplayer_externalid
+FROM oskari_maplayer
+WHERE externalid <> '';
+
+ALTER TABLE oskari_maplayer_externalid ADD PRIMARY KEY (maplayerid);
+ALTER TABLE oskari_maplayer_externalid ADD FOREIGN KEY (maplayerid) REFERENCES oskari_maplayer (id) ON DELETE CASCADE;
+ALTER TABLE oskari_maplayer_externalid ALTER externalid SET NOT NULL;
```
ZakarFin: External id is not unique at the moment, but I think it should be.

jampukka: Good idea. It should definitely be unique (as is assumed by the migrations in this PR for example).

content-resources/src/main/java/flyway/oskari/V1_46_9__replace_externalids_in_mapful_bglayerselection_plugin_config.java
```diff
@@ -0,0 +1,228 @@
...
+            JSONObject bgPlugin = findBGPlugin(plugins);
+            if (bgPlugin == null) {
+                return false;
+            }
+            JSONObject bgPluginConfig = bgPlugin.optJSONObject("config");
+            if (bgPluginConfig == null) {
+                return false;
+            }
+            JSONArray baseLayers = bgPluginConfig.optJSONArray("baseLayers");
+            return replaceExternalIds(baseLayers, externalIdToLayerId);
```
ZakarFin: In addition to Oskari.mapframework.bundle.mapmodule.plugin.BackgroundLayerSelectionPlugin also Oskari.mapframework.bundle.mapmodule.plugin.LayerSelectionPlugin might have layer references in the config:

- conf.defaultBaseLayer with a value of a single layer id
- conf.baseLayers with a value of array of layer ids

These need to be migrated as well.

ZakarFin:
Very nice and clean flyway migration. An additional plugin needs to be added to the migration, but otherwise ready for merging.

content-resources/src/main/java/flyway/oskari/V1_46_9__replace_externalids_in_mapful_bglayerselection_plugin_config.java
ZakarFin: Let's do another PR for this to get this show on the road.

PR #193, opened 2018-04-17
Layer<->Group modifications

jampukka:
Fix an issue with Layer belonging to multiple groups not working with the current version of GetHierarchicalMapLayerGroups.

Make OskariLayer unaware of the groups it belongs to.

Change the way links between maplayers and layergroups are handled:
- Add a service for accessing and managing the links that bind maplayers to layergroups
- Change SaveLayerHandler to use the new service
- Remove a JOIN from OskariLayerServiceIbatisImpl

Change the way GetHierarchicalMapLayerGroups is formed:
- Change the JSON of the response
- Write the layers as flat array instead of the group layout
- Write the layer groups as a tree with nodes having links to the layers (via their id)
- Improve performance by calling GroupService only once

Move `parseLayer()` from JSONLayerFormatter to LayerHelper as that is the only place the method is used.


control-base/src/main/java/fi/nls/oskari/control/layer/SaveLayerHandler.java
```diff
@@ -184,6 +185,12 @@ private SaveResult saveLayer(final ActionParameters params) throws ActionExcepti
 
                 ml.setUpdated(new Date(System.currentTimeMillis()));
                 mapLayerService.update(ml);
+
+                int[] groupIds = getMaplayerGroupIds(params.getHttpParam(PARAM_MAPLAYER_GROUPS, "-1"));
+                layerGroupLinkService.deleteLinksByLayerId(ml.getId());
```
ZakarFin: PARAM_MAPLAYER_GROUPS is not always sent by the frontend (when editing a layer and NOT opening the groups dialog) so this needs a check if the parameter was present.

PR #195, opened 2018-04-18
Id fix for myplaces/userlayer json generator

ZakarFin:
As LayerJSONFormatter no longer writes out the external id of layers as id a change is required for myplaces and userlayer functionality. After JSON has been created the id is overwritten with the generated id of myplaces_[id] or userlayer_[id]. Otherwise the frontend will receive a -1 id.


jampukka:
OK (hacky but a bit too costly to change _everything_)

PR #197, opened 2018-04-23
Remove unused database table: portti_stats_layer

ZakarFin:
The refactored statistics serverside code doesn't use the table anymore so removing it from the database


jampukka:
Good

PR #202, opened 2018-04-25
Layer and group order fix

kuosman:
This PR needs also https://github.com/oskariorg/oskari-frontend/pull/363

# GetMapLayerGroupsHandler
- group JSON layer array now includes layer object { "id":0, "orderNumber: 0 }

# SaveLayerHandler
- layer JSON now also return layer groups (before this layer groups not showed well when saving and updated layer)

### DIFF
--- a/service-map/src/main/java/fi/nls/oskari/map/layer/group/link/OskariLayerGroupLink.java
+++ b/service-map/src/main/java/fi/nls/oskari/map/layer/group/link/OskariLayerGroupLink.java
@@ -29,13 +29,6 @@ public class OskariLayerGroupLink {
         return layerId;
     }
 
-    public Map<String, Object> getLayerMap() {
-        Map<String, Object> map = new HashMap<>();
-        map.put("id", layerId);
-        map.put("orderNumber", orderNumber);
-        return map;
-    }
-
     public void setLayerId(int layerId) {
         this.layerId = layerId;
     }
--- a/service-map/src/main/java/fi/nls/oskari/map/layer/group/link/OskariLayerGroupLink.java
+++ b/service-map/src/main/java/fi/nls/oskari/map/layer/group/link/OskariLayerGroupLink.java
@@ -1,8 +1,5 @@
 package fi.nls.oskari.map.layer.group.link;
 
-import java.util.HashMap;
-import java.util.Map;
-
 public class OskariLayerGroupLink {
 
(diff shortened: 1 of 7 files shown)


service-map/src/main/java/fi/nls/oskari/map/layer/group/link/OskariLayerGroupLink.java
```diff
@@ -26,6 +29,13 @@ public int getLayerId() {
         return layerId;
     }
 
+    public Map<String, Object> getLayerMap() {
```
ZakarFin: Looks like this isn't used anywhere. Can it be removed?

service-map/src/main/java/fi/nls/oskari/map/layer/group/link/OskariLayerGroupLinkServiceMybatisImpl.java
```diff
@@ -118,4 +118,11 @@ public boolean hasLinks(int groupId) {
         }
     }
 
+    @Override
+    public List<OskariLayerGroupLink> findByMaplayerId(int layerId){
```
ZakarFin: This is a duplicate implementation for the  findByLayerId() method in the same file. Remove this one and change callers to use findByLayerId().

service-map/src/main/java/fi/nls/oskari/map/layer/group/link/OskariLayerGroupLinkService.java
```diff
@@ -17,5 +17,6 @@
 
     public void replace(OskariLayerGroupLink old, OskariLayerGroupLink link);
     public boolean hasLinks(int groupId);
+    public List<OskariLayerGroupLink> findByMaplayerId(int layerId);
```
ZakarFin: There's a findByLayerId() in the interface already. What would be the difference between these two methods?

service-map/src/main/java/fi/nls/oskari/map/layer/group/link/OskariLayerGroupLink.java
```diff
@@ -1,5 +1,8 @@
 package fi.nls.oskari.map.layer.group.link;
 
+import java.util.HashMap;
```
ZakarFin: Please cleanup the imports as well :)

PR #203, opened 2018-04-25
MapfullHandler transform userlayer wkt to map's srs

okauppinen:
Userlayer extent/coverage is stored in the WGS84 projection.
Now mapfullhandler transform's userlayer wkt to map's srs.


jampukka:
Good job. Too bad this code is needed in so many different places. Hope we don't find yet another place that still requires this.

PR #209, opened 2018-05-07
Statistical datasource (Pxweb) data fetching improvement

ZakarFin:
Some statistical datasources have prefixed region IDs. For example municipality ID could be prefixed with something like "MK" (region "003" in regionset would be referenced with "MK003" in the statistical data). This change allows such datasources to be used with Oskari and provides a plugin mechanism for future requirements.

This requires the following configuration to be added for datasource layer link table (oskari_statistical_layer):

    {
        "valueProcessor": "fi.nls.oskari.control.statistics.plugins.pxweb.parser.PrefixedRegionsValueProcessor",
        "statsRegionPrefix": "[prefix for region ids in statistical data like MK]"
    }


service-statistics-pxweb/src/main/java/fi/nls/oskari/control/statistics/plugins/pxweb/PxwebStatisticalDatasourcePlugin.java
```diff
@@ -179,4 +185,27 @@ private String createUrl(String baseUrl, String pathId) {
...
+        if(className == null) {
+            return DEFAULT_PROCESSOR;
+        }
+        ValueProcessor processor = processors.get(className);
+        if(processor != null) {
+            return processor;
+        }
+        try {
+            final Class clazz = Class.forName(className);
+            processor = (ValueProcessor) clazz.newInstance();
```
jampukka: Could add a `ValueProcessor.class.isAssignableFrom(clazz);` check here. The `Class.forName(String)` still needs to be inside the try-catch so doesn't make that much of a difference I guess.

ZakarFin: Added this check so misconfiguration now gives a more pleasant log-message without the stacktrace.

service-statistics-pxweb/src/main/java/fi/nls/oskari/control/statistics/plugins/pxweb/parser/PrefixedRegionsValueProcessor.java
```diff
@@ -0,0 +1,40 @@
...
+ *  {
+ *      "valueProcessor": "fi.nls.oskari.control.statistics.plugins.pxweb.parser.PrefixedRegionsValueProcessor",
+ *      "statsRegionPrefix": "[prefix for region ids in statistical data]"
+ *  }
+ */
+public class PrefixedRegionsValueProcessor extends ValueProcessor {
+
+    public Optional<RegionValue> getRegionValue(JSONArray values, String regionId, int valueIndex, DatasourceLayer layer) {
+        Optional<RegionValue> value = super.getRegionValue(values, regionId, valueIndex, layer);
+        if(!value.isPresent()) {
```
jampukka: If you want to have more Optional you could move the `regionIdPrefix == null` check up and then do a 
```
return value.filter(v -> v.getRegion().startsWith(prefix))
    .map(v -> new RegionValue(v -> v.getRegion().substring(prefix.length()), v.getValue()));
```

service-statistics-pxweb/src/main/java/fi/nls/oskari/control/statistics/plugins/pxweb/parser/ValueProcessor.java
```diff
@@ -0,0 +1,32 @@
...
+import fi.nls.oskari.control.statistics.data.RegionValue;
+import fi.nls.oskari.control.statistics.plugins.db.DatasourceLayer;
+import org.json.JSONArray;
+
+import java.util.Optional;
+
+/**
+ * Returns the region ID/value pair if one is found.
+ */
+public class ValueProcessor {
```
jampukka: Maybe this could be an exciting JDK8 interface with a default method. e.g. `default public Optional<RegionValue> getRegionValue(JSONArray values, String regionId, int valueIndex, DatasourceLayer layer) { ... }`

jampukka: Scratch that, we can't instantiate that :D Maybe this should still implement an interface still, because that's what we want to do here.

service-statistics-pxweb/src/main/java/fi/nls/oskari/control/statistics/plugins/pxweb/PxwebStatisticalDatasourcePlugin.java
```diff
@@ -179,4 +185,31 @@ private String createUrl(String baseUrl, String pathId) {
         }
         return baseUrl;
     }
+
+    private ValueProcessor getValueProcessor(DatasourceLayer layer) {
+        if(layer.getConfig() == null) {
```
jampukka: This check is not unnecessary since `DatasourceLayer#getConfig(String key)` does the null check for you.

service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/data/RegionValue.java
```diff
@@ -0,0 +1,20 @@
+package fi.nls.oskari.control.statistics.data;
+
+public class RegionValue {
+    private String region;
```
jampukka: These fields could be marked as `final` even though they already are effectively `final`.

service-statistics-pxweb/src/main/java/fi/nls/oskari/control/statistics/plugins/pxweb/parser/ValueProcessor.java
ZakarFin: I'm not at all sure that this is an hook/plugin interface we are going to keep in the long run. The function params seem kinda weird to me. Anyway it fulfills the requirement for now and can be improved upon later.

service-statistics-pxweb/src/main/java/fi/nls/oskari/control/statistics/plugins/pxweb/PxwebStatisticalDatasourcePlugin.java
ZakarFin: True, I used getConfig().optString() before in the code and this got left after the change. Good catch.

PR #212, opened 2018-05-16
Add admin route for flushing a cache

ZakarFin:
This adds the ability to make a POST request to /action?action_route=Cache&name=[cache name] when logged in as admin to flush a given cache. Also adds a more user-friendly formatting for the "time to expiration" info.


jampukka:
OK

PR #213, opened 2018-05-17
Improve .gpx importing

jampukka:
Currently .gpx import uses the first typeName GeoTools reports (that is not `track_points`). This typeName can easily be empty leading to errors in later stages of the importing process. Instead search the file for typeNames `tracks`, `routes`, `waypoints` (in that order) and use the first one that returns a non-empty collection of Features.
Also add a maximum (100) number of attempts we make to create a unique temporary directory. While the current implementation should never end up in an infinite loop this change makes sure it will never happen.
And also move away from Files.copy(), don't want to deal with CopyOptions (even though current implementation _might_ be correct)


ZakarFin:
Nice!

PR #214, opened 2018-05-18
Domain check is now case-insensitive

ZakarFin:
The domain given by the user when publishing a map doesn't need to be lowercase anymore.


jampukka:
OK

jratike80:
Näyttää menevän standardin mukaisesti nyt http://www.ietf.org/rfc/rfc1034.txt.
>> By convention, domain names can be stored with arbitrary case, but
domain name comparisons for all present domain functions are done in a
case-insensitive manner, assuming an ASCII character set, and a high
order zero bit.  This means that you are free to create a node with
label "A" or a node with label "a", but not both as brothers; you could
refer to either using "a" or "A"

PR #216, opened 2018-05-29
Updated user indicator operations for the new statistical maps

ZakarFin:
- Added a saveIndicator/saveIndicatorData interface skeletons for the statistical datasources plugins
- Created implementations for saveIndicator/saveIndicatorData for UserIndicatorsStatisticalDatasourcePlugin
- Added the missing user indicator methods in MyBatis service for the new user indicator db schema 
- Removed the deprecated Ibatis service
- Some other refactoring for constants and helpers


ZakarFin:
Blocks oskariorg/oskari-frontend#391

ZakarFin:
In short this adds the server operations needed to add, edit and remove user generated statistical indicators and/or datasets (year/regionset specific data) for these indicators.

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/AddIndicatorDataHandler.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public void init() {
+        super.init();
+        if (indicatorService == null) {
+            indicatorService = OskariComponentManager.getComponentOfType(StatisticalIndicatorService.class);
+        }
+    }
+
+    private static String PARAM_DATA = "data";
```
jampukka: This field should be `final` and should be declared at the top of the class.

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/AddIndicatorDataHandler.java
```diff
@@ -0,0 +1,109 @@
...
+
+        int datasourceId = params.getRequiredParamInt(StatisticsHelper.PARAM_DATASOURCE_ID);
+        StatisticalDatasourcePlugin datasource = StatisticalDatasourcePluginManager.getInstance().getPlugin(datasourceId);
+        if(datasource == null) {
+            throw new ActionParamsException("Invalid datasource:" + datasourceId);
+        }
+        if(!datasource.canModify(params.getUser())) {
+            throw new ActionDeniedException("User doesn't have permission to modify datasource:" + datasourceId);
+        }
+        String id = params.getRequiredParam(ActionConstants.PARAM_ID);
```
jampukka: Since these parameters are required it would probably be better if we checked for their presence and correctness right at the beginning of the function instead. That way we don't waste processing on requests that are bound to fail.

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/AddIndicatorDataHandler.java
```diff
@@ -0,0 +1,109 @@
...
+        try {
+            // read json as map with strings and doubles
+            Map<String, Double> values = MAPPER.readValue(data, new TypeReference<Map<String, Double>>(){});
+            // map doubles to IndicatorValueFloat
+            return values.entrySet().stream()
+                    .collect(Collectors
+                            .toMap( e -> e.getKey(),
+                                    e -> new IndicatorValueFloat(e.getValue())));
+        } catch (IOException e) {
+            throw new ActionParamsException("Invalid parameter value for key: " + StatisticsHelper.PARAM_SELECTORS + " - expected JSON object");
```
jampukka: `PARAM_DATA` is used to retrieve the data, but `PARAM_SELECTORS` is used in the error message.

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/AddIndicatorDataHandler.java
```diff
@@ -0,0 +1,109 @@
...
+        DatasourceLayer layer = datasource.getSource().getLayers().stream()
+                .filter(x -> x.getMaplayerId() == regionsetId)
+                .findFirst()
+                .orElseThrow(() -> new ActionParamsException("Invalid regionset: " + regionsetId));
+        if(indicator.getLayer(layer.getMaplayerId()) == null) {
+            indicator.addLayer(layer);
+        }
+        try {
+            indicator.setDataModel(model);
+            Map<String, IndicatorValue> data = parseIndicatorData(params.getHttpParam(PARAM_DATA));
```
jampukka: Consider moving this line and the previous line outside of the `try-catch-all` block so that the ActionExceptions thrown by `parseIndicatorData(String data)` are thrown as-is.

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/DeleteIndicatorHandler.java
```diff
@@ -0,0 +1,77 @@
...
+            throw new ActionDeniedException("Unknown indicator/not the owner: " + id );
+        }
+        JSONObject selectors = params.getHttpParamAsJSON(StatisticsHelper.PARAM_SELECTORS);
+        if(selectors != null) {
+            // if selectors present, regionset is also required
+            int regionset = params.getRequiredParamInt("regionset");
+            try {
+                indicatorService.deleteIndicatorData(id, regionset, selectors.getInt("year"));
+            } catch (JSONException ex) {
+                throw new ActionParamsException("Year was not part of the selectors parameter");
```
jampukka: This caused a moment of confusion. I thought to myself: why would `StatisticalIndicatorService#deleteIndicatorData(...)` throw a JSONException and how could one determine that the error must have been due to the selectors not having a year parameter. Then I understood that it is the `selectors.getInt("year")` that's being `try`'ed here.

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/DeleteIndicatorHandler.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+
+            StatisticsHelper.flushDataFromCache(datasourceId, Integer.toString(id), regionset, selectors);
+        } else if (!indicatorService.delete(id, params.getUser().getId())) {
```
jampukka: Maybe separate the `if` here, e.g.
```
} else {
    if (!delete()) {
        throw new Exception()
    }
}
```

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/DeleteIndicatorHandler.java
```diff
@@ -0,0 +1,77 @@
...
+            throw new ActionParamsException("Indicator wasn't removed: " +  + id);
+        }
+
+        StatisticalDatasourcePlugin datasource = StatisticalDatasourcePluginManager.getInstance().getPlugin(datasourceId);
+        if (datasource != null && datasource.canCache()) {
+            // TODO: flush/update caches
+            // Not an issue for now since user indicators are not cached and they are are the
+            //  only ones that can be added/edited/removed
+        }
+        LOG.info("Deleted indicator", id);
```
jampukka: Is this accurate? If `selectors` is not `null` the indicator isn't deleted, only the data, correct? This still gets executed if only the data is deleted.

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/SaveIndicatorHandler.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public void init() {
+        super.init();
+        if (indicatorService == null) {
+            indicatorService = OskariComponentManager.getComponentOfType(StatisticalIndicatorService.class);
+        }
+    }
+
+    public static String PARAM_NAME = "name";
```
jampukka: This should be `final` and place at the top of the class

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/SaveIndicatorHandler.java
```diff
@@ -0,0 +1,86 @@
...
+        if (id != null) {
+            existingIndicator = datasource.getIndicator(params.getUser(), id);
+            if(existingIndicator == null) {
+                // indicator removed or is not owned by this user
+                throw new ActionParamsException("Requested invalid indicator:" + id);
+            }
+        }
+        StatisticalIndicator indicator = parseIndicator(params, existingIndicator);
+        try {
+            existingIndicator = datasource.saveIndicator(indicator, params.getUser());
```
jampukka: Having two local variables for the "same" indicator is a bit confusing. I'd change to `parseIndicator(ActionParameters, StatisticalDatasourcePlugin)` and move the existingIndicator logic to that function.

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/DeleteIndicatorHandler.java
ZakarFin: I'll move the delete call to another method so the code looks cleaner.

control-statistics/src/main/java/fi/nls/oskari/control/statistics/user/DeleteIndicatorHandler.java
ZakarFin: True. I'll move the logging.

service-statistics/src/main/java/org/oskari/statistics/user/StatisticalIndicatorServiceMybatisImpl.java
```diff
@@ -94,37 +98,110 @@ public void deleteByUser(long userId) {
...
+        ind.getDataModel().getDimension("year").addAllowedValue(String.valueOf(row.year));
+    }
+
+    public StatisticalIndicator saveIndicator(StatisticalIndicator ind, long userId) {
+        int id = ConversionHelper.getInt(ind.getId(), -1);
+        UserIndicatorDataRow row = createDBRow(ind);
+        row.userId = userId;
+        row.id = id;
+        if (id != -1) {
+            // check that it's the users indicator
```
jampukka: I'd argue this check is unnecessary. The `UserIndicatorMapper#updateIndicator(row)` will return 0 if the row-to-update doesn't exist, so this case is covered twice.

service-statistics/src/main/java/org/oskari/statistics/user/StatisticalIndicatorServiceMybatisImpl.java
```diff
@@ -94,37 +98,110 @@ public void deleteByUser(long userId) {
...
+            // insert new data
+            getMapper(session).addData(indicator, regionset, year, data);
+            session.commit();
+        }
+    }
+    public boolean deleteIndicatorData(long indicator, long regionset, int year) {
+        try (final SqlSession session = factory.openSession()) {
+            int updated = getMapper(session).deleteData(indicator, regionset, year);
+            if(updated != 1) {
+                throw new ServiceRuntimeException("Delete wasn't successful");
```
jampukka: I wouldn't expect `public boolean delete()` to throw a RuntimeException in case there was nothing to delete. Instead I'd expect false to be returned, and an Exception to be thrown only in the case there was an actual Exception (like the DB connection unexpectedly error'd or something)

service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/data/StatisticalIndicatorDataDimension.java
```diff
@@ -51,7 +51,13 @@ public void addAllowedValue(String value, String label) {
         addAllowedValue(new IdNamePair(value, label));
     }
     public void addAllowedValue(IdNamePair pair) {
-        allowedValues.add(pair);
+        // the values need to be sorted so doing "set-like" functionality here
+        if(!getAllowedValues().stream()
```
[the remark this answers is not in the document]
ZakarFin: For now it doesn't need to. For possible optimization I guess it might be a nice touch. But as the existing methods don't have a return value I'll just leave the overloaded one behaving the same way.

jampukka:
Looks good. A bit too many changes, so can't be sure everything actually works. However any possible bugs should be easily fixable.

PR #219, opened 2018-06-05
Migrate AppSetups from OpenLayers 2 to 4

ZakarFin:
Oskari 1.47.0 drops support for OpenLayers 2 (oskariorg/oskari-docs#63). This migrates any existing appsetups of type 'DEFAULT' and 'USER' to use the OpenLayers 4 implementations of the bundles. The other types (PUBLISH and PUBLISHED) have been using OL 4 for a long time already and don't need the migration.

Note! Any app using custom view types needs to make an app specific migration for said types.


jampukka:
OK, hope it works :)

PR #222, opened 2018-06-08
Internal layers

ZakarFin:
Introduces a concept of internal/private layers. Layers that are only used internally by the oskari-server that you never want end-users to see. These include the baselayers for my places, userlayers and analyze layers.


jampukka:
Looks good

PR #224, opened 2018-06-08
Remove deprecated classes

ZakarFin:
This removes the proxying action route for servlet-printout (removed in 1.46.0) and classes for generating an SLD style for WMS-based statistical maps (current frontend uses vector features).


jampukka:
Love it

PR #225, opened 2018-06-11
Reverse processing order for loading statistics indicators list

ZakarFin:
When indicator metadatas are being processed by oskari-server they are first saved under temp-key as a list in Redis. When the information in the temporary list is written to the actual cache the results are processed from the last to the first item in the listing (using RPOP on Redis). This results in the list being in reversed order compared to the original one from the service. This change makes the processing take the first item/head of the list (using LPOP on Redis) to correct the order of the listing.


jampukka:
OK

PR #233, opened 2018-06-29
Statistics maps regionsets handling

ZakarFin:
Statistics regionsets are currently handled as maplayers, but they don't really function too well outside the thematic maps user interface. This change adds a listing of regionsets with id and name to the statsgrid bundle config. These can be used to create the vectorlayers that the thematic map uses without having the regionsets as "normal" maplayers in the frontend.

This a step towards removing statslayers from the maplayer listing on the frontend and allows the frontend to use this information instead of searching for maplayers of type "statslayer". After the frontend change the statslayer maplayers can be removed from the layers response going to the frontend.

### DIFF
--- a/service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/plugins/StatisticalDatasourceFactory.java
+++ b/service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/plugins/StatisticalDatasourceFactory.java
@@ -1,5 +1,6 @@
 package fi.nls.oskari.control.statistics.plugins;
 
+import fi.mml.portti.domain.permissions.Permissions;
 import fi.nls.oskari.control.statistics.plugins.db.DatasourceLayer;
 import fi.nls.oskari.control.statistics.plugins.db.DatasourceLayerMapper;
 import fi.nls.oskari.control.statistics.plugins.db.StatisticalDatasource;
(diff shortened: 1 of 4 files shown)


service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/plugins/StatisticalDatasourceFactory.java
```diff
@@ -28,15 +34,37 @@ public void setupSourceLayers(StatisticalDatasource source) {
         final DatasourceHelper helper = DatasourceHelper.getInstance();
         final DataSource dataSource = helper.getDataSource(helper.getOskariDataSourceName());
         final SqlSessionFactory factory = initializeIBatis(dataSource);
-        final SqlSession session = factory.openSession();
-        try {
+        try (final SqlSession session = factory.openSession()) {
             final List<DatasourceLayer> layerRows = session.getMapper(DatasourceLayerMapper.class).getLayersForDatasource(source.getId());
+
+            // fetch a list of permissions for the regionset layers
+            Map<Long, List<Permissions>> permissions = ServiceFactory.getPermissionsService()
```
jampukka: This is difficult to read due to current indenting and line breaks. Would be better if the maplayerIds were collected on a separate line.

PR #234, opened 2018-06-29
Add support for proxying all map layer requests

jheijari:


ZakarFin:
Adds an option to force using proxy urls for all layers even if they don't require credentials. Usable for example working around CORS issues.

PR #235, opened 2018-07-10
Printout scale text

kuosman:
This PR allows PDF print show scale text.
Scale text is showed when user has selected map scale on UI.

Also reorganized code so at layer filtering has done on GetPrinterHandler.

This option is visilbe when also https://github.com/oskariorg/oskari-frontend/pull/437 is merged and bundle config used.


service-print/src/main/java/org/oskari/print/request/PrintRequest.java
```diff
@@ -131,4 +133,23 @@ public void setShowLogo(boolean showLogo) {
...
+    public void setScaleText(String scaleText) {
+        this.scaleText = scaleText;
+    }
+
+    public boolean isScaleText(){
+        return (this.scaleText != null && !this.scaleText.isEmpty());
+    }
+
+    public boolean isUseConfiguredScale() {
+        return useConfiguredScale;
```
ZakarFin: I'd just use return isScaleText() here and remove setUseConfiguredScale()

control-base/src/main/java/fi/nls/oskari/control/data/GetPrintHandler.java
```diff
@@ -118,19 +120,44 @@ private PrintRequest createPrintRequest(ActionParameters params)
         request.setShowLogo(params.getHttpParam(PARM_LOGO, false));
         request.setShowScale(params.getHttpParam(PARM_SCALE, false));
         request.setShowDate(params.getHttpParam(PARM_DATE, false));
-
+        request.setScaleText(params.getHttpParam(PARM_SCALE_TEXT, ""));
         setPagesize(params, request);
         setCoordinates(params.getRequiredParam(PARM_COORD), request);
         setFormat(params.getRequiredParam(PARM_FORMAT), request);
+        request.setUseConfiguredScale(params.getHttpParam(PARM_USE_CONFIGURED_SCALE, false));
```
ZakarFin: I don't see a reason for using two parameters to trigger the UseConfiguredScale-mode.

control-base/src/main/java/fi/nls/oskari/control/data/GetPrintHandler.java
```diff
@@ -118,19 +120,44 @@ private PrintRequest createPrintRequest(ActionParameters params)
...
 
         List<PrintLayer> layers = getLayers(params.getRequiredParam(PARM_MAPLAYERS),
                 params.getUser());
+        layers = filterLayersWithZeroOpacity(layers);
```
ZakarFin: Layers with 0 opacity are already filtered out in getLayers: https://github.com/dimenteq/oskari-server/blob/7b798ba64ce0e35dfe9ddab56f8f2dd2dbd8bd31/control-base/src/main/java/fi/nls/oskari/control/data/GetPrintHandler.java#L282

So this can be removed.

service-print/src/main/java/org/oskari/print/request/PrintRequest.java
```diff
@@ -18,6 +18,8 @@
     private boolean showDate;
     private String title;
     private List<PrintLayer> layers;
+    private String scaleText;
+    private boolean useConfiguredScale;
```
jampukka: This isn't used

control-base/src/main/java/fi/nls/oskari/control/data/GetPrintHandler.java
```diff
@@ -95,16 +96,16 @@ public GetPrintHandler(OskariLayerService layerService,
...
-            handlePDF(pr, params);
-            break;
-        case PNG:
-            handlePNG(pr, params);
-            break;
-        default:
-            throw new ActionParamsException(String.format(
-                    "Invalid value for key '%s'. Allowed values are: %s",
-                    PARM_FORMAT, ALLOWED_FORMATS));
+            case PDF:
```
jampukka: These indent-only changes inside the switch blocks aren't necessary, please remove them.

service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -263,15 +263,15 @@ private static void drawScale(PDPageContentStream stream, PrintRequest request)
...
-        case "m":
-            mppx = request.getResolution();
-            break;
-        case "°":
-            LOG.info("Map units is deegrees, not drawing Scale Line");
-            return;
-        default:
-            LOG.info("Unknown unit", units, "- not drawing Scale line");
-            return;
+            case "m":
```
jampukka: Same thing here with the indent-only changes

service-print/src/main/java/org/oskari/print/request/PrintRequest.java
```diff
@@ -131,4 +133,16 @@ public void setShowLogo(boolean showLogo) {
...
 
+    public String getScaleText() {
+        return scaleText;
+    }
+
+    public void setScaleText(String scaleText) {
+        this.scaleText = scaleText;
+    }
+
+    public boolean isScaleText(){
```
jampukka: Maybe this function could have a more meaningful name? `isScaleText()` correctly explains what the function does, but the "if scaleText, force this"-part is missing

jampukka: Or maybe add a comment explaining the "why" to the `PDF#drawScale(PDPageContentStream stream, PrintRequest request)` function

service-print/src/main/java/org/oskari/print/PrintService.java
```diff
@@ -1,5 +1,6 @@
 package org.oskari.print;
 
+import fi.nls.oskari.domain.map.OskariLayer;
```
jampukka: This change seems unnecessary

jampukka:
This is rather safe to merge since it changes nothing for the requests that do not include the `scaleText` variable. However I'm not sure if these changes are enough to make sure the PDFs created actually are in the specified scale.

ZakarFin:
Some cleanup required before merging ^

service-print/src/main/java/org/oskari/print/PrintService.java
```diff
@@ -29,24 +26,12 @@ public PrintService(WMTSCapabilitiesCache wmtsCapsCache) {
...
         return PNG.getBufferedImage(request, wmtsCapsCache);
     }
 
     public void getPDF(PrintRequest request, PDDocument doc)
             throws IOException, ServiceException {
-        request.setLayers(filterLayersWithZeroOpacity(request.getLayers()));
         PDF.getPDF(request, doc, wmtsCapsCache);
     }
 
-    private static List<PrintLayer> filterLayersWithZeroOpacity(List<PrintLayer> layers) {
```
jampukka: This function shouldn't be removed. In one commit it was moved `GetPrintHandler` which was ~okay (breaks the functionality for other Action routes making straight calls to PrintService, but such routes do not exist, yet atleast), but now it is just removed.

jampukka:
Looks good to me

PR #240, opened 2018-08-10
Cross-site request forgery protection

ZakarFin:
Enables CSRF functionality for the Spring Security framework in Oskari. Other requests than HTTP GET now requires a token to be sent with the request as header or parameter.

**Note! If you have customized the JSP page for geoportal you need to add this to the login form:**

    <input type="hidden" name="${_csrf.parameterName}" value="${_csrf.token}"/>

Any write operations performed by the server now required POST request to be made (with the token). There's a couple of Action routes that are no longer used that are also removed in this PR. The frontend has been modified to send the token in oskariorg/oskari-frontend#450. Also oskariorg/oskari-frontend#451 changes the relevant calls made by the frontend to use HTTP POST instead of GET.


jampukka:
Looks good.

PR #242, opened 2018-08-15
Support for printing MyPlaces

jampukka:
Currently `myplaces_` layers are ignored in the GetPrint ActionRoute and the underlying service. This adds support for them. The images for the layers are created the same `MyPlacesTile` creates them (via ProxyService).

`CommandLoadImageMyPlaces` uses two magic keys (`"myCat"` and `"myplacestile"`) without referring to the original static final fields. We could use the original Strings, but that would require adding dependencies to the modules where they live to the print-service module.

Most of the +lines come from `ModifiedHttpServletRequest` which does nothing but serves as a "mock" HttpServletRequest with easily customisable parameter map.


control-base/src/main/java/fi/nls/oskari/control/data/GetPrintHandler.java
```diff
@@ -354,6 +360,23 @@ private PrintLayer findById(List<PrintLayer> layers, int id) {
...
+            // Ignore fully transparent layers
+            return null;
+        }
+
+        PrintLayer layer = new PrintLayer();
+        layer.setId(categoryId);
+        layer.setType("myplaces");
+        layer.setVersion("1.3.0");
+        layer.setName("oskari:my_places_categories");
+        layer.setSrsName("EPSG:3067");
```
ZakarFin: You should take srs name from properties. There's something like oskari.native.srs that's used elsewhere in the codebase.

jampukka: Better yet I should just remove the line all together, as it makes no difference what the value is in this case. The value added to the proxied GetMap request on the `crs`/`srs` key is determined by the `PrintRequest` not by the `PrintLayer`.

PR #244, opened 2018-08-15
Ignore CSRF for logout

ZakarFin:
Added http.csrf().ignoringAntMatchers(logoutUrl); to the common logout config to make it compatible with PTI preauth login.


ZakarFin:
Also changed Spring annotation from deprecated one to the current one in SAML config

jampukka:
Good

PR #246, opened 2018-08-17
Add CSRF tokens to user registration

ZakarFin:
Hopefully last piece of the CSRF puzzle. User registration forms/XHRs now include the token and are not rejected by the server.


jampukka:
OK

PR #251, opened 2018-08-29
Added handler for cam-parameter

jheijari:
Parameter handler to handle camera location in experimental 3D views.


jampukka:
Looks good, only complaint is that the logger is not used at all.

ZakarFin:
This could also be moved to nls-oskari/kartta.paikkatietoikkuna.fi as the frontend code is also ”app-specific” for now.

jheijari:
Moved to nls-oskari/kartta.paikkatietoikkuna.fi https://github.com/nls-oskari/kartta.paikkatietoikkuna.fi/pull/47

PR #254, opened 2018-09-10
Fix axis order issue for WFS-T requests in MyPlaces

jampukka:
Currently we use "new" URL-based format for `srsName` in the WFS-T requests. This leads to GeoServer handling them as if they were in the axis order specified by the CRS. However, what we actually want is to use fixed order of lon/lat (x/y). The solution is to use `http://www.opengis.net/gml/srs/epsg.xml#XXXX` which is the WFS 1.0.0 way to specify the CRS and which is always handled as lon/lat. For more information, see http://docs.geoserver.org/latest/en/user/services/wfs/basics.html#wfs-basics-axis


jampukka:
See https://github.com/oskariorg/oskari-docs/issues/71

PR #266, opened 2018-09-26
Initial support for creating MVT tiles from WFS layers

jampukka:
Add new module service-mvt which allows generating MVT tiles from GeoTools SimpleFeatureCollections.
Add new module service-mvt-wfs which contains useful methods for downloading and caching GeoTools SimpleFeatureCollections from WFS layers.
Add new module control-mvt. Contains ActionHandlers that deal with MVT tiles.

### DIFF
--- a/control-mvt/src/main/java/org/oskari/control/mvt/GetWFSVectorTileHandler.java
+++ b/control-mvt/src/main/java/org/oskari/control/mvt/GetWFSVectorTileHandler.java
@@ -46,6 +46,7 @@ public class GetWFSVectorTileHandler extends ActionHandler {
     static {
         KNOWN_TILE_GRIDS = new HashMap<>();
         KNOWN_TILE_GRIDS.put("EPSG:3067", new WFSTileGrid(new double[] { -548576, 6291456, -548576 + (8192*256), 6291456 + (8192*256) }, 15));
+        KNOWN_TILE_GRIDS.put("EPSG:3857", new WFSTileGrid(new double[] { -20037508.3427892, -20037508.3427892, 20037508.3427892, 20037508.3427892 }, 18));
     }
 
     private final Cache<byte[]> cache = new Cache<>();
@@ -139,7 +140,6 @@ public class GetWFSVectorTileHandler extends ActionHandler {
             sfc = sfc == null ? tileFeatures : union(sfc, tileFeatures);
         }
 
-        // long t0 = System.currentTimeMillis();
         double[] bbox = grid.getTileExtent(new TileCoord(z, x, y));
         byte[] encoded = SimpleFeaturesMVTEncoder.encodeToByteArray(sfc, layer.getName(), bbox, 4096, 256);
 


service-base/src/main/java/fi/nls/oskari/cache/Cache.java
```diff
@@ -9,6 +9,7 @@
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.function.Function;
```
ZakarFin: This can be removed.

service-mvt-wfs/src/main/java/org/oskari/service/mvt/wfs/OskariCachingSchemaLocator.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class OskariCachingSchemaLocator implements XSDSchemaLocator {
+
+    private static final Logger LOG = LogFactory.getLogger(OskariCachingSchemaLocator.class);
+    private static final Map<String, XSDSchema> CACHE = new ConcurrentHashMap<String, XSDSchema>();
```
ZakarFin: Should this use ComputeOnceCache as well with a big limit and long expiration?

ZakarFin: That way we could programmatically flush the cache.

jampukka: Yes

control-mvt/src/main/java/org/oskari/control/mvt/GetWFSVectorTileHandler.java
```diff
@@ -0,0 +1,270 @@
...
+import fi.nls.oskari.control.ActionException;
+import fi.nls.oskari.control.ActionHandler;
+import fi.nls.oskari.control.ActionParameters;
+import fi.nls.oskari.control.ActionParamsException;
+import fi.nls.oskari.domain.map.OskariLayer;
+import fi.nls.oskari.map.layer.OskariLayerService;
+import fi.nls.oskari.util.IOHelper;
+import fi.nls.oskari.util.ResponseHelper;
+
+@OskariActionRoute("GetWFSVectorTile")
```
ZakarFin: Add some comment about only supporting metric projections.

service-mvt/src/main/java/org/oskari/service/mvt/SimpleFeaturesMVTEncoder.java
```diff
@@ -0,0 +1,298 @@
...
+                        || geom instanceof MultiLineString
+                        || geom instanceof MultiPolygon) {
+                    double bboxWidth = geomEnvelope.getWidth();
+                    double bboxHeight = geomEnvelope.getHeight();
+                    if (bboxWidth * bboxWidth + bboxHeight * bboxHeight < resHalf * resHalf) {
+                        continue;
+                    }
+                }
+
+                // geom = editor.edit(geom, removeRepeatedPoints);
```
ZakarFin: To make the loop more compact, everything up to this point could possible be moved to some method like:

     Geometry geom = validate(sf.getDefaultGeometry(), tileEnvelope)

service-mvt/src/main/java/org/oskari/service/mvt/SimpleFeaturesMVTEncoder.java
```diff
@@ -0,0 +1,298 @@
...
+        Geometry tileClipGeom = GF.toGeometry(clipEnvelope);
+        RectangleIntersects rectIntersects = new RectangleIntersects((Polygon) tileEnvelopeGeom);
+
+        double res = tileEnvelope.getWidth() / extent;
+        double resHalf = res / 2;
+
+        double tx = tileEnvelope.getMinX();
+        double ty = tileEnvelope.getMaxY();
+        double sx = (double) extent / tileEnvelope.getWidth();
+        double sy = -((double) extent / tileEnvelope.getHeight());
```
ZakarFin: Maybe write some longer/more descriptive names for tx/sx/ty/sy.

service-mvt/src/main/java/org/oskari/service/mvt/ToMVTSpace.java
```diff
@@ -0,0 +1,101 @@
...
+    private final double sy;
+
+    public ToMVTSpace(double tx, double ty, double sx, double sy) {
+        this.tx = tx;
+        this.ty = ty;
+        this.sx = sx;
+        this.sy = sy;
+    }
+
+    public Geometry edit(Geometry geometry, GeometryFactory factory) {
```
ZakarFin: This returns a copy of the given geometry that is "cut" to fit in the tile "bbox"? And translates "proper" coordinates to pixel coordinates inside the tile?

control-mvt/src/main/java/org/oskari/control/mvt/GetWFSVectorTileHandler.java
jampukka: It only supports EPSG:3067 (KNOWN_TILE_GRIDS) so far

service-mvt/src/main/java/org/oskari/service/mvt/ToMVTSpace.java
jampukka: This translates the real-world coordinates of a geometry to the inner virtual-coordinate space of the tile. Inside the tile a coordinate consists of two integers (x,y) both in the range of [0,extent] (usually extent=4096).
Intersecting the geometry with the (real-world) spatial extent of the tile is a separate process, not connected to this operation.

control-mvt/src/main/java/org/oskari/control/mvt/GetWFSVectorTileHandler.java
jampukka: Nah, we just have to solve this in a smart way

PR #270, opened 2018-10-09
Remove TermsOfUseUserService

jampukka:
Not used anymore (in oskari-core at least)


ZakarFin:
True. There's PublishTermsOfUseServiceIbatisImpl that's used, but I guess this is some old generic tou service that's not used anymore.

ZakarFin:
Let's check if the table "portti_terms_of_use_user" is created automatically on startup for empty db. If it is we'll need to add a migration to remove it from the db.

ZakarFin:
Checked and the table is not created anymore (or it's removed as part of one of the previous migrations).

PR #271, opened 2018-10-10
In preparation of migrating permissions to id based

jampukka:
Step 1 of https://github.com/oskariorg/oskari-docs/issues/72

Work In Progress! May break existing functionality!

Add new `PermissionService` (currently used is called Permission**s**Service) with minimal functionality
Add a MyBatis implementation of newly created PermissionService.
Next step is to migrate (one at a time) users of PermissionsService to PermissionService while incrementally adding necessary functionality to PermissionService.

Refactor `Resource`:
- `id` is now int instead of long (reflect the current DB model)
- `type` can now be set with ResourceType (enum) for default types in addition to String (that is used for extensibility)
- Rewrite hasPermission(), removePermissionsOfType() functions in Java8 style
- Change package to `org.oskari.permissions.model`

Refactor `Permission`:
- `id` is now int instead of long (reflect the DB model)
- `type` can now be set with PermissionType (enum) for default types in addition to String (that is used for extensibility)
- `externalType` is now PermissionExternalType (enum) instead of String
- `externalId` is now int instead of String (references either Role.id (long in Java model but integer in DB model) or User.id (long in Java model but integer in DB model))
- Change package to `org.oskari.permissions.model`

Move `OskariLayerResource` to module oskari-permissions and package `org.oskari.permissions.model`

Remove un-used class `PermissionException`


control-base/src/main/java/fi/nls/oskari/control/data/CreateAnalysisLayerHandler.java
```diff
@@ -46,9 +43,9 @@
     private AnalysisDataService analysisDataService = new AnalysisDataService();
     private AnalysisWebProcessingService wpsService = new AnalysisWebProcessingService();
     private AnalysisParser analysisParser = new AnalysisParser();
-    private OskariLayerService mapLayerService = new OskariLayerServiceMybatisImpl();
+    private OskariLayerService mapLayerService = ServiceFactory.getMapLayerService();
 
-    private static PermissionsService permissionsService = new PermissionsServiceIbatisImpl();
+    private static PermissionService permissionsService = new PermissionServiceMybatisImpl();//ServiceFactory.getPermissionsService();
```
jampukka: Commented out part seems better

control-base/src/main/java/fi/nls/oskari/control/layer/GetAnalysisLayersHandler.java
```diff
@@ -29,34 +29,34 @@
 @OskariActionRoute("GetAnalysisLayers")
 public class GetAnalysisLayersHandler extends ActionHandler {
 
-    private static final Logger log = LogFactory.getLogger(GetAnalysisLayersHandler.class);
+    private static final Logger LOG = LogFactory.getLogger(GetAnalysisLayersHandler.class);
     private static final AnalysisDbService analysisService = new AnalysisDbServiceMybatisImpl();
-    private static PermissionsService permissionsService = new PermissionsServiceIbatisImpl();
+    private static PermissionService permissionsService = new PermissionServiceMybatisImpl();
```
jampukka: OskariComponentManager

control-base/src/main/java/fi/nls/oskari/control/layer/GetLayerIds.java
```diff
@@ -30,11 +30,8 @@
     private static final Logger log = LogFactory.getLogger(GetLayerIds.class);
 
     private static final String LAYER_IDS = "layerIds";
-    private static final String ID = "id";
-
-    private static final List<Integer> extra_layers = new ArrayList<Integer>();
 
-    private static PermissionsService permissionsService = new PermissionsServiceIbatisImpl();
+    private static final List<Integer> extra_layers = new ArrayList<>();
```
jampukka: This list could be non-static since it's modified in the init() function. If only one instance of this class is ever created then it makes no difference whether or not the field is static (it gets executed just once anyway), but if there's a chance another instance gets created then this solution no longer works as intended.

control-base/src/main/java/fi/nls/oskari/control/layer/GetMapLayerGroupsHandler.java
```diff
@@ -74,11 +63,8 @@ public void setLinkService(OskariLayerGroupLinkService linkService) {
     @Override
     public void init() {
         // setup services if they haven't been initialized
-        if (layerService == null) {
-            setLayerService(new OskariLayerServiceMybatisImpl());
-        }
         if (permissionsService == null) {
-            setPermissionsService(new PermissionsServiceIbatisImpl());
+            setPermissionsService(new PermissionServiceMybatisImpl());
```
jampukka: OskariComponentManager

control-base/src/main/java/fi/nls/oskari/control/view/modifier/bundle/MapfullHandler.java
```diff
@@ -86,6 +86,7 @@ public void init() {
         myPlaceService = OskariComponentManager.getComponentOfType(MyPlacesService.class);
         userLayerService = OskariComponentManager.getComponentOfType(UserLayerDbService.class);
         mapLayerService = OskariComponentManager.getComponentOfType(OskariLayerService.class);
+        permissionsService = new PermissionServiceMybatisImpl();
```
jampukka: OskariComponentManager

service-map/src/main/java/fi/mml/map/mapwindow/util/OskariLayerWorker.java
```diff
@@ -22,23 +14,26 @@
...
-
-    private static final String DOWNLOAD_PERMISSION_OK = "download_permission_ok";
+    public static final String PUBLICATION_PERMISSION_OK = "publication_permission_ok";
+    public static final String DOWNLOAD_PERMISSION_OK = "download_permission_ok";
 
     private static Logger log = LogFactory.getLogger(OskariLayerWorker.class);
 
     private static OskariLayerService mapLayerService = new OskariLayerServiceMybatisImpl();
-    private static PermissionsService permissionsService = new PermissionsServiceIbatisImpl();
+    private static PermissionService permissionService = new PermissionServiceMybatisImpl();
```
jampukka: ComponentManager

service-mybatis/src/main/java/fi/nls/oskari/mybatis/MyBatisHelper.java
```diff
@@ -15,6 +18,9 @@ public static SqlSessionFactory initMyBatis(DataSource ds, Class<?>... mappers)
     }
 
     public static Configuration getConfig(DataSource ds, Class<?>... mappers) {
+        if (ds == null) {
```
jampukka: Debug?

service-permissions/src/main/java/org/oskari/permissions/PermissionServiceMybatisImpl.java
```diff
@@ -0,0 +1,179 @@
...
+        } else {
+            factory = MyBatisHelper.initMyBatis(ds, MAPPER);
+        }
+    }
+
+
+    public List<Resource> findResourcesByUser(User user) {
+        try (SqlSession session = factory.openSession()) {
+            List<Resource> all = session.getMapper(MAPPER).findAll();
+            /*
```
jampukka: WIP? Nothing user specific here

service-permissions/src/main/java/org/oskari/permissions/PermissionServiceMybatisImpl.java
```diff
@@ -0,0 +1,179 @@
...
+            PermissionExternalType externalIdType,
+            String permissionsType) {
+
+        if (externalId == null || externalId.isEmpty()) {
+            return Collections.emptySet();
+        }
+
+        try (SqlSession session = factory.openSession(false)) {
+            ResourceMapper mapper = session.getMapper(MAPPER);
+            String commaseparatedIds = String.join(",", externalId.stream()
```
jampukka: Check out `Collectors.joining(",")`

service-permissions/src/main/java/org/oskari/permissions/PermissionServiceMybatisImpl.java
ZakarFin: Sure, I thought it better to have the API closer to what is needed but the query hasn't been optimized/result filtered yet.

ZakarFin: Actually I think having resourcetype as second param would benefit this as we can filter out unneeded resources.

service-mybatis/src/main/java/fi/nls/oskari/mybatis/MyBatisHelper.java
ZakarFin: Sure, could be taken out of this and put on another draft PR

service-map/src/main/java/fi/mml/map/mapwindow/util/OskariLayerWorker.java
ZakarFin: Using OskariComponentManager here breaks all the GetAppSetup*Tests. There's something funky going on with mocking stuff coming out of a ServiceLoader.

service-permissions/src/main/java/org/oskari/permissions/PermissionServiceMybatisImpl.java
jampukka: Sounds like a good idea

ZakarFin:
Still need more work but there's more than too much changes in this PR already so let's merge the changes so far and continue with more granular PRs. Still needs:
- Performance boost when loading resources (makes additional queries to get permissions -> needs join on sql and manual mapping)
- Also user-specific resources loading can be optimized.
- GetPermissionsLayerHandlers  and SaveLayerPermissionHandler still uses Ibatis implementation. Everything else has been migrated to MyBatis.
- The actual change from layer type+url+name string to an id int has not been done yet.

PR #273, opened 2018-10-16
Halt app startup on db migration failure

data-ux:
If an unchecked exception is thrown while running migrations, it will bubble up the stack all the way to oskariInit(). It is invoked with @PostConstruct annotation with which any unhandled exception will stop the app initialization.

Added ext.properties configuration `db.ignoreMigrationFailures` option to override behaviour.

### DIFF
--- a/service-webapp/src/main/java/fi/nls/oskari/servlet/WebappHelper.java
+++ b/service-webapp/src/main/java/fi/nls/oskari/servlet/WebappHelper.java
@@ -108,7 +108,7 @@ public class WebappHelper {
             log.warn("Skipping flyway migration! Remove 'db.flyway' property or set it to 'true' to enable migration");
             return;
         }
-        boolean ingoreMigrationFailures = PropertyUtil.getOptional("db.ingoreMigrationFailures", false);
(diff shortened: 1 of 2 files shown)


service-webapp/src/main/java/fi/nls/oskari/servlet/WebappHelper.java
```diff
@@ -108,13 +108,18 @@ private static void migrateDB() {
             log.warn("Skipping flyway migration! Remove 'db.flyway' property or set it to 'true' to enable migration");
             return;
         }
+        boolean ingoreMigrationFailures = PropertyUtil.getOptional("db.ingoreMigrationFailures", false);
```
ZakarFin: Lets fix ingore -> ignore on naming. Otherwise looks good.

ZakarFin:
Changes behavior on startup: any failed Flyway-module will stop the application from starting. 

To continue using migrations as before add `db.ignoreMigrationFailures=true` to oskari-ext.properties

ZakarFin:
Changes behavior on startup: any failed Flyway-module will stop the application from starting. 

To continue using migrations as before add `db.ignoreMigrationFailures=true` to oskari-ext.properties

PR #275, opened 2018-10-22
Update spring, spring-security and jackson

ZakarFin:
Upgrade new patch-versions for Spring/Jackson.


jampukka:
Hope it doesn't break anything :) Guess only time will tell, can always roll back to old version.

PR #278, opened 2018-10-26
CometD 4.x changes for Jetty 9

ZakarFin:
There are some not-proud-to-be-contributor parts on these changes where params map is being written to string as JSON just to be parsed again, but message.getJSON() returns null for CometD 4.x and the whole param parsing stuff should be changed to use the params map instead of partial JSON parsing of the same thing. The previous version didn't really make sense either since both the raw JSON and the preparsed params were passed to some of the methods...

Anyway, there's still a problem with writing the SessionStore to Redis (there's some GeoTools classes that refuse to be serialized) that's preventing clicks and highlights from working, The same problem is in the CometD 3.1.4 version on oskari-server/jetty-9 branch.

These changes require the oskari-frontend/jetty-9 branch modifications (namely the cometd-client update).


jampukka:
Wish we could map the data JSON object POJO classes representing the requests.

PR #280, opened 2018-10-30
Errorhandling for xalan in classpath

ZakarFin:
Fixes an issue where setting security features for XML-parsing are not supported by the implementation. 

TransformerFactory transformerFactory = TransformerFactory.newInstance(); returns implementation from the Xalan-library instead of the default one IF Xalan is part of the classpath. You can replace Xalan in the sentence with anything that provides a custom implementation for TransformerFactory. This happens for example if the SAML-support is enabled on Oskari.

From user perspective for example the analysis functionality didn't work when SAML was enabled and there was no clear error message on the log why this happened (the exception was not logged). Now when you get hit by this the logs shows:

`2018-10-30 17:54:04,369 WARN fi.nls.oskari.util.XmlHelper.newTransformerFactory(XmlHelper.java:240) - Unable to disable external DTD and stylesheets for XML parsing. Transformer class impl is org.apache.xalan.processor.TransformerFactoryImpl . Error was: Not supported: http://javax.xml.XMLConstants/property/accessExternalDTD`

And we settle for what ever this means for the implementation:

            transformerFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);

### DIFF
--- a/service-base/src/main/java/fi/nls/oskari/util/XmlHelper.java
+++ b/service-base/src/main/java/fi/nls/oskari/util/XmlHelper.java
@@ -237,8 +237,8 @@ public class XmlHelper {
             // Fox example Xalan is providing a custom TransformerFactory which doesn't support this so having it in the
             // classpath will give you this error and getting the actual impl class name is a huge win for debugging the reason.
             // You can check which dependency brings for example Xalan to classpath by running "mvn dependency:tree"
-            LOGGER.error(e, "Unable to disable external DTD and stylesheets for XML parsing. Transformer class impl is",
-                    transformerFactory.getClass().getCanonicalName());
+            LOGGER.warn("Unable to disable external DTD and stylesheets for XML parsing. Transformer class impl is",
+                    transformerFactory.getClass().getCanonicalName(), ". Error was:", e.getMessage());
         }
         // Disable resolving of any kind of URIs, not sure if this is actually necessary
         transformerFactory.setURIResolver((String href, String base) -> null);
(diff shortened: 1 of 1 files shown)


service-base/src/main/java/fi/nls/oskari/util/XmlHelper.java
```diff
@@ -224,11 +224,22 @@ public static TransformerFactory newTransformerFactory() throws TransformerFacto
...
+            transformerFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+            transformerFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_STYLESHEET, "");
+        } catch (Exception e) {
+            // https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet#JAXP_DocumentBuilderFactory.2C_SAXParserFactory_and_DOM4J
+            // https://stackoverflow.com/questions/27128578/set-feature-accessexternaldtd-in-transformerfactory#29021326
+
+            // Fox example Xalan is providing a custom TransformerFactory which doesn't support this so having it in the
+            // classpath will give you this error and getting the actual impl class name is a huge win for debugging the reason.
+            // You can check which dependency brings for example Xalan to classpath by running "mvn dependency:tree"
+            LOGGER.error(e, "Unable to disable external DTD and stylesheets for XML parsing. Transformer class impl is",
```
ZakarFin: Maybe this should be a warning as well?

PR #282, opened 2018-11-09
Publisher / appsetup saving fix

ZakarFin:
The startup information for bundles is no longer stored on the database with 1.49.0/Webpack changes. The database now has null-value constraint for the field in portti_view_bundle_seq and portti_bundle tables. Publisher and view saving functionalities still try to insert startup json which makes them fail the server operations. This fixes the issue by removing any startup references from insert/update SQLs.


jampukka:
OK

PR #296, opened 2018-11-29
Language handling for statistical indicators

ZakarFin:
Indicator.getName() now throws an exception if there was no name for the requested OR default language.


jampukka:
Looks good

PR #299, opened 2018-12-07
Improvements to MVT-WFS module

jampukka:
Currently Polygons that fully contain the tile extent get ignored. This fixes that.
Also add the id as a property to the MVT tiles. We don't use the actual 'id' field of the mvt.Feature because that only allows integer values and in many cases the id is a string prefixed integer or an uuid presented as a string.


ZakarFin:
Nice, I'll merge this after the 1.50.0 release.

PR #300, opened 2018-12-19
Manage Hystrix dependency

jampukka:
Currently multiple different versions are in use and none of them is managed by the parent pom.


ZakarFin:
Good!

PR #301, opened 2018-12-19
Userlayer kml parser

okauppinen:
Changed KML parser to use pullparser which parser only KML Placemarks (A Placemark is a Feature with associated Geometry). Other Features (Overlays, NetworkLinks, gx:Tour) are ignored. From Placemarks only name, description and untyped extended data (user data) are parsed to features' properties. Coordinates are transformed to Oskari native srs (default: 4326) which can be defined in oskari.properties. 

Added localization for fields object because name and description is often localized (and GML has also name property which is excluded from results). Maybe localization is too complex and label or title is better. Next improvement would be that user can edit localization/label/title.

Changed GeoJSONWriter to write null property with empty string instead of skipping property because UserLayerProcessor assumes that layer's all features have same properties.

Localized fields requires frontend changes.


oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONWriter.java
```diff
@@ -86,7 +86,8 @@ public JSONObject writeFeature(SimpleFeature f)
             }
             Object value = p.getValue();
             if (value == null) {
-                continue;
+                if (value instanceof Geometry) continue;
```
jampukka: Having `"key": null` would be better, but I'm not sure JSONObject serializes null values to anything, would have to check.

service-userlayer/src/main/java/org/oskari/map/userlayer/input/KMLParser.java
```diff
@@ -28,24 +40,126 @@
...
+            SimpleFeatureIterator iter = fc_kml.features();
+            while (iter.hasNext()){
+                f = iter.next();
+                builder.set(KML_NAME, f.getAttribute("name"));
+                // Basic builder
+                if (extendedData.isEmpty()){
+                    builder.set(KML_DESC, f.getAttribute("description"));
+                // Extended builder
+                } else {
+                    Map<String, Object> untypedData = (Map<String, Object>) f.getUserData().get("UntypedExtendedData");
```
jampukka: This is a bit difficult to follow. Maybe add a comment explaining _why_ this is done?

service-userlayer/src/main/java/org/oskari/map/userlayer/input/KMLParser.java
```diff
@@ -28,24 +40,126 @@
...
+        return new SimpleFeatureBuilder(builder.buildFeatureType());
+    }
+    private SimpleFeatureBuilder getExtendedBuilder (CoordinateReferenceSystem targetCRS, Set<String> extendedData) {
+        SimpleFeatureTypeBuilder builder = new SimpleFeatureTypeBuilder();
+        builder.setName( "ExtendedKMLBuilder" );
+        builder.setNamespaceURI( "http://www.oskari.org" );
+        builder.setCRS(targetCRS);
+        builder.add(KML_NAME, String.class );
+        builder.add(KML_DESC, String.class );
+        extendedData.forEach((k) -> {
```
ZakarFin: Can these be combined to a single method with just checking if extendedData is null -> don't do the loop? Does the builder name matter? Looks like the method's are the same other than that.

service-userlayer/src/main/java/org/oskari/map/userlayer/input/KMLParser.java
ZakarFin: Looks like it just tries to dig up a description from multiple places? Does calling set(KML_DESC, [something]) several times trigger an exception? If not this could be simplified by always doing:

    builder.set(KML_NAME, f.getAttribute("name")); 
    builder.set(KML_DESC, f.getAttribute("description"));

and overriding the desc value if there's a description found on the extended data? It's a bit confusing that untypedData is checked for non-null -> extendedData is looped, but the value is fetched from untypedData with the key from extendedData. Are the keys in untypedData not in sync with extendedData? Maybe refactor this logic to a new method with some comment on it.

service-userlayer/src/main/java/org/oskari/map/userlayer/service/UserLayerDataService.java
```diff
@@ -86,8 +87,26 @@ private static String parseFields(SimpleFeatureType schema) {
...
             for (PropertyDescriptor type : types) {
                 JSONObject obj = new JSONObject();
-                obj.put("name", type.getName().getLocalPart());
+                String name = type.getName().getLocalPart();
+                obj.put("name", name);
                 obj.put("type", type.getType().getBinding().getSimpleName());
+                JSONObject locales = new JSONObject();
+                // KML handling
+                if (KMLParser.KML_NAME.equals(name)){
+                    locales.put("fi", "Nimi");
```
ZakarFin: Not a big fan about having localization in code, but not sure where to put it either.

PR #302, opened 2018-12-19
Userlayer error handling

okauppinen:
Added error codes and additional info to exceptions. When exception occurs and it falls towards action route handler additional info is added to exception. Added getSuffix method for parsers to notice that which parser fails. Ignored files are also listed to inform user if files were found under folder.

Requires frontend changes.


control-userlayer/src/main/java/org/oskari/control/userlayer/UserLayerHandlerHelper.java
```diff
@@ -23,5 +29,48 @@ public static UserLayer getUserLayer(UserLayerDbService service, ActionParameter
...
+        JSONHelper.putValue(obj, key, new JSONObject(map));
+    }
+    public static JSONObject addStringToErrorJSON (JSONObject obj, String key, String str) {
+        if (obj == null) {
+            obj = new JSONObject();
+        }
+        JSONHelper.putValue(obj, key, str);
+        return obj;
+    }
+    public static JSONArray parseStringSet (Set<String> set){
```
jampukka: This isn't required, can use `new JSONArray(Collection collection)` constructor for the same effect.

control-userlayer/src/main/java/org/oskari/control/userlayer/UserLayerHandlerHelper.java
```diff
@@ -23,5 +29,48 @@ public static UserLayer getUserLayer(UserLayerDbService service, ActionParameter
...
+        if (obj == null) {
+            obj = new JSONObject();
+        }
+        JSONHelper.put(obj, key, parseStringSet(set));
+    }
+    public static void addMapToErrorJSON (JSONObject obj, String key, Map<String,String> map) {
+        if(map.isEmpty()){
+            return;
+        }
+        if (obj == null) {
```
jampukka: This doesn't really make sense. If the obj is null this path has no effect ouside of this function (only the local reference to `obj` changes)

control-userlayer/src/main/java/org/oskari/control/userlayer/UserLayerHandlerHelper.java
```diff
@@ -23,5 +29,48 @@ public static UserLayer getUserLayer(UserLayerDbService service, ActionParameter
...
+        return obj;
+    }
+    public static JSONObject createErrorJSON (String error) {
+        return JSONHelper.createJSONObject("error", error);
+    }
+    public static void addSetToErrorJSON (JSONObject obj, String key, Set<String> set) {
+        if (set.isEmpty()){
+            return;
+        }
+        if (obj == null) {
```
jampukka: This doesn't really make sense. When obj is null nothing changes outside of this function (only the local reference to obj changes)

jampukka:
There are a lot of different error codes that aren't managed anywhere. Should add an enum with the possible error codes and use those instead of just `"epsg_decode"` etc.

service-base/src/main/java/fi/nls/oskari/service/ServiceException.java
```diff
@@ -1,15 +1,23 @@
 package fi.nls.oskari.service;
 
+import org.json.JSONObject;
+
 public class ServiceException extends Exception {
-    
-	private static final long serialVersionUID = -3303389634376753825L;
+    private JSONObject options;
```
jampukka: This change feels a bit much. Maybe instead add an `UserLayerException` that extends ServiceException?

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
```diff
@@ -235,9 +258,12 @@ private void checkZipContainsExactlyOneMainFile(Set<String> extensions) throws A
                 .limit(2)
                 .count();
         if (mainFileExtensions == 0) {
-            throw new ActionParamsException("Couldn't find valid file for import in zip file");
+            throw new ActionParamsException("Couldn't find valid file for import in zip file", UserLayerHandlerHelper.createErrorJSON("no_main_file"));
         } else if (mainFileExtensions == 2) {
-            throw new ActionParamsException("Found too many valid files for import in zip file");
+            Set <String> mainExtensions = extensions.stream()
+                    .filter(FeatureCollectionParsers::hasByFileExt)
+                    .collect(Collectors.toCollection(HashSet::new));
```
jampukka: Or just `Collectors.toSet()` since we don't care about the implementation

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
```diff
@@ -346,12 +377,8 @@ private static String getFileExt(String name) {
...
     private SimpleFeatureCollection parse(FeatureCollectionParser parser, File file,
             CoordinateReferenceSystem sourceCRS,
-            CoordinateReferenceSystem targetCRS) throws ActionException {
-        try {
-            return parser.parse(file, sourceCRS, targetCRS);
-        } catch (ServiceException e) {
-            throw new ActionException("Failed to parse Feature Collection", e);
-        }
+            CoordinateReferenceSystem targetCRS) throws ServiceException {
+        return parser.parse(file, sourceCRS, targetCRS);
```
jampukka: This function isn't needed anymore since it only transformed the ServiceException into a ActionException

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
```diff
@@ -252,13 +278,18 @@ private void checkZipContainsExactlyOneMainFile(Set<String> extensions) throws A
...
+            CoordinateReferenceSystem targetCRS) throws ActionParamsException {
         File dir = null;
+        FeatureCollectionParser parser = null;
         try {
             dir = makeRandomTempDirectory();
             File mainFile = unZip(zipFile, cs, validFiles, dir);
-            FeatureCollectionParser parser = getParser(mainFile);
+            parser = getParser(mainFile);
             return parse(parser, mainFile, sourceCRS, targetCRS);
+        } catch (ServiceException e){
```
jampukka: This catch block is catching potential exceptions from makeRandomTempDirectory() and unZip() that have nothing to do with `"parser_error"`s. Should add a nested try block that calls `parser.parse()` and catch that ServiceException.

control-userlayer/src/main/java/org/oskari/control/userlayer/CreateUserLayerHandler.java
```diff
@@ -360,19 +387,19 @@ private FeatureCollectionParser getParser(File mainFile) {
...
 
     private UserLayer store(SimpleFeatureCollection fc, String uuid, Map<String, String> formParams)
-            throws ActionException {
+            throws ActionParamsException {
         try {
             UserLayer userLayer = createUserLayer(fc, uuid, formParams);
             UserLayerStyle userLayerStyle = createUserLayerStyle(formParams);
             List<UserLayerData> userLayerDataList = UserLayerDataService.createUserLayerData(fc, uuid);
-            userLayer.setFeatures_count(fc.size());
+            userLayer.setFeatures_count(userLayerDataList.size());
```
jampukka: Nice catch

service-userlayer/src/main/java/org/oskari/map/userlayer/input/MIFParser.java
```diff
@@ -40,6 +41,9 @@ public SimpleFeatureCollection parse(File file, CoordinateReferenceSystem source
             if (crs != null) {
                 sourceCRS = crs;
             }
+            if (sourceCRS == null) {
+                throw new ServiceException("Failed to parse MIF", JSONHelper.createJSONObject("cause", "no_source_crs"));
```
jampukka: Maybe move this check into `FeatureCollectionParser.read` function? It already checks that sourceCRS can not be null inside getTransform() but could also handle the invalid/unkown crs case.

control-userlayer/src/main/java/org/oskari/control/userlayer/UserLayerHandlerHelper.java
```diff
@@ -23,5 +29,48 @@ public static UserLayer getUserLayer(UserLayerDbService service, ActionParameter
         }
         return userLayer;
     }
-
+    public static JSONObject createErrorJSON (String error, String key, Set<String> set) {
```
jampukka: All these JSON helper functions here seem a bit off. Maybe consider an `UserLayerException` class that extends ServiceException and then move all the functionality in these functions to methods inside that exception class.

jampukka:
Looks good

PR #308, opened 2019-01-16
Maplayer coverage data handling

okauppinen:
Parse layer's coverage/bbox from actual layer (WMS 1.3.0 + capabilities). Use WKTHelper to create wkt.

Move wkt/geom transform to layerjsonformatter. Remove wkt/geom handling from userlayer. Userlayer formatter adds wkt to layer before getbasejson parsing so no need for own wkt handling.

Add possibility to ignore adding coverage to layerjson.
layer attributes: {"ignoreCoverage": true}


jampukka:
Looks good

PR #311, opened 2019-01-21
Refactor MVT WFS module

jampukka:
More clearly separate the parts of code that work with MVT tiles and the ones that act as a WFS client.
Separate the caching aspect of the WFS client by creating an extended implementation, `CachingWFSClient` on top of `WFS110Client`.

Add a workaround WFS services that return numberOfFeatures="unknown" (or wfs:numberOfFeatures="unknown") which otherwise leads to parsing errors.
Also add support for WFS services that require username/password when accessing schema files (DescribeFeatureType or other).
And finally check and follow redirects when trying to parse WFS responses.


control-mvt/src/main/java/org/oskari/control/mvt/GetWFSVectorTileHandler.java
```diff
@@ -61,106 +62,32 @@ public void init() {
...
-        byte[] encoded = SimpleFeaturesMVTEncoder.encodeToByteArray(sfc, layer.getName(), bbox, 4096, 256);
-
-        byte[] gzip = null;
+        final int layerId = params.getRequiredParamInt(ActionConstants.PARAM_ID);
+        final String srs = params.getRequiredParam(ActionConstants.PARAM_SRS);
+        final int z = params.getRequiredParamInt(PARAM_Z);
+        final int x = params.getRequiredParamInt(PARAM_X);
+        final int y = params.getRequiredParamInt(PARAM_Y);
+
+        final OskariLayer layer = findLayer(layerId);
```
ZakarFin: Add a credentials check here to know if the user has permission to view the layer?

jampukka: Will do

ZakarFin:
Nice :+1:

PR #313, opened 2019-01-23
 Add GetWFSFeatureHandler

jampukka:
For fetching features from WFS layers as GeoJSON FeatureCollection.
Currently only supports only WebMercator.

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
@@ -1,0 +1,228 @@
+package fi.nls.oskari.control.feature;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
+import org.geotools.data.simple.SimpleFeatureCollection;
+import org.geotools.data.simple.SimpleFeatureIterator;
+import org.geotools.feature.DefaultFeatureCollection;
+import org.geotools.feature.simple.SimpleFeatureBuilder;
+import org.geotools.feature.simple.SimpleFeatureTypeBuilder;
+import org.geotools.geojson.feature.FeatureJSON;
+import org.geotools.geometry.jts.JTS;
+import org.geotools.referencing.CRS;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.opengis.geometry.DirectPosition;
+import org.opengis.geometry.MismatchedDimensionException;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.opengis.referencing.operation.MathTransform;
+import org.opengis.referencing.operation.TransformException;
+import org.oskari.service.util.ServiceFactory;
+import org.oskari.service.wfs.client.OskariWFS110Client;
+
+import com.vividsolutions.jts.geom.Envelope;
+import com.vividsolutions.jts.geom.Geometry;
+
+import fi.nls.oskari.annotation.OskariActionRoute;
+import fi.nls.oskari.control.ActionConstants;
+import fi.nls.oskari.control.ActionException;
+import fi.nls.oskari.control.ActionHandler;
+import fi.nls.oskari.control.ActionParameters;
+import fi.nls.oskari.control.ActionParamsException;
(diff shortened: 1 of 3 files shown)


control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
```diff
@@ -0,0 +1,236 @@
...
+
+    protected SimpleFeatureCollection getFeatures(OskariLayer layer, Envelope bbox,
+            String requestSrsName, String targetSrsName) throws ActionException {
+        CoordinateReferenceSystem requestCRS;
+        CoordinateReferenceSystem targetCRS;
+        try {
+            requestCRS = CRS.decode(requestSrsName, true);
+            targetCRS = CRS.decode(targetSrsName, true);
+        } catch (Exception e) {
+            throw new ActionException(ERR_CRS_DECODE_FAIL);
```
ZakarFin: This should be ActionParamsException since we don't pass the cause and either way that might not be too interesting.

ZakarFin: ^Just to prevent excessive logging.

control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
```diff
@@ -0,0 +1,236 @@
...
+
+        MathTransform transformFromRequested;
+        MathTransform transformToRequested;
+        Envelope requestBbox;
+        if (shouldTransform) {
+            try {
+                transformFromRequested = CRS.findMathTransform(requestCRS, targetCRS, true);
+                transformToRequested = CRS.findMathTransform(targetCRS, requestCRS, true);
+            } catch (Exception e) {
+                throw new ActionException(ERR_TRANSFORM_FIND_FAIL);
```
ZakarFin: Use ActionParamsException here as well

control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
```diff
@@ -0,0 +1,236 @@
...
+            }
+            try {
+                requestBbox = JTS.transform(bbox, transformToRequested);
+            } catch (Exception e) {
+                throw new ActionException(ERR_REPOJECTION_FAIL);
```
ZakarFin: Pass the original exception as cause.

control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
```diff
@@ -0,0 +1,236 @@
...
+            transformToRequested = null;
+            requestBbox = bbox;
+        }
+
+        SimpleFeatureCollection features = getFeatures(layer, requestSrsName, requestBbox);
+        if (shouldTransform) {
+            try {
+                features = transform(features, targetCRS, transformFromRequested);
+            } catch (Exception e) {
+                throw new ActionException(ERR_REPOJECTION_FAIL);
```
ZakarFin: Pass the message from the cause exception or the exception itself.

control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
```diff
@@ -0,0 +1,236 @@
...
+        return true;
+    }
+
+    private SimpleFeatureCollection getFeatures(OskariLayer layer, String srsName, Envelope bbox) {
+        String endPoint = layer.getUrl();
+        String typeName = layer.getName();
+        String user = layer.getUsername();
+        String pass = layer.getPassword();
+        // TODO: Figure out the maxFeatures from the layer 
+        int maxFeatures = 10000;
```
ZakarFin: This is stored per layer in the old wfs-tables. Perhaps this could be stored in layer.getAttributes()?

ZakarFin:
Nice :+1:

control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
jampukka: Yes, that would be nice.

PR #314, opened 2019-01-23
Add oskari style support to UserLayerStyle

atlain:
Adds oskari style support to UserLayerStyle


atlain:
> Looks good. Unit test asserting `parseUserLayerStyleToOskariJSON` and `populateFromOskariJSON` are mirrorable would be great, lots of error-prone getting and setting going on.

Yep that's true, I'll add them.

PR #317, opened 2019-01-24
Action route for getting localized WFS feature property names

jheijari:
Added a new action handler for requesting localized property names for wfs layers


ZakarFin:
We need to refactor the database where these are stored, but the response from this looks good and we can refactor the implementation later.

PR #319, opened 2019-01-25
Fix NPE in GetWFSFeaturesHandler

jampukka:
Fix issue: https://github.com/oskariorg/oskari-server/pull/313#discussion_r250887441


ZakarFin:
Build "fails" because Sonar couldn't be contacted -> merging

PR #325, opened 2019-02-07
Prevent excessive logging

ZakarFin:
With an instance having many map layers AND log level "info" the log is spammed with messages like:

      INFO  fi.nls.oskari.util.JSONHelper - Couldn't get JSONArray from {} with key = forcedSRS
      INFO  fi.nls.oskari.util.JSONHelper - Couldn't get JSONArray from {"formats":{"available":["text/html","text/plain","application/vnd.ogc.gml"],"value":"text/html"},"isQueryable":true,...,"version":"1.3.0"} with key = srs

This helps with that.


jampukka:
OK

PR #326, opened 2019-02-12
Initial WFS 3 Core client

jampukka:
Initial version of a client for services adhering to WFS 3.0 Core, First Draft Release (https://cdn.rawgit.com/opengeospatial/WFS_FES/3.0.0-draft.1/docs/17-069.html). Includes support for reading paginated responses (via `next` links).

Changed the way `CachingWFSClient` works: it no longer extends `OskariWFS110Client` but acts as a pure cache in front of a new `OskariWFSClient` which in turn just proxies the calls to `OskariWFS110Client` or `OskariWFS3Client` based on the version parameter provided.

Moved the follow-redirection logic from `OskariWFS110Client` to `OskariWFSHttpUtil` and used the same code in`OskariWFS3Client`.

Modified both `GetWFSVectorTile` and `GetWFSFeatures` to use CachingWFSClient (and by doing so they now support WFS 3 layers).


control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
```diff
@@ -54,44 +49,60 @@
...
     }
 
     @Override
     public void handleAction(ActionParameters params) throws ActionException {
         int layerId = params.getRequiredParamInt(ActionConstants.PARAM_ID);
-        Envelope bbox = parseBbox(params.getRequiredParam(PARAM_BBOX));
+        String bboxStr = params.getRequiredParam(PARAM_BBOX);
         OskariLayer layer = findLayer(layerId, params.getUser());
-        // Only support WebMercator for now
-        // String targetSrs = params.getRequiredParam(ActionConstants.PARAM_SRS);
```
ZakarFin: Lets keep the param option to be flexible and default to 3857 if not provided.

service-wfs-client/src/main/java/org/oskari/service/wfs/client/CoordinateTransformer.java
```diff
@@ -0,0 +1,77 @@
...
+import org.opengis.referencing.NoSuchAuthorityCodeException;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.opengis.referencing.operation.MathTransform;
+import org.opengis.referencing.operation.TransformException;
+
+import com.vividsolutions.jts.geom.Geometry;
+
+public class CoordinateTransformer {
+
+    private final CoordinateReferenceSystem from;
```
ZakarFin: Any opinion if we should add these to https://github.com/oskariorg/oskari-server/blob/master/service-map/src/main/java/fi/nls/oskari/map/geometry/ProjectionHelper.java or deprecate the helper or just have these as separate "helpers"?

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSHttpUtil.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.util.Collections;
+import java.util.Map;
+
+import fi.nls.oskari.log.LogFactory;
+import fi.nls.oskari.log.Logger;
+import fi.nls.oskari.util.IOHelper;
+
+public class OskariWFSHttpUtil {
```
ZakarFin: This doesn't really have anything WFS-specific. Add these to IOHelper so other functionalities benefit from the redirect following as well?

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSClient.java
```diff
@@ -0,0 +1,27 @@
...
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.oskari.service.wfs3.client.OskariWFS3Client;
+
+public class OskariWFSClient {
+
+    private static final String WFS_3_VERSION = "3.0.0";
+
+    private OskariWFSClient() {}
+
+    public static SimpleFeatureCollection tryGetFeatures(
```
ZakarFin: Any reason to have try at the start of the method name instead of just getFeatures? 

I'd really like this to work like OskariWFSClient.getVersion(ver).getFeatures(endPoint, user, pass, typeName, bbox, crs, maxFeatures) or maybe even something like OskariWFSClient.create(endPoint, user, pass, ver).useCache(cache).getFeatures(typeName, bbox, crs, maxFeatures), but this would mean making a lot of static methods non-static and I'm not sure if it's worth it. 

Maybe it would enable better control for the amount of requests sent to a specific endpoint and we might want to wrap it as a Hystrix command etc. But... maybe we can use this as is and worry about DOSing services in the next version :)

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSHttpUtil.java
jampukka: Sure

service-wfs-client/src/main/java/org/oskari/service/wfs/client/CoordinateTransformer.java
jampukka: Hard to say, this one is a bit different, since the "user" doesn't have to manage CoordinateReferenceSystems or MathTransforms in their code, but they still get reused.

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSClient.java
jampukka: The try-prefix is trying to denote that the function is trying to do something, and that it might not succeed in doing it, but it will not throw an Exception if it fails, but rather it handles the possible errors itself and returns null on failure (could also be Optional<T>).

This class is nothing but a dumb proxy, we could make this class more "client-y" pretty easily. This class could handle all those caching, request pooling etc. and we wouldn't have to touch OskariWFS110Client and OskariWFS3Client at all. Those could still be just a set of static functions. The caching and request pooling mechanisms should probably work the same way independent of the version of the underlying WFS services.

control-base/src/main/java/fi/nls/oskari/control/feature/GetWFSFeaturesHandler.java
jampukka: Changed that in https://github.com/oskariorg/oskari-server/pull/326/commits/20a5ac386b24a8ae2e57af061a0211032cecccbb

PR #329, opened 2019-02-15
Add own implementation of GeoJSON FeatureCollection parser to WFS3 client

jampukka:
GeoTools default GeoJSON parser doesn't:
- handle WFS 3 'links' array
- work with complex properties (other than geometries and "simple" arrays)
- breaks when properties comes before geometry and properties contains a geometry property
- heterogeneous FeatureCollections (where type might vary between features)

This one does support all of the above.


service-wfs-client/src/main/java/org/oskari/service/wfs/client/geojson/MapToGeoJSONFeature.java
```diff
@@ -0,0 +1,137 @@
...
+            Geometry geometry = null;
+            if (maybeGeometry != null) {
+                geometry = MapToGeoJSONGeometry.tryConvertToGeometry((Map<String, Object>) maybeGeometry);
+            }
+
+            Map<String, Object> properties = (Map<String, Object>) maybeFeature.get("properties");
+
+            return buildFeature(id, geometry, properties);
+        } catch (Exception ignore) {
+            ignore.printStackTrace();
```
ZakarFin: Remove stack trace or use logger.debug or something if we are not interested.

service-wfs-client/src/main/java/org/oskari/service/wfs/client/geojson/WFS3FeatureCollectionIterator.java
```diff
@@ -52,8 +48,9 @@ public SimpleFeature next() {
     SimpleFeature readNext() {
         try {
             parser.parse(reader, handler, true);
-            return handler.getValue();
+            return handler.getFeature();
         } catch (Exception e) {
+            e.printStackTrace();
```
ZakarFin: Remove stack trace as exception is thrown wrapping the cause.

service-wfs-client/src/test/java/org/oskari/service/wfs/client/OskariWFS3ClientTest.java
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    public void testGetFeatures() throws ServiceException, IOException {
+        String endPoint = "https://beta-paikkatieto.maanmittauslaitos.fi/geographic-names/wfs3/v1/";
+        String user = null;
+        String pass = null;
+        String collectionId = "places";
+        CoordinateReferenceSystem crs = OskariWFS3Client.getCRS84();
+        Envelope envelope = new Envelope(21.35, 21.40, 61.35, 61.40);
+        ReferencedEnvelope bbox = new ReferencedEnvelope(envelope, crs);
+        int limit = 10;
```
ZakarFin: Limit is more like page size since i is expected to be 29?

service-wfs-client/src/test/java/org/oskari/service/wfs/client/geojson/WFS3FeatureCollectionIteratorTest.java
```diff
@@ -0,0 +1,139 @@
...
+            assertEquals("self", link.getRel());
+            assertEquals("application/geo+json", link.getType());
+            assertEquals("This document", link.getTitle());
+        }
+    }
+    
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testDeeplyComplex() throws IOException, ParseException {
+        // We delved too greedily and too deep
```
ZakarFin: Loving the comment :D

ZakarFin:
Good job, couple of stack traces to remove to avoid unnecessary logging and ready for merging.

service-wfs-client/src/test/java/org/oskari/service/wfs/client/OskariWFS3ClientTest.java
jampukka: Yea limit is the upper limit for a single request. If there's a link to next page we will follow it. Maybe we another, hard limit for the number of features we want to read (probably we want to stop before reading billions of features). Or maybe we should let the frontend to decide, and paginate all the way to the frontend, that could be interesting. Obviously for MVT tiles that's not what we want, but for GeoJSON maybe.

PR #334, opened 2019-03-18
Add support for MyPlaces in GetWFSFeatures and in GetWFSVectorTile

jampukka:
The `id` parameter now accepts Strings instead of only integers in `GetWFSFeatures` and `GetWFSVectorTile` action routes. If the `id` parameter starts with `myplaces_` prefix then the part after the prefix is handled as the id of the MyPlaces category. Otherwise it is handled as previously (expecting the maplayer id as an integer).
To make this work `WFS110Client` now accepts a `Filter` that can be used for the request. If the `Filter` is set then it is used instead of the `ReferencedEnvelope bbox` when constructing the GetFeature request (so be sure to include the bounding box filter in the actual Filter if you want one to be used).
Change `ComputeOnceCache` to extend `Cache` so that the `CacheManager` can actually manage them. Also refactor how `CacheManager` works both internally and also externally by adding a `getCache(String key, Supplier<Cache>)` function to the class.

### DIFF
--- a/service-base/src/main/java/fi/nls/oskari/cache/CacheManager.java
+++ b/service-base/src/main/java/fi/nls/oskari/cache/CacheManager.java
@@ -3,6 +3,7 @@ package fi.nls.oskari.cache;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.function.Supplier;
 
 /**
  * Generic cache factory for Oskari.
@@ -37,13 +38,17 @@ public class CacheManager {
      * @return
      */
     public static <T> Cache<T> getCache(final String name) {
+        return getCache(name, () -> new Cache<>());
+    }
+
+    public static <T> Cache<T> getCache(final String name, final Supplier<? extends Cache<T>> supplier) {
         final Cache existing = CACHE_STORE.get(name);
         if (existing != null) {
             return existing;
         }
 
         // create a new one
-        final Cache<T> cache = new Cache<T>();
+        final Cache<T> cache = supplier.get();
         cache.setName(name);
         CACHE_STORE.put(name, cache);
         return cache;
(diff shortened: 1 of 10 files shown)


service-base/src/main/java/fi/nls/oskari/cache/CacheManager.java
```diff
@@ -37,16 +39,27 @@
...
+     * Returns a cache matching name or creates one if it doesn't exist
+     * by calling the provided supplier function.
+     *
+     * @param name name of the cache
+     * @param supplier function which creates a new cache if one doesn't exists for the key
+     * @param <T>  type mapping for cache
+     * @return
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> Cache<T> getCache(final String name, final Supplier<? extends Cache<T>> supplier) {
```
ZakarFin: Can we use something like <? extends Cache> as return value to get rid of some nasty casting when using this?

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFS110Client.java
```diff
@@ -37,17 +41,17 @@ private OskariWFS110Client() {}
      * @return SimpleFeatureCollection containing the parsed Features, or null if all fails
      */
     public static SimpleFeatureCollection tryGetFeatures(String endPoint, String user, String pass,
-            String typeName, ReferencedEnvelope bbox, CoordinateReferenceSystem crs, Integer maxFeatures) {
+            String typeName, ReferencedEnvelope bbox, CoordinateReferenceSystem crs, Integer maxFeatures, Filter filter) {
```
ZakarFin: Should we have an overloaded version of the method without the filter that would call this one with null value.

service-base/src/main/java/fi/nls/oskari/cache/CacheManager.java
jampukka: Updated, usage example now: https://github.com/oskariorg/oskari-server/blob/33db601e0e19b4746480bf5e8c9b96274da511b9/service-wfs-client/src/main/java/org/oskari/service/wfs/client/CachingWFSClient.java#L24-L25

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFS110Client.java
jampukka: Added

PR #337, opened 2019-03-21
Add new module service-wfs3

jampukka:
Add a new module `service-wfs3` which is used for working (as a client) with WFS 3 services that conform to the Core and the GeoJSON conformance classes of the WFS 3 draft spec.


service-wfs3/src/test/java/org/oskari/service/wfs3/WFS3ServiceTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.oskari.service.wfs3.model.WFS3CollectionInfo;
+import org.oskari.service.wfs3.model.WFS3Exception;
+
+public class WFS3ServiceTest {
+
+    @Test
+    @Ignore("Depends on outside service")
+    public void testJSONWorks() throws WFS3Exception, IOException {
+        String url = "https://beta-paikkatieto.maanmittauslaitos.fi/geographic-names/wfs3/v1/";
+        WFS3Service expected = WFS3Service.fromURL(url);
```
ZakarFin: Can we add a response as test resource so we can test parsing and "document" an expected response from server?

PR #341, opened 2019-03-27
Fix support for user generated content (myplaces etc) for the new WFS-client

ZakarFin:
New base class for GetWFSFeatures and GetWFSVectorTile action routes to remove duplicated code. 

PermissionHelper now returns layer without permisssions if it's tagged as "internal" like myplaces, userlayers, analysislayers are. 

Also requires layers to be updated in the database to reference the combined view of layer/feature to make filters work properly:

    UPDATE oskari_maplayer set name = 'oskari:my_places_categories' where name = 'oskari:my_places';
    UPDATE oskari_maplayer set name = 'oskari:user_layer_data_style' where name = 'oskari:vuser_layer_data';
    UPDATE oskari_maplayer set name = 'oskari:analysis_data_style' where name = 'oskari:analysis_data';

I'll make a flyway update of this when this is ready.

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/feature/AbstractWFSFeaturesHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/feature/AbstractWFSFeaturesHandler.java
@@ -1,0 +1,153 @@
+package fi.nls.oskari.control.feature;
+
+import fi.nls.oskari.control.*;
+import fi.nls.oskari.control.layer.PermissionHelper;
+import fi.nls.oskari.domain.User;
+import fi.nls.oskari.domain.map.OskariLayer;
+import fi.nls.oskari.log.LogFactory;
+import fi.nls.oskari.log.Logger;
+import fi.nls.oskari.service.ServiceRuntimeException;
+import fi.nls.oskari.util.PropertyUtil;
+import org.geotools.data.simple.SimpleFeatureCollection;
+import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.referencing.CRS;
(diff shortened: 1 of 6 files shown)


control-base/src/main/java/fi/nls/oskari/control/feature/AbstractWFSFeaturesHandler.java
```diff
@@ -0,0 +1,153 @@
...
+            myPlacesHelper = new MyPlacesWFSHelper();
+        }
+        if (userlayerHelper == null) {
+            userlayerHelper = new UserLayerWFSHelper();
+        }
+        this.wfsClient = new CachingWFSClient();
+    }
+
+    @Override
+    public abstract void handleAction(ActionParameters params) throws ActionException;
```
jampukka: This is unnecessary, no?

ZakarFin:
Updating the layer names would break current implementation that works with the transport webapp so I'm not including Flyway-scripts on those for now. It also results in the features having tons of unnecessary attributes for mvt/geojson. So we either need to:

- refactor the "publisher_name" logic for "public" layers to get rid of the view as layer name OR
- update transport to work with the views OR 
- any other options?

Anyways this fixes the code for mvt/geojson producing action routes and doesn't break anything that's currently working. So this could be merged as is and let the instance admin decide which to use by either updating the db or not OR we can let this boil for a bit and figure out a solution that would work with both.

ZakarFin:
I commented out the publisher_name logic so guest-users can't see published myplaces/userlayers, but otherwise this works with the same layer names as transport now. We need to solve the permission issue separately.

PR #343, opened 2019-03-29
Add myplaces support for GetLocalizedPropertyNamesHandler

ZakarFin:
Use the same methods as geojson/mvt action route for getting localized attribute names on layers. Without this the GetLocalizedPropertyNamesHandler assumes id is number and fails on the user generated layers.


jampukka:
Looks good

PR #345, opened 2019-04-04
Limit amount of requests sent to service on MVT caching

ZakarFin:
Also changes the cache size from 100 to 10000 to accommodate more concurrent users. Previously the cache was overflowing on even a single request at the outer zoom levels.


jampukka:
Good job. The z=8 solution was done to actually _ reduce_ the amount of requests (which it does for quite succesfully for zooms levels >8). Vector tiles aren't _that_ useful for zoom levels < 7 so I like this solution. For feature layers with low feature count (thousands or less) and very large geometries using a bbox and GeoJSON makes more sense than vector tiles.

One thing to note is that the hard coding the zoom levels as we've done now will come back to bite us once someone tries it in WebMercator, but let's worry about that once we have a sane solution for EPSG:3067

PR #346, opened 2019-04-04
Add initial support for analysis layers in MVT/GeoJSON producing WFS-endpoint

ZakarFin:
Adds support for analysis layers and refactors myplaces/userlayer support so all of these implement a common interface 'UserLayerService'. Annotated the helpers as OskariComponents so the code can be placed in modules providing the functionality removing the hardcoded stuff from the WFS-endpoint.

The functionality as a whole still needs a bit development but the idea is that UserLayerService would also have a method for checking if the user is allowed to see the layer. Known changes that are required include:
- GeoJSON action route should not use CachingWFSclient so it doesn't fill out the cache as the bbox is always different
-  user content layers are only visible to the user even if they are published/used in embedded maps
- analysis features need to be retyped/postprocessed similar to userlayer features (the attribute names need to be fetched from the db)

MapfullBundleHandler and similar code with hardcoded handling could also benefit from UserLayerService. We might need to change the myplaces handling in it anyway since it gives produces WMS-layers for myplaces in embedded maps.

Note! Includes commits from #345.


ZakarFin:
Btw not sure if UserLayerService is the best of names since "UserLayer" is something that's already used for stuff the user imports from shapefiles etc. UserContentService might be better, but we already have that as well and it's in service-base that doesn't have the map related services/libraries available.

control-base/src/main/java/fi/nls/oskari/control/feature/AbstractWFSFeaturesHandler.java
```diff
@@ -44,13 +50,11 @@ public void init() {
                     ServiceFactory.getMapLayerService(),
                     ServiceFactory.getPermissionsService());
         };
-        if (myPlacesHelper == null) {
-            myPlacesHelper = new MyPlacesWFSHelper();
-        }
-        if (userlayerHelper == null) {
-            userlayerHelper = new UserLayerWFSHelper();
-        }
         this.wfsClient = new CachingWFSClient();
```
jampukka: _GeoJSON action route should not use CachingWFSclient so it doesn't fill out the cache as the bbox is always different_ 
First idea to pop into my head would be to create a WFSClient class with method tryGetFeatures(...) that the CachingWFSClient would then extend. Then we'd add a method `public WFSClient getWFSClient()` and call that here. The method would return a new CachingWFSClient by default and the GeoJSON action route could overload the method to return a non-caching version. Obviously this would also require changing the type of the field `wfsClient` from `CachingWFSClient` to `WFSClient`.

control-base/src/main/java/fi/nls/oskari/control/feature/AbstractWFSFeaturesHandler.java
```diff
@@ -127,19 +133,17 @@ protected SimpleFeatureCollection getFeatures(String id, String uuid, OskariLaye
...
-        } else if (userlayerHelper.isUserlayerLayer(layer)) {
-            int userlayerId = userlayerHelper.getUserlayerId(id);
-            filter = userlayerHelper.getFilter(userlayerId, uuid, bbox);
-        }
+        // Find out if we need custom filter or postProcessing for features
+        Optional<UserLayerService> processor =
+                userContentProcessors.stream()
+                .filter(proc -> proc.isUserContentLayer(id))
+                .findAny();
+        Filter filter = processor.map(proc-> proc.getWFSFilter(id, uuid, bbox)).orElse(null);
```
jampukka: Got bored of chaining method calls so the map and orElse calls are separate from the original chain? :)

jampukka:
Looks good. Agree that `UserLayerService` is poorly named :) We'll change it once we figure out a good replacement.

PR #350, opened 2019-04-08
Metadata search fix

ZakarFin:
The metadata search stopped working on 1.51.0. Looks like the cause is the Deegree dependency that uses older version from Log4J that we just upgraded.

We only use Deegree to write the query payload for CSW-service like GeoNetwork so it makes sense to not have the dependency and write our own XML-generator.

Also removes Village* files that are not used anymore even without the actual changes.


service-csw/src/main/java/org/oskari/csw/request/GetRecords.java
```diff
@@ -0,0 +1,154 @@
...
+        } catch (XMLStreamException e) {
+            throw new ServiceRuntimeException("Couldn't create GetRecords request", e);
+        }
+        filterFactory = CommonFactoryFinder.getFilterFactory();
+    }
+
+    public void setFilters(List<Filter> filters) {
+        this.filters = filters;
+    }
+    public void addEqualFilter(String name, String value) {
```
ZakarFin: This and the filterFactory will be removed from the class. setFilters is the way to go here since there's some configurable logic how the filters are joined so it's not something that we want in this class.

jampukka: If you're only going to use `setFilters(List<Filter> filters)` then consider `getXML(List<Filter> filters)` (or just `getXML(Filter filter)` with the caller making the decision how multiple Filters should be joined together (now they are ANDed together)

ZakarFin: True. With the removal of addXX() methods this class can just provide static method to get the XML.

service-csw/src/main/java/org/oskari/csw/request/GetRecords.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    /**
+     * Serialize stream to get a String
+     * @param stream
+     * @return
+     */
+    private static String getAsString(ByteArrayOutputStream stream) {
+        return new String (stream.toByteArray(), Charset.forName("UTF-8"));
```
jampukka: See `ByteArrayOutputStream#toString(StandardCharsets.UTF_8.name())`

jampukka: Better yet if we want to return a `String` from `createRequest(Filter)` we can use `StringWriter` instead of `ByteArrayOutputStream` here and avoid encoding and decoding chars to UTF-8 bytes and back.

jampukka: Or we can change the function header to `ByteArrayOutputStream createRequest(Filter, Charset)` and `MetadataCatalogueChannelSearchService` could then use `IOHelper.post(String url, String contentType,         ByteArrayOutputStream baos)`. This would be the most efficient way of doing it.
If in unit tests we do want Strings then we can easily create them there from the ByteArrayOutputStream.

jampukka: While this certainly isn't performance critical code I think we should still strive for (the most) efficient implementation throughout the codebase while maintaining high readability.
In my opinion passing around a ByteArrayOutputStream instead of a String would not make this code any less readable.

service-csw/src/main/java/org/oskari/csw/request/GetRecords.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Nothing fancy. Just wrapping possible exception to a runtime exception.
+     * @param stream
+     * @return
+     */
+    private static XMLStreamWriter getWriter(OutputStream stream) {
+        try {
+            return xof.createXMLStreamWriter(stream);
```
jampukka: Missing explicit charset information. Can pass a Writer instead of an OutputStream here.

PR #351, opened 2019-04-10
Added support numeric values

kuosman:
This PR adds support numeric values to Excel Export. 

Also coming front-end code change (Grid.js) but this also works without it.


ZakarFin:
Thanks!

PR #353, opened 2019-04-11
Route error handling

ZakarFin:
- Remove extra stacktrace on error
- Respond with 400 bad request instead of 501 not implemented when route throws ActionParamsException


jampukka:
Wonder if this breaks any frontend code expecting a 501?

ZakarFin:
Only the coordinatetool checks for 501 status code to internally mark that the route is not there to prevent it from being called again. This is fine since 501 is still returned if the route is not included in the oskari-server instance.

PR #355, opened 2019-04-12
Convert oskari style dasharray values

okauppinen:


ZakarFin:
Fixes an issue where line is always a dasharray after modifying the style (instead of the one user selected like "solid").

PR #358, opened 2019-04-17
WFS client refactoring

jampukka:
Remove streaming GeoJSON parser in service-wfs3 which was problematic to maintain and to extend.

Add `GeoJSONReader2` to `oskari-geojson` module that works with `Map<String, Object>`s compared to `GeoJSONReader` which uses `JSONObject`s and `JSONArray`s (with dependency to service-base and the custom version of org.json we have). `GeoJSONReader2` will eventually replace `GeoJSONReader` once we can refactor away current code using `GeoJSONReader`.

Add `GeoJSONSchemaDetector` to `oskari-geojson` module (also working with `Map<String, Object>`s). This class can be used to automatically detect the schema of a GeoJSON FeatureCollection. Also include some automatic widening principals, for example if some features have `Point` geomeries and some have `MultiPoint` we will use `MultiPoint` for the `FeatureType`.

Add `GeoJSONFeatureCollection`, a simple ArrayList based `SimpleFeatureCollection` implementation. Refactor bunch of code to use GeoJSONFeatureCollection instead of DefaultFeatureCollection.

Also other random refactoring and optimizations where easily available.


oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONReader2.java
```diff
@@ -0,0 +1,311 @@
...
+     * in the properties map that were eligible geometries
+     */
+    public static void replaceFeatureCollectionsMapsWithGeometries(Map<String, Object> featureCollection) {
+        String type = GeoJSONUtil.getString(featureCollection, GeoJSON.TYPE);
+        if (!GeoJSON.FEATURE_COLLECTION.equals(type)) {
+            if (GeoJSON.FEATURE.equals(type)) {
+                replaceFeaturesMapWithGeometries(featureCollection);
+                return;
+            }
+            throw new IllegalArgumentException("type was not " + GeoJSON.FEATURE_COLLECTION);
```
ZakarFin: Should the message be "Type was not GeoJSON.FEATURE" instead?

oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONReader2.java
```diff
@@ -0,0 +1,311 @@
...
+
+    /**
+     * Replaces geometries that are currently represented as Map<String, Object>
+     * with JTS Geometries in the map
+     * Replaces both the default geometries ('geometry' key) as well as entries
+     * in the properties map that were eligible geometries
+     */
+    public static void replaceFeatureCollectionsMapsWithGeometries(Map<String, Object> featureCollection) {
+        String type = GeoJSONUtil.getString(featureCollection, GeoJSON.TYPE);
+        if (!GeoJSON.FEATURE_COLLECTION.equals(type)) {
```
ZakarFin: All of this is trying to cope with invalid input for the method? So if the input isn't feature collection you try to parse it as a single feature instead? I think it would read better if you used try catch (which isn't really necessary either since you can just pass the exception from replaceFeaturesMapWithGeometries if not for the message difference):

```
if (!GeoJSON.FEATURE_COLLECTION.equals(type)) {
    // expected to get a feature collection, but lets try parsing as single feature instead
    try {
        replaceFeaturesMapWithGeometries(featureCollection);
    } catch(IllegalArgumentException e) {
        // wasn't a feature either -> can't deal with this
        throw new IllegalArgumentException("type was not " + GeoJSON.FEATURE_COLLECTION, e);
    }
}
```

Having if instead of try catch is probably faster, but I think this needs at least a comment since the method names are too similar to instantly figuring out what's happening.

oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONReader2.java
ZakarFin: I guess the message is correct. Just took me 5 minutes to see what's going on here. Hence the above comment :)

oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONReader2.java
```diff
@@ -0,0 +1,311 @@
...
+            replaceFeaturesMapWithGeometries(feature);
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    public static void replaceFeaturesMapWithGeometries(Map<String, Object> feature) {
+        String type = GeoJSONUtil.getString(feature, GeoJSON.TYPE);
+        if (!GeoJSON.FEATURE.equals(type)) {
+            if (GeoJSON.FEATURE_COLLECTION.equals(type)) {
+                replaceFeatureCollectionsMapsWithGeometries(feature);
```
ZakarFin: Same here. Add a comment in the lines of "didn't get a feature that was expected, trying to parse as collection"

ZakarFin: Maybe renaming the methods to replaceGeometriesInFeature() and replaceGeometriesInFeatureCollection() could do the trick as well.

oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONReader2.java
```diff
@@ -0,0 +1,311 @@
...
+     * Replaces geometries that are currently represented as Map<String, Object>
+     * with JTS Geometries in the map
+     * Replaces both the default geometries ('geometry' key) as well as entries
+     * in the properties map that were eligible geometries
+     */
+    public static void replaceFeatureCollectionsMapsWithGeometries(Map<String, Object> featureCollection) {
+        String type = GeoJSONUtil.getString(featureCollection, GeoJSON.TYPE);
+        if (!GeoJSON.FEATURE_COLLECTION.equals(type)) {
+            if (GeoJSON.FEATURE.equals(type)) {
+                replaceFeaturesMapWithGeometries(featureCollection);
```
ZakarFin: If not the above this could have a comment saying:

    // expected collection but got a feature -> redirecting to proper method.

oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONSchemaDetector.java
```diff
@@ -0,0 +1,183 @@
...
+import com.vividsolutions.jts.geom.MultiPoint;
+import com.vividsolutions.jts.geom.MultiPolygon;
+import com.vividsolutions.jts.geom.Point;
+import com.vividsolutions.jts.geom.Polygon;
+
+public class GeoJSONSchemaDetector {
+
+    @SuppressWarnings("unchecked")
+    public static SimpleFeatureType getSchema(Map<String, Object> json, CoordinateReferenceSystem crs) {
+        GeoJSONReader2.replaceMapsWithGeometries(json);
```
ZakarFin: This creates a side-effect by modifying the input as well as returning the SimpleFeatureType. That might cause problems later on.

PR #368, opened 2019-04-26
Fix GetWFSFeatures action route

ZakarFin:
Implement getBounds for GeoJSONFeatureCollection to fix FeatureJSON.writeFeatureCollection().


jampukka:
Great, thanks

PR #370, opened 2019-04-29
Usercontent permissions

ZakarFin:
User generated content requires customized permission checks for the new WFS-client. 

This PR will add these checks for:
- my places
- user layers
- analysis (with a temporary short cache on the helper to ease some of the load from db)


jampukka:
Great

PR #381, opened 2019-05-15
Thematic maps adapter for UN stats

ZakarFin:
Adapter for https://unstats.un.org/SDGAPI/swagger/. To use it you need a region set layer for thematic maps ("statslayer") that has countries with ISO-2 country codes as IDs for regions. Each goal is added as a separate datasource. 

The datasource config can have the following values:

- goal {String} (defaults to "1")
- url {String} (defaults to "https://unstats.un.org/SDGAPI/v1/sdg")
- timeVariable {String} (defaults to "timePeriod")
- "timePeriod" {int[]} (or the key referenced by "timeVariable". Defaults to 2000- (current year-2))

Example SQLs:
```
-- Goal 2
INSERT INTO oskari_statistical_datasource (locale, config, plugin)
VALUES ('{"en":{"name":"UN SDG goal 2: Zero hunger"}}', '{ "goal": "2" }', 'UNSD');

INSERT INTO oskari_statistical_layer(datasource_id, layer_id, config)
VALUES(
    (SELECT id FROM oskari_statistical_datasource
        WHERE locale like '%UN SDG goal 2: Zero hunger%'),
    (SELECT id FROM oskari_maplayer WHERE type='statslayer' AND name = 'ne_110m_countries'),
    '{}');

-- Goal 7
INSERT INTO oskari_statistical_datasource (locale, config, plugin)
VALUES ('{"en":{"name":"UN SDG goal 7: Affordable and clean energy"}}', '{ "goal": "7"} ', 'UNSD');

INSERT INTO oskari_statistical_layer(datasource_id, layer_id, config)
VALUES(
    (SELECT id FROM oskari_statistical_datasource
        WHERE locale like '%UN SDG goal 7: Affordable and clean energy%'),
    (SELECT id FROM oskari_maplayer WHERE type='statslayer' AND name = 'ne_110m_countries'),
    '{}');
```

You will also need to add the dependency to your app:

```
        <dependency>
            <groupId>org.oskari.service</groupId>
            <artifactId>oskari-statistics-unsd</artifactId>
            <version>${oskari.version}</version>
        </dependency>
```


service-statistics-unsd/src/main/java/fi/nls/oskari/control/statistics/plugins/unsd/parser/UnsdParser.java
```diff
@@ -0,0 +1,156 @@
...
+            return indicatorList;
+        }
+        try {
+            for (int i = 0; i < goals.length(); i++) {
+                JSONObject item = goals.getJSONObject(i);
+                parseIndicatorsResultItem(item)
+                        .ifPresent(list -> indicatorList.addAll(list));
+            }
+        } catch (JSONException e) {
+            LOG.error("Error parsing UNSD indicators: " + e.getMessage(), e);
```
jampukka: We break the loop by catching the exception, but if we've successfully parsed some `StatisticalIndicator`s before the exception we still return the list as if the response was OK.

service-statistics-unsd/src/main/java/fi/nls/oskari/control/statistics/plugins/unsd/UnsdIndicatorValuesFetcher.java
```diff
@@ -0,0 +1,81 @@
...
+        boolean allFetched = false;
+        String response;
+        while(!allFetched) {
+            try {
+                response = request.getIndicatorData(selectors);
+                result.putAll(parseIndicatorData(response));
+                allFetched = isLastPage(response);
+                if (!allFetched) {
+                    request.nextPage();
+                    response = request.getIndicatorData(selectors);
```
jampukka: `request.getIndicatorData(selectors)` is called twice per loop?

jampukka: Few suggestions to improve readability: 1) Remove String response; declaration from outside the loop. 2) Remove allFetched flag, instead use plain while (true) and break; if (isLastPage(response)) 3) try-catch the whole loop, not each iteration - on exception we break the loop anyway

service-statistics-unsd/src/main/java/fi/nls/oskari/control/statistics/plugins/unsd/UnsdIndicatorValuesFetcher.java
```diff
@@ -0,0 +1,81 @@
...
+        try {
+            JSONObject response = JSONHelper.createJSONObject(indicatorDataResponse);
+            int totalPages = response.getInt("totalPages");
+            if (totalPages < 1) {
+                return true;
+            }
+            int pageNumber = response.getInt("pageNumber");
+            return pageNumber == totalPages;
+        } catch (JSONException e) {
+            throw new ServiceRuntimeException("Error parsing UNSD indicator data page info: " + e.getMessage(), e);
```
jampukka: This method throwing a ServiceRuntimeException seems off. `parseIndicatorData` throws JSONException which is catched by the `get` function. This should either throw an `APIException` or just declare that it might throw `JSONException` and let the `get` function handle it.

service-statistics-unsd/src/main/java/fi/nls/oskari/control/statistics/plugins/unsd/parser/RegionMapper.java
```diff
@@ -0,0 +1,51 @@
+package fi.nls.oskari.control.statistics.plugins.unsd.parser;
```
jampukka: IMHO this class should be in the same package as `CountryRegion`

jampukka: I wouldn't even oppose removing `parser` and `requests` packages altogether especially since they only contain three classes in total.

service-statistics-unsd/src/main/java/fi/nls/oskari/control/statistics/plugins/unsd/UnsdConfig.java
```diff
@@ -0,0 +1,58 @@
...
+        goal = json.optString("goal", goal);
+        timeVariableId = json.optString("timeVariable", timeVariableId);
+    }
+
+    public long getId() {
+        return datasourceId;
+    }
+    public String getUrl() {
+        if(url == null) {
+            // default
```
jampukka: Is there a reason for `url` field having a different logic for default value compared to `goal` and `timeVariableId` fields that I can not spot?

service-statistics-unsd/src/main/java/fi/nls/oskari/control/statistics/plugins/unsd/UnsdIndicatorValuesFetcher.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+public class UnsdIndicatorValuesFetcher {
+    private UnsdConfig config;
+
+    public void init(UnsdConfig config) {
```
jampukka: `UnsdIndicatorValuesFetcher` requires this init method to be called before it can be used. It really should be the constructor method instead.

service-statistics-unsd/src/main/java/fi/nls/oskari/control/statistics/plugins/unsd/parser/RegionMapper.java
```diff
@@ -0,0 +1,51 @@
...
+
+
+    private static Optional<CountryRegion> parseRow(String row) {
+        // FI;FIN;246;Finland
+        String[] data = row.split(";");
+        if (data.length != 4) {
+            return Optional.empty();
+        }
+
+        CountryRegion c = new CountryRegion(data[0].trim(), data[1].trim(), data[2].trim(), data[3].trim());
```
jampukka: Consider moving the String#trim() calls to the constructor of `CountryRegion`.

service-statistics-unsd/src/main/java/fi/nls/oskari/control/statistics/plugins/unsd/requests/UnsdRequest.java
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Base class for UNSD statistics queries.
+ */
+public class UnsdRequest {
+
+    private static final String PLACEHOLDER_GOAL = "{goalCode}";
+    private static final String PARAM_INDICATOR = "indicator";
+    private static final String PARAM_PAGE = "page";
+    private static final String PATH_GOAL = "Goal/" + PLACEHOLDER_GOAL;
+    private static final String PATH_TARGETS = PATH_GOAL + "/Target/List";
```
jampukka: I'd remove `PATH_GOAL` and replace `PATH_TARGETS` and `PATH_DIMENSIONS` with
`private static final String PATH_TARGETS = "Goal/{goalCode}/Target/List";`
`private static final String PATH_DIMENSIONS = "Goal/{goalCode}/Dimensions";`
to better distinct PARAM_ and PATH_ fields and to make it more clear that the PATH_s must not begin with a `/`

service-statistics-unsd/src/main/java/fi/nls/oskari/control/statistics/plugins/unsd/UnsdConfig.java
ZakarFin: Not really no -> will change this to work the same way

service-base/src/main/java/fi/nls/oskari/util/IOHelper.java
```diff
@@ -102,6 +104,20 @@ public static String readString(HttpURLConnection conn, final String charset) th
         return readString(conn.getInputStream(), charset);
     }
 
+
+    public static List<String> readLines(InputStream in) throws IOException {
+        return readLines(in, Charset.forName(CHARSET_UTF8));
```
jampukka: See `StandardCharsets.UTF_8`

PR #384, opened 2019-05-20
Parse OWSExceptions

jampukka:
Add support for parsing OWSExceptions in service-wfs-client.
`OskariWFS110Client#getFeatures` now checks if the error was due to bad outputFormat parameter. If it wasn't then it doesn't fallback to GML.

### DIFF
--- a/service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFS110Client.java
+++ b/service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFS110Client.java
@@ -45,7 +45,8 @@ public class OskariWFS110Client {
      * @return SimpleFeatureCollection containing the parsed Features, or null if all fails
      */
     public static SimpleFeatureCollection getFeatures(String endPoint, String user, String pass,
-            String typeName, ReferencedEnvelope bbox, CoordinateReferenceSystem crs, int maxFeatures, Filter filter) {
+            String typeName, ReferencedEnvelope bbox, CoordinateReferenceSystem crs,
+            int maxFeatures, Filter filter) {
         // First try GeoJSON
         Map<String, String> query = getQueryParams(typeName, bbox, crs, maxFeatures, filter);
         query.put("OUTPUTFORMAT", "application/json");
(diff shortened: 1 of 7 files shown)


service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFS110Client.java
```diff
@@ -46,58 +47,57 @@ private OskariWFS110Client() {}
...
-            return getFeaturesGeoJSON(endPoint, user, pass, typeName, bbox, crs, maxFeatures, filter);
+            return parseGeoJSON(new ByteArrayInputStream(response), crs);
         } catch (Exception e) {
-            LOG.warn(e, "Failed to read WFS response as GeoJSON");
+            if (!isOutputFormatInvalid(new ByteArrayInputStream(response))) {
+                // If we can not determine that the exception was due to bad
+                // outputFormat parameter then don't bother trying GML
+                throw new ServiceRuntimeException("Unable to get features");
+            }
+            LOG.debug(e, "Failed to parse GeoJSON response");
```
ZakarFin: Move logging to be done before the exception throw to get the cause.

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFS110Client.java
```diff
@@ -46,58 +47,57 @@ private OskariWFS110Client() {}
...
         // Fallback to GML
+        query.remove("OUTPUTFORMAT");
+        response = getResponse(endPoint, user, pass, query);
+
         try {
-            return getFeaturesGML(endPoint, user, pass, typeName, bbox, crs, maxFeatures, filter);
+            return OSKARI_GML.decodeFeatureCollection(new ByteArrayInputStream(response), user, pass);
         } catch (Exception e) {
-            LOG.warn(e, "Failed to read WFS response as GML");
+            throw new ServiceRuntimeException("Unable to get features");
```
ZakarFin: Attach the cause here?

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFS110Client.java
jampukka: Actually the logging should be done only inside the if block (definitely not after as it is now). If we can determine that the error was because the response was a valid XML <ows:ExceptionReport> stating that the error is because of the `&outputFormat=application/json` then we don't need to log that exception.

jampukka: Decided to just not log anything here, but instead attach the cause to the ServiceRuntimeException if one is thrown

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFS110Client.java
```diff
@@ -44,60 +45,62 @@ private OskariWFS110Client() {}
...
+        try (InputStream in = new BufferedInputStream(conn.getInputStream())) {
+            return parseGeoJSON(in, crs);
         }
+    }
 
-        Map<String, Object> geojson = readMap(conn);
+    private static SimpleFeatureCollection parseGeoJSON(InputStream in,
+            CoordinateReferenceSystem crs) throws IOException {
+        Map<String, Object> geojson = OM.readValue(in, TYPE_REF);
+        in.close();
```
ZakarFin: Move close() to a finally-block?

PR #389, opened 2019-05-22
Service worker for request prioritizing

jheijari:
Leave concurrent connections available for important requests by limiting concurrent lower priority requests.


servlet-map/src/main/resources/service-workers/xhr-prioritizer.js
```diff
@@ -0,0 +1,73 @@
+const DEBUG = false;
+const MAX_CONCURRENT_REQUESTS = 6;
+const LIMIT_CONCURRENT_REQUESTS = 2;
+const PRIO_TRAFFIC_LIMIT = 3;
```
ZakarFin: Remember this doesn’t go through babel so it’s best to use es5

jheijari: Downgraded to ES5. Left fetch intact because it's more widely supported than service workers.

ZakarFin:
The service worker is in the server package so we don't have to instruct users to add a special headers to nginx etc to have it control all the traffic. With the header we could have moved it to the frontend repo:
```
// Set the scope to an upper path of the script location
// Response included "Service-Worker-Allowed : /"
navigator.serviceWorker.register("/js/sw.js", { scope: "/" }).then(() => {
  console.log("Install succeeded as the max allowed scope was overriden to '/'.");
});
```

Note that this only works on sites with HTTPS-support.

ZakarFin:
The point of this is working around the same-domain concurrent request limit that browsers have. Requests that are proxied to external services (and can take a lot of time to complete) can fill this limit before everything the application needs to start has been loaded. Requests that might take a long time to complete are queued manually in the service worker to let requests that are essential to the applications startup to go through.

PR #397, opened 2019-06-05
UN data sources query improvement at thematic maps

mikkomustikkamaa:


ZakarFin:
The default page size is 25 and in demo.oskari.org we are showing the whole world. This reduces the amount of requests that have to be made to get the data.

PR #398, opened 2019-06-05
Add bbox filter support

kuosman:
This PR adds bbox filter support for OpenStreetMapSearchChannel.

BBOX need defined WGS84 format in oskari-ext.properties -file.

For example (Finland's metropolitan area): 
```
# OpenStreetMap bbox filter (WGS84)
search.channel.OPENSTREETMAP_CHANNEL.search.bbox = 24.557944811,59.831111777,25.286173268,60.413952380
```

### DIFF
--- a/service-search-opendata/src/main/java/fi/nls/oskari/search/OpenStreetMapSearchChannel.java
+++ b/service-search-opendata/src/main/java/fi/nls/oskari/search/OpenStreetMapSearchChannel.java
@@ -60,7 +60,7 @@ public class OpenStreetMapSearchChannel extends SearchChannel {
         if (maxResults > 0) {
             params.put("limit", Integer.toString(maxResults));
         }
-        params.put("q", URLEncoder.encode(searchCriteria.getSearchString(),"UTF-8"));
+        params.put("q", searchCriteria.getSearchString());
 
         return IOHelper.constructUrl(serviceURL, params);
     }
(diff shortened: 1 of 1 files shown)


ZakarFin:
This is great but can you refactor the implementation to build the URL with IOHelper? This will make the configuration more fault tolerant (spaces in the property value will break the url with this implementation).

kuosman:
Ok, i change implementation to use IOHelper.constructUrl

service-search-opendata/src/main/java/fi/nls/oskari/search/OpenStreetMapSearchChannel.java
```diff
@@ -40,6 +43,28 @@ public void init() {
...
+            params.put("viewbox", filterBBOX);
+        } else {
+            log.debug("Search BBOX not configured. Add property with key", PROPERTY_BBOX);
+        }
+        params.put("accept-language", searchCriteria.getLocale());
+        int maxResults = getMaxResults(searchCriteria.getMaxResults());
+        if (maxResults > 0) {
+            params.put("limit", Integer.toString(maxResults));
+        }
+        params.put("q", URLEncoder.encode(searchCriteria.getSearchString(),"UTF-8"));
```
ZakarFin: constructUrl does the encoding so no need to do it here.

PR #399, opened 2019-06-06
Made possible to specify alternative TileMatrixSet to be used in print.

Qmppu842:
If one desires, now one can specify alternative TileMatrixSet to be used in print.
This can be done by adding "useThisAlternativeTileMatrixSet" to the oskari_maplayer table in options column.

Example:
UPDATE public.oskari_maplayer 
SET attributes='{"preferredTileMatrix": {"EPSG:3879":"ETRS-GK25"}}' WHERE id =20;
After that query printing layer, which id is 20, will use ETRS-GK25 instead of first available one.

### DIFF
--- a/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
+++ b/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
@@ -201,8 +201,9 @@ public class CommandLoadImageWMTS extends CommandLoadImageBase {
         List<TileMatrixSet> possibleTileMatrixSets = new ArrayList<>();
 
         for (TileMatrixSet tms : capabilities.getTileMatrixSets()) {
-            String key = tms.getTileMatrixMap().keySet().iterator().next();
-            possibleTileMatrixSets.add(tms);
+            if (srs.equals(ProjectionHelper.shortSyntaxEpsg(tms.getCrs()))) {
+                possibleTileMatrixSets.add(tms);
+            }
         }
 
         if (possibleTileMatrixSets.isEmpty()) {
@@ -215,7 +216,7 @@ public class CommandLoadImageWMTS extends CommandLoadImageBase {
         return determineTileMatrixSetToUse(possibleTileMatrixSets);
     }
 
-    private TileMatrixSet determineTileMatrixSetToUse(List<TileMatrixSet> possibleTileMatrixSets) {
+    private TileMatrixSet determineTileMatrixSetToUse(List<TileMatrixSet> possibleTileMatrixSets) throws IllegalArgumentException {
         HashMap<String, String> alternativeTileMatrixSets = new HashMap<>();
         JSONObject useThisTileMatrixSetInstead = layerService.find(layer.getId()).getOptions().optJSONObject("useThisAlternativeTileMatrixSet");
 
--- a/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
+++ b/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
@@ -23,9 +23,6 @@ import fi.nls.oskari.wmts.domain.TileMatrix;
 import fi.nls.oskari.wmts.domain.TileMatrixSet;
 import fi.nls.oskari.wmts.domain.WMTSCapabilities;
 import fi.nls.oskari.wmts.domain.WMTSCapabilitiesLayer;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
 import org.json.JSONObject;
 
 /**
@@ -217,34 +214,22 @@ public class CommandLoadImageWMTS extends CommandLoadImageBase {
     }
 
     private TileMatrixSet determineTileMatrixSetToUse(List<TileMatrixSet> possibleTileMatrixSets) throws IllegalArgumentException {
-        HashMap<String, String> alternativeTileMatrixSets = new HashMap<>();
         JSONObject useThisTileMatrixSetInstead = layerService.find(layer.getId()).getOptions().optJSONObject("useThisAlternativeTileMatrixSet");
-
         if (useThisTileMatrixSetInstead == null) {
             return possibleTileMatrixSets.get(0);
-        } else if (useThisTileMatrixSetInstead.length() > 0) {
-            Iterator iterator = useThisTileMatrixSetInstead.keys();
-            while (iterator.hasNext()) {
-                String nextKey = (String) iterator.next();
-                alternativeTileMatrixSets.put(nextKey, useThisTileMatrixSetInstead.optString(nextKey));
-            }
-
-        }
-
+        } 
+        
         TileMatrixSet setToReturn = null;
         for (TileMatrixSet tms : possibleTileMatrixSets) {
-            String tileMatrixSetName = tms.getTileMatrixMap().keySet().iterator().next();
-            for (Map.Entry<String, String> entry : alternativeTileMatrixSets.entrySet()) {
-                String key = entry.getKey();
-                String value = entry.getValue();
-                if (tileMatrixSetName.contains(key) || tileMatrixSetName.contains(value)) {
+            String tileMatrixSetName = tms.getTileMatrixMap().keySet().iterator().next();            
+            for (int i = 0; i < useThisTileMatrixSetInstead.length(); i++) {
+                String key = (String) useThisTileMatrixSetInstead.keys().next();
+                if (tileMatrixSetName.contains(useThisTileMatrixSetInstead.optString(key, "Something not in tmsName."))) {
                     setToReturn = tms;
-                    break;
+                    break;                    
                 }
-
             }
         }
-
         if (setToReturn != null) {
             return setToReturn;
         }
--- a/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
+++ b/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
@@ -215,6 +215,11 @@ public class CommandLoadImageWMTS extends CommandLoadImageBase {
 
     private TileMatrixSet determineTileMatrixSetToUse(List<TileMatrixSet> possibleTileMatrixSets) throws IllegalArgumentException {
         JSONObject useThisTileMatrixSetInstead = layerService.find(layer.getId()).getAttributes().optJSONObject("preferredTileMatrix");
+        
+        if (useThisTileMatrixSetInstead == null) {
+            return possibleTileMatrixSets.get(0);
+        }
+        
         String id = useThisTileMatrixSetInstead.optString(srs);
         if (id == null) {
             return possibleTileMatrixSets.get(0);
--- a/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
+++ b/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
@@ -220,7 +220,7 @@ public class CommandLoadImageWMTS extends CommandLoadImageBase {
         JSONObject useThisTileMatrixSetInstead = layerService.find(layer.getId()).getOptions().optJSONObject("useThisAlternativeTileMatrixSet");
 
         if (useThisTileMatrixSetInstead == null) {
-            alternativeTileMatrixSets.put(srs, srs);
+            return possibleTileMatrixSets.get(0);
         } else if (useThisTileMatrixSetInstead.length() > 0) {
             Iterator iterator = useThisTileMatrixSetInstead.keys();
             while (iterator.hasNext()) {
--- a/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
+++ b/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
@@ -212,7 +212,10 @@ public class CommandLoadImageWMTS extends CommandLoadImageBase {
         if (possibleTileMatrixSets.size() == 1) {
             return possibleTileMatrixSets.get(0);
         }
+        return determineTileMatrixSetToUse(possibleTileMatrixSets);
+    }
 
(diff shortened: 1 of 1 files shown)


jampukka:
So the use-case is that there is a WMTS service that has multiple TileMatrixSets that all use the same coordinate reference system and choosing the "first" one of them isn't correct? If that is the case then I have couple of suggestions how we can improve the solution.

Currently this change would always use the specified TileMatrixSet even if that TileMatrixSet does not support the coordinate reference system of the request which doesn't make much sense.

Qmppu842:
Yes, that is the case.
I had feeling I had forgotten some cases.
I will happily take improvement suggestions.

jampukka:
First we should change the logic that finds the possibleTileMatrixSets.
We should check the available `TileMatrixSets` for the ones that support the `srs` and collect them to a `List<TileMatrixSet>` (instead of just returning the first one that matches, like the current code does).
If the list is empty we should throw the IllegalArgumentException like is done now. If the list is of size 1 then we just return the `.get(0)` one.

Otherwise we check the DB for the layer options. If no there is no useThisAlternativeTileMatrixSet for the layer then we can just return the first one (like is done currently).

The value of useThisAlternativeTileMatrixSet should instead be a JSONObject with short versions of the CRS as keys and TileMatrixSet identifiers as values, e.g. `"useThisAlternativeTileMatrixSet":{ "EPSG:3879": "ETRS-GK25", "EPSG:3880": "ETRS-GK26" }`
That way the same layer can support multiple coordinate reference systems with this one solution. We the TileMatrixSet identifier we should use by calling `alternativeTileMatrixSet.get(srs)` and then we simply iterate over possibleTileMatrixSets to find the one with the correct identifier.

If none of the possibleTileMatrixSets has the correct identifier then we should probably throw an Exception since that's clearly a configuration error and the admin of the Oskari Instance should be notified by that.

jampukka:
On top of that, the options field for the layer should be made available to the PrintLayer so that we don't have to make DB calls in the middle of the code here. But you don't have to worry about that it will be there once #364 gets completed and I'll make the necessary adjustments myself then.

Qmppu842:
If I interpreted all things correctly, this latest commit should contain your suggested improvements.

Also thanks for helping.

service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
```diff
@@ -190,10 +198,51 @@ private TileMatrix getTileMatrix(TileMatrixSet tms) throws IllegalArgumentExcept
     }
 
     private TileMatrixSet getTileMatrixSet() throws IllegalArgumentException {
+        List<TileMatrixSet> possibleTileMatrixSets = new ArrayList<>();
+
         for (TileMatrixSet tms : capabilities.getTileMatrixSets()) {
-            if (srs.equals(ProjectionHelper.shortSyntaxEpsg(tms.getCrs()))) {
-                return tms;
+            String key = tms.getTileMatrixMap().keySet().iterator().next();
```
jampukka: This gets the "first" identifier of a TileMatrix within the TileMatrixSet. Usually the value should be "0" or something like that. This most probably isn't what we are looking for.
Instead check that the TileMatrixSet supports the CRS of the request `if (srs.equals(ProjectionHelper.shortSyntaxEpsg(tms.getCrs())))` and only if it does then add it to the possibleTileMatrixSets list.

service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
```diff
@@ -190,10 +198,51 @@ private TileMatrix getTileMatrix(TileMatrixSet tms) throws IllegalArgumentExcept
...
+            possibleTileMatrixSets.add(tms);
+        }
+
+        if (possibleTileMatrixSets.isEmpty()) {
+            throw new NullPointerException("Could not find any TileMatrixSets");
```
jampukka: Instead of NullPointerException use the same exception as is thrown now if none of the TileMatrixSets support the CRS of the request, e.g. `throw new IllegalArgumentException("Could not find TileMatrixSet for the requested crs");`

service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
```diff
@@ -190,10 +198,51 @@ private TileMatrix getTileMatrix(TileMatrixSet tms) throws IllegalArgumentExcept
...
+        }
+
+        if (possibleTileMatrixSets.size() == 1) {
+            return possibleTileMatrixSets.get(0);
+        }
+
+        HashMap<String, String> alternativeTileMatrixSets = new HashMap<>();
```
jampukka: Consider moving the rest of the logic to a separate function with a describing name, like `private TileMatrixSet 
 determineTileMatrixSetToUse(List<TileMatrixSet> possibleTileMatrixSets)`.

service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
```diff
@@ -190,10 +198,51 @@ private TileMatrix getTileMatrix(TileMatrixSet tms) throws IllegalArgumentExcept
...
+        JSONObject useThisTileMatrixSetInstead = layerService.find(layer.getId()).getOptions().optJSONObject("useThisAlternativeTileMatrixSet");
+
+        if (useThisTileMatrixSetInstead == null) {
+            alternativeTileMatrixSets.put(srs, srs);
```
jampukka: Just `return possibleTileMatrixSets.get(0);` here instead - if there is no special configuration for the layer then we default to the first value the same way it is done before this PR.

service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
```diff
@@ -190,10 +198,51 @@ private TileMatrix getTileMatrix(TileMatrixSet tms) throws IllegalArgumentExcept
...
+        } else if (useThisTileMatrixSetInstead.length() > 0) {
+            Iterator iterator = useThisTileMatrixSetInstead.keys();
```
jampukka: This whole else block is unnecessary. Here you're constructing a `HashMap<String, String>` out of the JSONObject which already acts as a `Map<String, String>` if you only call `JSONObject#optString(key)`.

service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
```diff
@@ -190,10 +198,51 @@ private TileMatrix getTileMatrix(TileMatrixSet tms) throws IllegalArgumentExcept
...
+            while (iterator.hasNext()) {
+                String nextKey = (String) iterator.next();
+                alternativeTileMatrixSets.put(nextKey, useThisTileMatrixSetInstead.optString(nextKey));
             }
+
+        }
+
+        TileMatrixSet setToReturn = null;
```
jampukka: I can't really understand what the rest of the function is trying to do. Let's think about what the current state here is:
- We have a `List<TileMatrixSet> possibleTileMatrixSets` all supporting the correct `srs`
- We have a `JSONObject useThisTileMatrixSetInstead` that maps the `srs`to the correct `TileMatrixSet#getIdentifier()` - To get the identifier we should use to use we call `useThisTileMatrixSetInstead.optString(srs);`
- The identifier we got from the `JSONObject` might be null, which we should handle accordingly, either by defaulting to `possibleTileMatrixSets.get(0)` or by throwing an exception
- Now we search the possible `possibleTileMatrixSets` for the one with the exact same `#getIdentifier()` value, for example by for-each looping the `possibleTileMatrixSets` and returning the list entry if we determine the identifiers match.
- If we can't find a `TileMatrixSet` from the `possibleTileMatrixSets` with the correct identifier we should throw an exception since this is clearly a configuration error, for example the layer tells us to use `ETRS-GK25` when srs is `EPSG:3879` but the WMTS service doesn't have a TileMatrixSet which both has the identifier `ETRS-GK25` and supports `EPSG:3879`.

service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
```diff
@@ -190,11 +195,44 @@ private TileMatrix getTileMatrix(TileMatrixSet tms) throws IllegalArgumentExcept
...
+    }
+
+    private TileMatrixSet determineTileMatrixSetToUse(List<TileMatrixSet> possibleTileMatrixSets) throws IllegalArgumentException {
+        JSONObject useThisTileMatrixSetInstead = layerService.find(layer.getId()).getOptions().optJSONObject("useThisAlternativeTileMatrixSet");
+        if (useThisTileMatrixSetInstead == null) {
+            return possibleTileMatrixSets.get(0);
+        } 
+        
+        TileMatrixSet setToReturn = null;
+        for (TileMatrixSet tms : possibleTileMatrixSets) {
```
jampukka: This loop is still doing weird things.
This is what I had in mind:
```
String id = useThisTileMatrixSetInstead.optString(srs);
if (id == null) {
    return possibleTileMatrixSets.get(0);
}
for (TileMatrixSet tms : possibleTileMatrixSets) {
    if (tms.getId().equals(id)) {
        return tms;
    }
}
throw new IllegalArgumentException("Could not find TileMatrixSet with id " + id + " for layer " + layer.getId());
```
Or if you prefer Streams then instead of the foor loop:
```
return possibleTileMatrixSets.stream()
        .filter(tms -> tms.getId().equals(id))
        .findAny()
        .orElseThrow(() -> new IllegalArgumentException("Could not find TileMatrixSet with id " + id + " for layer " + layer.getId()));

Comment (Qmppu842) [reply service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java]: I thing I like loops more.
They seem more readable.

Oh... 
Well then I had misunderstanding.
Since you said: 

> ... TileMatrixSet identifiers as values, e.g. "useThisAlternativeTileMatrixSet":{ "EPSG:3879": "ETRS-GK25", "EPSG:3880": "ETRS-GK26" }

I thought that for some reason all of them should be looked to know if any of them is available.

Comment (jampukka) [reply service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java]: Yea the idea is that if the PrintRequest is for example in EPSG:3067 then we'd use TileMatrixSet with id "ETRS-TM35FIN" or if the request is in EPSG:3879 we'd use TileMatrixSet with id "ETRS-GK25" etc. That way the same OskariLayer can be used regardless of the current projection.

**Comment by ZakarFin** [comment]
Could you print this out https://github.com/oskariorg/oskari-docs/blob/master/documents/CLA.txt and fill it, scan/take a picture of it and send it to oskari@nls.fi to get the legal stuff out of the way. This basically says that you can't request us to remove the code you have submitted. And you will only need to do this once. Any subsequent PR can be merged without doing this again.

**Comment by ZakarFin** [comment]
I have a few issues with having "useThisAlternativeTileMatrixSet" in options:
1) Options is meant to be flags for the frontend while this is targeted at only print? Attributes is something that should be used for server-side toggles.
2) We already have the matrixSetIds in the capabilities column in this kind of format:

	`{"srs":["EPSG:3067"],"tileMatrixIds":[{"EPSG:3067":"ETRS-TM35FIN"}]}`

I would much rather see the capabilities field reformatted to have this kind of info instead of adding another config on options or attributes:

	{
    "srs": [
        "EPSG:3067"
    ],
    "tileMatrixIds": {
        "EPSG:3067": [
            {
                "id": "ETRS-TM35FIN",
                "preferred": true
            }
        ]
    }

Of course this results in many more changes than the proposed solution but I think it would be more manageable.

**Comment by ZakarFin** [comment]
Then again I'm not too eager to move user-editable parts to capabilities as the column value should only hold metadata parsed from GetCapabilities response. But currently the data model doesn't really work with multiple matrices for single projection so it would need to be fixed as well. Maybe just migrate/refactor capabilities to:
`{"srs":["EPSG:3067"],"tileMatrixIds":{"EPSG:3067":["ETRS-TM35FIN"]}`

and have an optional preferredTileMatrix in attributes like:
```
{
    "preferredTileMatrix ": { 
        "EPSG:3879": "ETRS-GK25",
        "EPSG:3880": "ETRS-GK26"
    }
}
```

**Comment by Qmppu842** [comment]
> Could you print this out https://github.com/oskariorg/oskari-docs/blob/master/documents/CLA.txt and fill it, scan/take a picture of it and send it to [oskari@nls.fi](mailto:oskari@nls.fi) to get the legal stuff out of the way. This basically says that you can't request us to remove the code you have submitted. And you will only need to do this once. Any subsequent PR can be merged without doing this again.

Okei, 
I will do this tomorrow.

To your second comment:
Sorry, didn't know that.
I will make it use srs.
Or at least that solution from last comment if for some reason first seems to not start working.

**Comment by ZakarFin** [comment]
Sure no problem. The differences between options, params, capabilities and attributes fields are not exactly documented :)

File: service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java [hunk 8]
```diff
@@ -190,12 +195,36 @@ private TileMatrix getTileMatrix(TileMatrixSet tms) throws IllegalArgumentExcept
...
+
+        if (possibleTileMatrixSets.size() == 1) {
+            return possibleTileMatrixSets.get(0);
+        }
+        return determineTileMatrixSetToUse(possibleTileMatrixSets);
+    }
+
+    private TileMatrixSet determineTileMatrixSetToUse(List<TileMatrixSet> possibleTileMatrixSets) throws IllegalArgumentException {
+        JSONObject useThisTileMatrixSetInstead = layerService.find(layer.getId()).getAttributes().optJSONObject("preferredTileMatrix");
+        String id = useThisTileMatrixSetInstead.optString(srs);
```
ZakarFin: useThisTileMatrixSetInstead might be null so you need to check for that.

ZakarFin:
Got the CLA so we are ready for merging. Still need to add a null-check and you could update the original description/sql to match the current implementation. But otherwise good for merging.

ZakarFin:
Thanks!

PR #403, opened 2019-06-14
Unit-testing improvements

ZakarFin:
- Fixed modules that were using shared-test-resources module with default scope to use it with test-scope (so we don't package testing related code to build-artifacts).
- Removed unused JettyLauncher from shared-test-resources
- Refactored code from #271 that enables memory-based H2 database to be used for unit-testing  as TestHelper.createMemDBforUnitTest(List<String> sqlStatementsForInit)


jampukka:
Looks good

PR #404, opened 2019-06-15
Resolve indicator specific parameters

mikkomustikkamaa:
PR contains changes to resolve indicator specific dimensions with UN data sources at thematic maps. With these changes user should see only dimensions relevant to selected indicator(s) in thematic map UI flyout.


service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
```diff
@@ -65,4 +75,63 @@ protected static boolean isLastPage(JSONObject response) throws JSONException {
...
+    }
+
+    public static void parseTimePeriod(Map<Integer, Integer> countOfAreaCodesForYear, JSONObject response) {
+        try {
+            JSONArray data = response.getJSONArray(DATA_KEY);
+            for (int i = 0; i < data.length(); i++) {
+                JSONObject o = (JSONObject) data.get(i);
+                Integer year = getTimeperiodYear(o);
+
+                if (countOfAreaCodesForYear.containsKey(year)) {
```
jampukka: This if-else is very readable and I'm not asking you to change it, but for fun and curiosity I'll note that `countOfAreaCodesForYear.merge(year, 1, Integer::sum);` would've done the same thing.

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
```diff
@@ -65,4 +75,63 @@ protected static boolean isLastPage(JSONObject response) throws JSONException {
...
+    }
+    
+    public static List<IdNamePair> getSortedListOfYearsThatBelongToSeveralGeoAreas(
+            Map<Integer, Integer> countOfAreaCodesForYear) {
+        return (List<IdNamePair>) (countOfAreaCodesForYear.entrySet().stream().filter(e -> e.getValue() > 1)
+                .map(e -> new IdNamePair(String.valueOf(e.getKey()), null))).sorted().collect(Collectors.toList());
+    }
+
+    private static Integer getTimeperiodYear(JSONObject o) throws JSONException {
+        return ((Double) o.get(TIME_PERIOD_START_KEY)).intValue();
```
jampukka: This probably works, but there's also `JSONObject#getInt(String key)`

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
```diff
@@ -65,4 +75,63 @@ protected static boolean isLastPage(JSONObject response) throws JSONException {
...
+            }
+        } catch (JSONException e) {
+            LOG.error("Error parsing time period selectors for indicator: " + e.getMessage(), e);
+        }
+    }
+    
+    public static List<IdNamePair> getSortedListOfYearsThatBelongToSeveralGeoAreas(
+            Map<Integer, Integer> countOfAreaCodesForYear) {
+        return (List<IdNamePair>) (countOfAreaCodesForYear.entrySet().stream().filter(e -> e.getValue() > 1)
+                .map(e -> new IdNamePair(String.valueOf(e.getKey()), null))).sorted().collect(Collectors.toList());
```
jampukka: I'd split the map() here into two separate phases. First mapping the Entry to an Integer, then sorting the Stream<Integer> and only then mapping the years to IdNamePairs, e.g.
```
.stream()
.filter(e -> e.getValue() > 1)
.map(e -> e.getKey())
.sorted()
.map(year -> new IdNamePair(String.valueOf(year), null))
.collect(Collectors.toList())
```

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
mikkomustikkamaa: Changed code as proposed.

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
mikkomustikkamaa: Changed code as proposed. Also removed that private method getTimeperiodYear as unnecessary complexity. Logic was originally implemented as private method because of that not so nice type casting and i felt that there was no point for that private method since we get int directly from json object.

mikkomustikkamaa: Changed code as proposed. Also removed that private method getTimeperiodYear as unnecessary complexity. Logic was originally implemented as private method because of that not so nice type casting and i felt that there was no point for that private method since we get int directly from json object.

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
mikkomustikkamaa: Changed code as proposed

PR #407, opened 2019-06-24
Fixes to thematic map UN data source handling

mikkomustikkamaa:
PR contains following improvements to thematic map UN data source handling:

- Add similar geo area code restriction to indicator dimension resolving that is used in actual data query. This is done to get only indicator dimensions of supported geo areas and improve performance of indicator dimension resolving.
- Add valueType check and logging to data parsing since String data values are received and these are currently not supported.
- Added data object value type check to timeperiod dimension resolving. For example with indicator 10.5.1 geoarea 246 year 2018 data is returned with unsupported data type String. Data with this type cannot be currently shown in user interface so its better to exclude mentioned geoarea from count of years 2018 geoareas with mentioned indicator.   
- Also fixed timeperiod dimension resolving logic related to geo area counting  for each year.

### DIFF
--- a/service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
+++ b/service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
@@ -50,7 +50,12 @@ public class UnsdDataParser {
         JSONArray dataArray = response.getJSONArray("data");
         for (int i = 0; i < dataArray.length(); i++) {
             JSONObject data = dataArray.getJSONObject(i);
-            results.put(data.getString("geoAreaCode"), new IndicatorValueFloat(data.getDouble("value")));
+            String valueType = data.getString("valueType");
+            if("Float".equals(valueType)){
+                results.put(data.getString("geoAreaCode"), new IndicatorValueFloat(data.getDouble("value")));
+            } else {
+                LOG.error(String.format("Not supported valueType %s received.",valueType));
+            }
         }
         return results;
     }
--- a/service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
+++ b/service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
@@ -28,6 +28,7 @@ public class UnsdDataParser {
     private static final String DATA_KEY = "data";
     private static final String TIME_PERIOD_START_KEY = "timePeriodStart";
     private static final String GEO_AREA_CODE_KEY = "geoAreaCode";
+    private static final String SUPPORTED_VALUE_TYPE = "Float";
     private static final Logger LOG = LogFactory.getLogger(UnsdDataParser.class);
 
     private UnsdConfig config;
@@ -50,11 +51,10 @@ public class UnsdDataParser {
         JSONArray dataArray = response.getJSONArray("data");
         for (int i = 0; i < dataArray.length(); i++) {
             JSONObject data = dataArray.getJSONObject(i);
-            String valueType = data.getString("valueType");
-            if("Float".equals(valueType)){
+            if(dataObjectValueTypeIsSupported(data)){
                 results.put(data.getString("geoAreaCode"), new IndicatorValueFloat(data.getDouble("value")));
             } else {
-                LOG.error(String.format("Not supported valueType %s received.",valueType));
+                LOG.error(String.format("Not supported valueType %s received.",data.getString("valueType")));
             }
         }
         return results;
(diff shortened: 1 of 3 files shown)


service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
```diff
@@ -105,22 +117,33 @@ public StatisticalIndicatorDataDimension getTimeperiodDimensionFromIndicatorData
...
+                    });
+                }
             }
         } catch (JSONException e) {
             LOG.error("Error parsing time period selectors for indicator: " + e.getMessage(), e);
         }
     }
     
+    private static boolean dataObjectValueTypeIsSupported(JSONObject o) throws JSONException {
+        return SUPPORTED_VALUE_TYPE.equals(o.getString("valueType"));
```
jampukka: Does the JSONObject here always have a value for `"valueType"` key? If not then o.optString(key); would be better. Also, maybe store the property name `"valueType"` in a field the same way as `DATA_KEY`, `TIME_PERIOD_START_KEY`, `GEO_AREA_CODE_KEY`.

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
```diff
@@ -105,22 +117,33 @@ public StatisticalIndicatorDataDimension getTimeperiodDimensionFromIndicatorData
...
         try {
             JSONArray data = response.getJSONArray(DATA_KEY);
             for (int i = 0; i < data.length(); i++) {
                 JSONObject o = (JSONObject) data.get(i);
                 Integer year = o.getInt(TIME_PERIOD_START_KEY);
-                countOfAreaCodesForYear.merge(year, 1, Integer::sum);
+                Integer geoAreaCode = o.getInt(GEO_AREA_CODE_KEY);
+
+                if(dataObjectValueTypeIsSupported(o)) {
+                    geoAreaCodesForYears.merge(year, new HashSet<Integer>(Arrays.asList(geoAreaCode)), (oldSet,newSet) -> {
```
jampukka: Here `merge` gets too complicated. More readable solution would be: `map.computeIfAbsent(year, k -> new HashSet<>()).add(geoAreaCode);`

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
```diff
@@ -105,22 +117,33 @@ public StatisticalIndicatorDataDimension getTimeperiodDimensionFromIndicatorData
...
+                        return Stream.of(oldSet,newSet).flatMap(x -> x.stream()).collect(Collectors.toSet());
+                    });
```
jampukka: There's no `else` to this `if`, is this correct?

jampukka: If yes, then maybe move `Integer year = ...; and Integer geoAreaCode = ...;` inside the if-block

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
```diff
@@ -44,7 +51,11 @@ protected static boolean isLastPage(JSONObject response) throws JSONException {
         JSONArray dataArray = response.getJSONArray("data");
         for (int i = 0; i < dataArray.length(); i++) {
             JSONObject data = dataArray.getJSONObject(i);
-            results.put(data.getString("geoAreaCode"), new IndicatorValueFloat(data.getDouble("value")));
+            if(dataObjectValueTypeIsSupported(data)){
+                results.put(data.getString("geoAreaCode"), new IndicatorValueFloat(data.getDouble("value")));
```
jampukka: Since we already have GEO_AREA_CODE_KEY let's use it, `data.getString("geoAreaCode")` => `data.getString(GEO_AREA_CODE_KEY)`

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
mikkomustikkamaa: According to swagger document (https://unstats.un.org/SDGAPI/swagger/#!/Indicator/V1SdgIndicatorDataGet) valueType is marked as optional (as almost every field) but according to testing it seems that it is always present. I changed implementation to use optString() method as suggested.

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
mikkomustikkamaa: Updated code as suggested.

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
mikkomustikkamaa: Updated code as suggested.

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdDataParser.java
mikkomustikkamaa: I added else block with logging so implementation does not silently ignore data items with unsupported  value types.

jampukka:
Looks good to me

PR #408, opened 2019-06-24
Additions to Flyway helper methods

mikkomustikkamaa:
- Added helper method to remove registered bundle 
- Added helper method to remove bundle from view

Use case for these is situation when bundle is no longer used and there is need to first remove bundle registration from view(s) and after that remove the bundle.

### DIFF
--- a/content-resources/src/main/java/fi/nls/oskari/db/BundleHelper.java
+++ b/content-resources/src/main/java/fi/nls/oskari/db/BundleHelper.java
@@ -91,14 +91,14 @@ public class BundleHelper {
         }
     }
 
-    public static void removeBundle(final Bundle bundle, Connection conn) throws SQLException {
-        if (isBundleRegistered(bundle.getName(), conn)) {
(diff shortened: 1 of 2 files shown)


content-resources/src/main/java/fi/nls/oskari/db/BundleHelper.java
```diff
@@ -89,4 +90,15 @@ public static Bundle getRegisteredBundle(final String id, Connection conn) throw
             }
         }
     }
+
+    public static void removeBundle(final Bundle bundle, Connection conn) throws SQLException {
```
ZakarFin: You should just have String bundleName instead of Bundle as param so it's easier to see what this method needs/uses (it could getId() from bundle). Also you should rename this to unregisterBundle() as it's an opposite of registerBundle()

mikkomustikkamaa: Updated method name and parameters as proposed

PR #409, opened 2019-06-26
Feature handlers refactoring and support to edit multiple features properties

kuosman:
* Removed StringWriters and refactored to use XMLStreamWriter/Geotools
* Refactored InsertFeatureHandler and SaveFeatureHandler featuredata param to use JSONArray (before JSONObject)
* SaveFeatureHandler now supports also multiple features properties updating (not geometries)


control-base/src/main/java/fi/nls/oskari/control/feature/AbstractFeatureHandler.java
```diff
@@ -112,91 +122,131 @@ protected boolean isAllowedGeomType(String type) {
...
+        feature.setNamespace(lc.getFeatureNamespace());
+        feature.setNamespaceURI(lc.getFeatureNamespaceURI());
+        feature.setGMLGeometryProperty(lc.getGMLGeometryProperty());
+        feature.setId(jsonObject.getString("featureId"));
+
+        if(jsonObject.has("featureFields")) {
+            JSONArray jsonArray = jsonObject.getJSONArray("featureFields");
+            for (int i = 0; i < jsonArray.length(); i++) {
+                JSONObject property = jsonArray.getJSONObject(i);
+                if(property.getString("value").isEmpty()) {
```
jampukka: Is there a need to set property value to null/empty value?

control-base/src/main/java/fi/nls/oskari/control/feature/SaveFeatureHandler.java
```diff
@@ -4,80 +4,70 @@
...
+                flushLayerTilesCache(layer.getId());
 
-            if (responseString.indexOf("Exception") > -1) {
-                ResponseHelper.writeResponse(params, "Exception");
-            } else if (responseString.indexOf("<wfs:totalUpdated>1</wfs:totalUpdated>") > -1) {
-                ResponseHelper.writeResponse(params, "");
+                if (responseString.indexOf("Exception") > -1) {
+                    throw new ActionException("Cannot save feature: " + responseString);
+                } else if (responseString.indexOf("<wfs:totalUpdated>1</wfs:totalUpdated>") > -1) {
+                    ResponseHelper.writeResponse(params, "Feature updated");
```
jampukka: We either throw an exception or write a response for each iteration of the loop.

control-base/src/main/java/fi/nls/oskari/control/feature/AbstractFeatureHandler.java
```diff
@@ -112,91 +122,131 @@ protected boolean isAllowedGeomType(String type) {
...
-        writer.append(srsName);
-        writer.append("\">");
+    protected Geometry getMultipoint(int srid, JSONArray data) throws JSONException {
+        List<Point> points = new ArrayList<>();
         for (int i = 0; i < data.length(); i++) {
-            writer.append("<gml:pointMember><gml:Point><gml:coordinates decimal=\".\" cs=\",\" ts=\" \">"
-                    + data.getJSONObject(i).getString("x") + "," + data.getJSONObject(i).getString("y")
-                    + "</gml:coordinates></gml:Point></gml:pointMember>");
+            Point point = gf.createPoint(new Coordinate(
+                    Double.parseDouble(data.getJSONObject(i).getString("y")),
```
jampukka: Axis order gets swapped here. This was probably only tested with a CRS that has axis order northing/easting (y/x, lat/lon)? GML3Writer writes the JTS Geometry as x/y, but here we read y to x and x to y swapping the order. Probably should just read x to x and y to y here and then swap the order if necessary (depending on `CoordinateReferenceSystem` we decode with CRS.decode) using `CoordinateSequenceFilter` (see https://stackoverflow.com/a/29634389)

jampukka:
`FeatureWFSTRequestBuilder` supports inserting, updating or deleting multiple features in a single WFS-T request but none of the ActionHandlers actually use this, instead they send a separate WFS-T request for each feature one at a time. Is this intentional?

kuosman:
Ok, base case in front-end sends only one feature (multipoint, multipolygon, multilinestring) to handlers, but if editing multiple features then there comes one to n features without geometries. Front-end code sends always JSONArray even then if there is only one feature to inserting, updating or deleting. Maybe these multiple features supports has too much (because not actually used this time). 

I check also others / needs some testing now.

control-base/src/main/java/fi/nls/oskari/control/feature/FeatureWFSTRequestBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+        xsw.writeEndDocument();
+        xsw.close();
+    }
+
+    private static void addGeometries(XMLStreamWriter xsw, Feature feature, boolean isUpdate) throws XMLStreamException {
+        if(!feature.hasGeometry()) {
+            return;
+        }
+
+        if(isUpdate == false) {
```
jampukka: This function is only called from two different places but still it requires a special `isUpdate` flag which totally changes what the function actually does. Also, the insert version (when `isUpdate` is `false`) is only three lines of code. I think it's more readable if we move these three lines here to the `insertFeature` function and wrap them inside a `if (feature.hasGeometry()) {}` block there instead. Then we can remove the `boolean isUpdate` argument and rename the function to something more detailed like `writeGeometryProperty`

control-base/src/main/java/fi/nls/oskari/control/feature/InsertFeatureHandler.java
```diff
@@ -35,58 +39,62 @@
...
-        final String responseString = postPayload(layer, wfstMessage);
-        final String updatedFeatureId = parseFeatureIdFromResponse(responseString);
+                final String wfstMessage = createWFSTMessage(featureJSON);
+                LOG.debug("Inserting feature to service at", layer.getUrl(), "with payload", wfstMessage);
+                final String responseString = postPayload(layer, wfstMessage);
+                updatedFeatureIds.put(parseFeatureIdFromResponse(responseString));
 
-        flushLayerTilesCache(layer.getId());
-        ResponseHelper.writeResponse(params, JSONHelper.createJSONObject("fid", updatedFeatureId));
+                flushLayerTilesCache(layer.getId());
```
jampukka: This could be improved by collecting all distinct layerIds updated in the loop in to a Set and then flushing the layers after this loops is finished. That way we only flush each different layer once.

control-base/src/main/java/fi/nls/oskari/control/feature/InsertFeatureHandler.java
```diff
@@ -35,58 +39,62 @@
...
-        }
+        try {
+            JSONArray paramFeatures = new JSONArray(params.getHttpParam("featureData"));
+            JSONArray updatedFeatureIds = new JSONArray();
+            for (int i = 0; i < paramFeatures.length(); i++) {
+                JSONObject featureJSON = paramFeatures.getJSONObject(i);
+                OskariLayer layer = getLayer(featureJSON.optString("layerId"));
+
+                if (!canEdit(layer, params.getUser())) {
+                    throw new ActionDeniedException("User doesn't have edit permission for layer: " + layer.getId());
```
jampukka: Should we check this before for each of the features/layers first? Say I try inserting five features in one request, the first four are OK but for the fifth I don't have the edit permission for that layer. The response I get is that I don't have a permission, but the first four actually completed successfully.

jampukka: Consider solving both of these issues by first iterating the `paramFeatures` array once collecting the `OskariLayer`s into a `Map<String, OskariLayer>`. Then iterate over the values of the map checking that user has edit permission for all of the layers. Then iterate over `paramFeatures` again (but this time get the `OskariLayer` from the `Map<String, OskariLayer>`) and send the WFS-T requests. Finally iterate over the values of the `Map` once again flushing the layerTilesCache once for each layer.

control-base/src/main/java/fi/nls/oskari/control/feature/InsertFeatureHandler.java
```diff
@@ -35,58 +39,62 @@
...
-                    requestData.append("<" + key + ">" + value + "</" + key + ">");
-                }
-            }
 
-            if (jsonPayload.has("geometries")) {
-                insertGeometries(wfsMetadata.getGMLGeometryProperty(), requestData, jsonPayload.getJSONObject("geometries"), srsName);
-            }
-        } catch (JSONException ex) {
-            throw new ActionParamsException("Couldn't create WFS-T message from params", jsonPayload.toString(), ex);
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
jampukka: Could move the ByteArrayOutputStream and the return clause to inside the try-catch block.

control-base/src/main/java/fi/nls/oskari/control/feature/InsertFeatureHandler.java
jampukka: Same thing for `SaveFeatureHandler`. `DeleteFeatureHandler` doesn't have the same problem as it only allows one feature per request.

control-base/src/main/java/fi/nls/oskari/control/feature/SaveFeatureHandler.java
```diff
@@ -4,80 +4,73 @@
...
-            throw new ActionException("JSON processing error", ex);
-        }
-    }
 
-    protected void insertGeometries(String geometryProperty, StringBuilder requestData, JSONObject geometries, String srsName) throws ActionParamsException, JSONException {
-        String geometryType = geometries.getString("type");
-        if(!isAllowedGeomType(geometryType)) {
-            throw new ActionParamsException("Invalid geometry type: " + geometryProperty);
+            if(exceptions.size() > 0 ) {
+                throw new ActionException("Cannot save feature(s): " + exceptions.toString());
```
jampukka: This is confusing since we might have successfully updated multiple features but one more more failed.

jampukka:
Looks pretty good already, added few comments about the logic.

jampukka:
Looks good, thanks a lot!

PR #410, opened 2019-06-27
UNStats statistics adapter tuning

ZakarFin:
- Longer read timeout (30seconds instead of the standard 6s) as the service might answer slowly
- Sort years in reverse order to get the latest statistics first on the UI


service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdRequest.java
```diff
@@ -151,6 +151,8 @@ private String getData(String path, Map<String, String> params) throws APIExcept
         try {
             final String url = getUrl(path, params);
             con = IOHelper.getConnection(url);
+            // 30 seconds as this might take a while and we are _usually_ doing this on the background thread
+            con.setReadTimeout(30000);
```
jampukka: If there's a lot of data we could add a `Accept-Encoding: gzip` header here and call `IOHelper.readString(HttpURLConnection)` instead.

jampukka:
Added a note about maybe enabling gzip for the UnsdRequest. Other than that looks good and ready for merging.

service-statistics-unsd/src/main/java/org/oskari/statistics/plugins/unsd/UnsdRequest.java
ZakarFin: I'll add this as a configuration option as well. The response isn't huge but for some reason connection to the API can be a bit shaky at times.

PR #411, opened 2019-06-27
Initial support for common JUnit DB

ZakarFin:
Creates a membased DB datasource for unit testing if a class from the shared-test-resources is present in the classpath. Checking things with reflection API so we don't need to add the test-module to dependencies but can detect test env if the class file is available.

Separated from #271 waiting for further development for this to be actually usable.


ZakarFin:
Redoing as draft pr

PR #412, opened 2019-06-27
Initial support for common JUnit DB

ZakarFin:
Creates a membased DB datasource for unit testing if a class from the shared-test-resources is present in the classpath. Checking things with reflection API so we don't need to add the test-module to dependencies but can detect test env if the class file is available.

Separated from #271 waiting for further development for this to be actually usable.


ZakarFin:
More recent testing: https://github.com/oskariorg/oskari-server/compare/develop...ZakarFin:testdb?expand=1
Still struggling with the cyclic dependency ref. Probably needs refactoring of current modules or detaching tests to another module to make it work properly.

ZakarFin:
Looks like running the Flyway-migrations to H2 database would require quite a bit of changes to the current scripts.
Things that don't work:

- The `SET` commands at the start of https://github.com/oskariorg/oskari-server/blob/2.14.0/content-resources/src/main/resources/flyway/oskari/V2_0_1__tables_for_empty_db.sql#L5-L13

- The `ALTER TABLE ONLY` <- only isn't supported, used in many of the sequence value as id column defaults https://github.com/oskariorg/oskari-server/blob/2.14.0/content-resources/src/main/resources/flyway/oskari/V2_0_1__tables_for_empty_db.sql#L1240

- Some constraint names have duplicates with the same name in for example:
    - `nullchk` in https://github.com/oskariorg/oskari-server/blob/2.14.0/content-resources/src/main/resources/flyway/oskari/V2_0_1__tables_for_empty_db.sql#L919
    - `oskari_maplayer_id_fkey` here https://github.com/oskariorg/oskari-server/blob/2.14.0/content-resources/src/main/resources/flyway/oskari/V2_0_1__tables_for_empty_db.sql#L1813-L1823

- `COPY` doesn't work so this would need to be changed: https://github.com/oskariorg/oskari-server/blob/2.14.0/content-resources/src/main/resources/flyway/oskari/V2_0_2__initial_bundles.sql#L21

into something like this:
```sql
INSERT INTO public.portti_bundle  (id, name, config, state)
VALUES (4, 'toolbar', '{}', '{}'),
      (5, 'divmanazer', '{}', '{}');
```

ZakarFin:
After those the https://github.com/oskariorg/oskari-server/blob/2.14.0/content-resources/src/main/resources/flyway/oskari/V2_0_3__internal_data.sql#L2 fails with an error:

```
Caused by: org.flywaydb.core.internal.sqlscript.FlywaySqlScriptException: Script V2_0_3__internal_data.sql failed
---------------------------------------
SQL State  : 23502
Error Code : 23502
Message    : NULL not allowed for column "id"; SQL statement:
```

ZakarFin:
After this I'm not certain it's worth pursuing having the Flyway migrations getting run on h2. Maybe if there's something out there that would just eat the stuff we feed into PostgreSQL. Also it's worth noting that seems that we do have Postgres as the required db as of now :)

PR #414, opened 2019-07-30
Use map view specific publishing templates

jheijari:
Setting `publishTemplateUuid` to map view's metadata enables view specific publishing template.


jampukka:
Looks good

PR #415, opened 2019-08-01
Reset remaining session time

mikkomustikkamaa:
PR contains new ActionRoute implementation intended to call when user session needs to be extended.


jampukka:
Looks good

PR #429, opened 2019-09-02
Print vector features

okauppinen:
Add Oskari style support for vector layer printing.

Uses png icons for point style (created from svg-markers.json), but own external icons are not supported.
Uses pdf renderer to create BufferedImage for raster prints (style support added only for pdf).

GetPrintHandler uses AbstractWFSFeaturesHandler findLayer which throwed exception for other than WFS layers. Now returns layer. Maybe needs some refactoring.

Includes changes: https://github.com/oskariorg/oskari-server/pull/364
Frontend chages for custom style: https://github.com/oskariorg/oskari-frontend/pull/935


service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -351,44 +411,349 @@ private static String getUnits(String srsName) {
...
+        add(stream, g.getExteriorRing().getCoordinateSequence(), true);
+        for (int i = 0; i < g.getNumInteriorRing(); i++) {
+            add(stream, g.getInteriorRingN(i).getCoordinateSequence(), true);
+        }
+        if (fill && stroke) {
+            stream.fillAndStrokeEvenOdd();
+        } else if (fill) {
+            stream.fillEvenOdd();
+        }
+        stream.stroke();
```
jampukka: This should be in an else block

jampukka: We'll fix it if it really is an issue

PR #437, opened 2019-09-10
Show designated text on map

jheijari:
My places input form has a field "Text visible on map" which is not used for the label.
Fixes that issue.

Makes the field visible in feature properties also.


ZakarFin:
Might cause changes in functionality if transport is used. But we are migrating away from it by default so let's roll forward.

PR #438, opened 2019-09-10
Enable publications without domain restrictions

jheijari:
Enables publication usage from any domain when publication has no domain restriction set.


ZakarFin:
Relates to oskariorg/oskari-frontend#946

PR #450, opened 2019-09-18
Ignore property geometries from GeoJSON WFS responses

jampukka:
For whatever reason sometimes secondary geometries take over in the process when GetWFSFeatures gets called. Solve the issue by ignoring the secondary geometries for now.

### DIFF
--- a/oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONSchemaDetector.java
+++ b/oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONSchemaDetector.java
@@ -31,26 +31,19 @@ public class GeoJSONSchemaDetector {
 
         // Map feature.geometry fields to JTS Geometries
         replaceGeometry(json, GeoJSONReader2::toGeometry);
-        if (ignoreGeometryProperties) {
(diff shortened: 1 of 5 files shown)


oskari-geojson/src/main/java/org/oskari/geojson/GeoJSONSchemaDetector.java
```diff
@@ -20,11 +20,37 @@
...
         // Might cause problems later on...
-        GeoJSONReader2.replaceMapsWithGeometries(json);
+
+        // Map feature.geometry fields to JTS Geometries
+        replaceGeometry(json, GeoJSONReader2::toGeometry);
+        if (ignoreGeometryProperties) {
+            // Remove properties of type Map<String, Object> that were mappable to JTS geometries, otherwise leave as is
+            replaceMapProperties(json, it -> {
+                try {
+                    GeoJSONReader2.toGeometry(it);
```
ZakarFin: Would this read better if the if is inside the try catch? 

    Object geom = GeoJSONReader2.toGeometry(it);
    if (ignoreGeometryProperties) {
        return null;
    }
    return geom;

jampukka: Let me try

PR #452, opened 2019-09-26
Use layer ID instead of search channel ID for permission

ZakarFin:
While refactoring the permissions to use layer id instead of type+url+name combo this broke. The hasPermission tried to get a layer permission using search channel id instead of layer id. This resulted in no user having permission to any wfs backed search channel. (Or in worst case permissions were used from a wrong layer if search channel id happened to match some layer id)


kuosman:
Tested and working now fine with Valkeakoski environment

PR #454, opened 2019-09-27
wfs-client force gml

okauppinen:
If layer has forceGML true in attributes then skip geojson


ZakarFin:
Admins can force a layer to always use GML instead of JSON by configuring the attributes field with:
```
{
   "forceGML": true
}

PR #458, opened 2019-10-08
Move Resource caching from PermissionHelper to the service implementation

jampukka:
`PermissionHelper`s cache of `Resource`s gets outdated when changes are made via PermissionService. Move the caching of Resource objects by type and mapping from `PermissionHelper` to the actual service implementation.


ZakarFin:
Nice 👍

ZakarFin:
Fixes the issue of permissions not being updated immediately when admin updates permissions (if the layer was viewed on the map before updating permissions).

PR #459, opened 2019-10-08
Refactoring the background process updating IndicatorList caches

jampukka:
Split logic from DataSourceUpdater into two separate classes, more clearly separate the cases of having an empty cache vs having cache but just want to refresh it.

Check if the update process has been running for too long. In case it has, restart it.

Only send the `"complete": false` to the client if we are running a "full update" e.g. the cache was empty when the update process started.

Remove the boolean isUpdating/complete flag from `DataStatus`. Instead use `updateStarted` field to determine whether it is updating or not.

### DIFF
--- a/service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/plugins/db/StatisticalDatasource.java
+++ b/service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/plugins/db/StatisticalDatasource.java
@@ -20,7 +20,7 @@ public class StatisticalDatasource {
     private String plugin;
     private List<DatasourceLayer> layers = new ArrayList<>();
     private long maxUpdateDuration = TimeUnit.HOURS.toMillis(6);
-    private long updateInterval = 1000 * 60 * 60 * 24;
+    private long updateInterval = TimeUnit.HOURS.toMillis(24);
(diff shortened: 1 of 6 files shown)


service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/plugins/db/StatisticalDatasource.java
```diff
@@ -18,6 +19,7 @@
     private String config;
     private String plugin;
     private List<DatasourceLayer> layers = new ArrayList<>();
+    private long maxUpdateDuration = TimeUnit.HOURS.toMillis(6);
     private long updateInterval = 1000 * 60 * 60 * 24;
```
ZakarFin: Change this to TimeUnit as well so it doesn't feel out of place with maxUpdateDuration

ZakarFin:
Looks good for merging. Just the small TimeUnit consistency change and we are good to go.

service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/plugins/db/StatisticalDatasource.java
jampukka: Done

PR #461, opened 2019-10-11
Layer admin handler

okauppinen:
New action route LayerAdmin for add, edit and delete maplayers. Uses template.json like structure for passing layers. New services for capabilities parsing which parses layers to OskariLayer (not to json). OskariLayers are parsed to json (template.json structure) in LayerAdminHandler. Doesn't use GetWSCapabilitiesHandler or GetGtXXXCapabilities. Uses OskariLayerCapabilitiesHelper and LayerJSONFormatterXXX to creating OskariLayer's capabilities json, which is created when layer is received (not updated when layer is saved). Assumes that frontend passes capabilities json back when requesting insert or update. 

LayerAdmin returns also permissions for roles. For admin returns whole role list and non-admin gets only own roles + guest. Editing layer's permissions are allowed only for admins. For new layer's pemission template is returned which contains all permissions for admin and wiew published for guest. So frontend has to request/get available permission set only to populate edit permissions tab.

From GetCapabilities all succesfully parsed layers all listed to layers. existingLayers contains layer names which are already added. unsupportedLayers contains layer names which capabilities dosen't contain current srs. capabilitiesFailed contains layer names which capabilities parsing failed.

Added CapabilitiesConstants to be sure that same keys and values are used when handling capabilities.

Tries to parse geomName, geom, srs, keywords, GetFeature formats and max features for wfslayer's (1.1.0, 2.0.0). Updated OskariWFSClient's try GeoJSON and get max features.

Includes also changes for old SaveLayer and GetWSCapabilitiesHandler:
https://github.com/oskariorg/oskari-server/pull/455/files
Added allowed formats parsing for GetFeature request.
Added getLayerCapabilities to get json object for Oskari layer. To keep sure that get wsscapabilties, save layer and update capabilties creates similar objects.
Some refactoring GetGTWSCapabilities exception handling to pass exceptions instead of overriding them with more generic error message.


control-admin/src/main/java/fi/nls/oskari/control/admin/AbstractLayerAdminHandler.java
```diff
@@ -0,0 +1,140 @@
...
+
+    }
+
+    protected PermissionService getPermissionsService() {
+        return permissionsService;
+    }
+
+    protected UserService getUserService() throws ServiceException {
+        if (userService == null) {
+            userService = UserService.getInstance();
```
jampukka: Why is this lazily initialized but PermissionService isn't?

control-admin/src/main/java/fi/nls/oskari/control/admin/AbstractLayerAdminHandler.java
```diff
@@ -0,0 +1,140 @@
...
+        return userService;
+    }
+
+    protected boolean userHasEditPermission(User user, OskariLayer layer) {
+        return user.isAdmin() || permissionsService.findResource(ResourceType.maplayer, Integer.toString(layer.getId()))
+                .filter(r -> r.hasPermission(user, PermissionType.EDIT_LAYER)).isPresent();
+    }
+
+    protected boolean userHasAddPermission(User user) {
+        return user.isAdmin() || permissionsService.findResource(ResourceType.functionality, "generic-functionality")
```
jampukka: Is `"generic-functionality"` stored somewhere as a constant? If it is then should re-use that.

control-admin/src/main/java/fi/nls/oskari/control/admin/AbstractLayerAdminHandler.java
```diff
@@ -0,0 +1,140 @@
...
+    protected boolean userHasCapabilitiesPermission(User user) {
+        return user.hasAnyRoleIn(capabilitiesRoles) || userHasAddPermission(user);
+    }
+    protected int getRoleId(String name) throws ServiceException {
+        final Role role = getUserService().getRoleByName(name);
+        return (int) role.getId();
+
+    }
+    protected JSONObject getPermissionsForLayer (User user, OskariLayer layer) throws ActionException, ServiceException {
+        if(!userHasEditPermission(user, layer)) throw new ActionDeniedException("User doesn't have edit permission for layer: " + layer.getId());
```
jampukka: This would be prettier wrapped in a `{ }` block

jampukka: Also I'm not sure if this is a correct place for permission checking. Seems like this function is called pretty deep into another function inside LayerAdminHandler. Also, the name of the function gives no clues about requiring the user to have an edit permission for the layer.

jampukka: In addition, while we will eventually write the permissions to the frontend (or store them somewhere else) as JSON, I really dislike mixing JSONObjects into functions dealing with logic and then passing them around the codebase. In my opinion it makes reading, maintaining and unit testing the code a lot more difficult. This function could easily be named `getPermissionsGroupByRole` returning a `Map<Role, List<String>>` and it would make it crystal clear what the function actually does. Then for JSON serialization we can later choose to use `Role#getName` if that's sufficient.

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariFeatureClient.java
```diff
@@ -18,6 +18,7 @@
 public class OskariFeatureClient {
 
     protected static final String PROPERTY_NATIVE_SRS = "oskari.native.srs";
+    protected static final String PROPERTY_FORCE_GML = "forceGML";
```
jampukka: This same constant appears in OskariWFSClient

control-admin/src/main/java/fi/nls/oskari/control/admin/AbstractLayerAdminHandler.java
okauppinen: Now tries also in init. Reason for getUserService() is to handle ServiceException. Could also use UserService.getInstance() without getUserService() and initialization in AbstractLayerAdminHandler.

control-admin/src/main/java/fi/nls/oskari/control/admin/AbstractLayerAdminHandler.java
okauppinen: Added constant to PermissionService

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariFeatureClient.java
okauppinen: removed

service-map/src/main/java/fi/nls/oskari/map/layer/OskariLayerServiceMybatisImpl.java
```diff
@@ -302,6 +302,18 @@ public OskariLayer find(int id) {
         }
         return Collections.emptyList();
     }
+    public List<String> findAllLayerNames () {
```
ZakarFin: Should this be "for url"?

ZakarFin: When we have something like "buildings" on two different services this doesn't make much sense.

ZakarFin:
This still needs some work but it shouldn't break anything and it's easier to test/integrate with the frontend when it's merged. Also any changes that this needs are easier to check when they are separate PRs instead of parts in this monster :)

PR #462, opened 2019-10-14
new action route to get all roles and permission types

mikkomustikkamaa:


control-admin/src/main/java/fi/nls/oskari/control/admin/GetAllRolesAndPermissionTypesHandler.java
```diff
@@ -0,0 +1,121 @@
...
+    public void init() {
+        try {
+            userService = UserService.getInstance();
+        } catch (Exception ex) {
+            log.error(ex, "Unable to initialize User service!");
+        }
+        try {
+            permissionsService = OskariComponentManager.getComponentOfType(PermissionService.class);
+        } catch (Exception ex) {
+            log.error(ex, "Unable to initialize permission service!");
```
ZakarFin: Since we never try to recover from the error later on it would be best to let exceptions be thrown from the init(). That way the ActionHandler is not available for use by the frontend and we don't need to check if the services are initialized correctly for every request on preProcess().

control-admin/src/main/java/fi/nls/oskari/control/admin/GetAllRolesAndPermissionTypesHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    @Override
+    public void handleGet(ActionParameters params) throws ActionException {
+        try {
+            
+            final JSONObject root = new JSONObject();
+            getRoles(root);
+            getPermissionTypes(root,params.getLocale().getLanguage());
```
ZakarFin: Refactor getRoles() and getPermissionTypes() to return the array and construct the response object here since methods don't really need the root as param.

ZakarFin:
A couple of minor fixes but after that looks good for merging.

control-admin/src/main/java/fi/nls/oskari/control/admin/GetAllRolesAndPermissionTypesHandler.java
mikkomustikkamaa: I now fixed this such way that checked exceptions are converted to unchecked RuntimeExceptions and those are thrown. In this way we don't need to add checked exception(s) to init method signatures of parent classes and cause need to make changes to other action handlers.

control-admin/src/main/java/fi/nls/oskari/control/admin/GetAllRolesAndPermissionTypesHandler.java
mikkomustikkamaa: Implemented approach was attempt to get less code in handleGet method by putting role and permission related queries and loops to their own private methods. I now changed the implementation such way that root json is now constucted in handleGet method. Role and permission related json arrays are still constructed in private methods.

control-admin/src/main/java/fi/nls/oskari/control/admin/GetAllRolesAndPermissionTypesHandler.java
ZakarFin: Remove the log.error() calls as well so we don't get the exception logged twice.

ZakarFin: You can also use ServiceRuntimeException instead of RuntimeException to notify it's a exception from Oskari instead of a plain runtime one.

mikkomustikkamaa: Fixed as proposed

PR #476, opened 2019-11-26
Add camera-controls-3d plugin to 3d views

mikkomustikkamaa:
Flyway migration to add new 3D camera controls bundle to 3D views.

### DIFF
--- a/content-resources/src/main/java/fi/nls/oskari/util/FlywayHelper.java
+++ b/content-resources/src/main/java/fi/nls/oskari/util/FlywayHelper.java
@@ -149,4 +149,19 @@ public class FlywayHelper {
             statement.execute();
         }
     }
+    public static List<Long> get3DApplicationViewIds(Connection connection)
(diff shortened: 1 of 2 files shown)


ZakarFin:
I think this is more of a application specific thing than something that should be migrated for all instances. If this would be a generic migration the application names should come from properties.

mikkomustikkamaa:
I thought at implementation time that it would be good idea to have this migration at one place in oskari-server since 3D code will be part of oskari. I agree that using hard coded expected application names in migration script is poor choice but i did not figure out better way to include bundle only to 3D views and not to 2D views. I will close this PR and make this migration logic to application spesific repositories.

ZakarFin:
The part where the bundle is registered is ok to be in oskari-server. And there is a case to be made for a helper function for adding the bundle to appsetups but I guess the addition of the bundle to appsetups is so short with existing helper functions that we can just link an example on MigrationGuide.md

PR #480, opened 2019-12-02
Fix IOHelper timing issues

ZakarFin:
IOHelper is referenced from code before all the properties have been read. This fix probably makes it a bit slower, but fixes it in a way that oskari-ext.properties actually affect timeouts and _dangerously_ trusting certificates that would otherwise need tuning of the Java certificate truststore.


jampukka:
I like it

PR #489, opened 2019-12-11
Register shadow plugin bundle

Somac:
Flyway migration for shadowing plugin


ZakarFin:
Requires oskariorg/oskari-frontend#1098

PR #490, opened 2019-12-12
Read MIF/MID without OGR

jampukka:
In preparation to remove the dependency to OGR.


ZakarFin:
Don't know if it works but looks good enough for merging to me. Move the module under the geotools-ext folder like the other one and I'll merge after that.

PR #492, opened 2019-12-17
Update GeoTools from 19.2 to 22.1

tsallinen:
Also updated dependencies as required:

* Since gt-ogr is yet again supported, removed version override to older version
* Updated Apache Commons Lang 3 to 3.8.1
* Updated Java Topology Suite to 1.16.1
* Updated Mapbox Vector Tile library to 3.1.0


ZakarFin:
Using this as base and fixed the new code/conflicts for version 2.0 that is now available as branch "v2.0": https://github.com/oskariorg/oskari-server/tree/v2.0

PR #494, opened 2019-12-17
Transport cleanup

ZakarFin:
The transport webapp for WFS integrations is no longer supported in Oskari. Cleaning up the code.

Note! Includes commits from #492


ZakarFin:
Turns out removing transport while upgrading GeoTools was a bad idea. Made a new PR for just removal #504

PR #499, opened 2019-12-20
WFS allowed feature formats capabilities parsing fix

okauppinen:


ZakarFin:
Fixes an issue where we didn't detect that service only supports GML and tried asking for JSON anyways.

PR #508, opened 2020-01-15
Add action route handler

mikkomustikkamaa:
Pull request contains implementation of action route handler to query single data provider with id


control-base/src/main/java/fi/nls/oskari/control/layer/GetDataProviderHandler.java
```diff
@@ -0,0 +1,39 @@
...
+
+	private DataProviderService service = ServiceFactory.getDataProviderService();
+
+	@Override
+	public void handleAction(ActionParameters params) throws ActionException {
+		try {
+			final int id = Integer.valueOf(params.getHttpParam(PARAM_ID));
+			DataProvider d = service.find(id);
+			final JSONObject result = new JSONObject();
+			JSONHelper.putValue(result, "dataprovider", d.getAsJSON());
```
ZakarFin: I dont think the response needs to we wrapped in an additional JSON object. Just return the getAsJSON() as is? Also consider renaming this to DataProvider extending RestActionHandler so we could combine the save/delete to this as well and get rid of https://github.com/oskariorg/oskari-server/blob/master/control-base/src/main/java/fi/nls/oskari/control/layer/SaveOrganizationHandler.java etc

control-base/src/main/java/fi/nls/oskari/control/layer/GetDataProviderHandler.java
```diff
@@ -0,0 +1,39 @@
...
+public class GetDataProviderHandler extends ActionHandler {
+
+	private static final String PARAM_ID = "id";
+
+	private DataProviderService service = ServiceFactory.getDataProviderService();
+
+	@Override
+	public void handleAction(ActionParameters params) throws ActionException {
+		try {
+			final int id = Integer.valueOf(params.getHttpParam(PARAM_ID));
```
ZakarFin: There's probably some params.getRequiredParamInt() that you could use here

control-base/src/main/java/fi/nls/oskari/control/layer/GetDataProviderHandler.java
mikkomustikkamaa: Fixed but left mentioned combination of save and delete to future

control-base/src/main/java/fi/nls/oskari/control/layer/GetDataProviderHandler.java
mikkomustikkamaa: Fixed. This is nice and makes code cleaner in the handler

PR #515, opened 2020-01-28
ProjectinHelper doesn't function like it should

osqq:
ProjectionHelper is not currently working like it was intended so we have to put in condition for EPSG:3879 for it to flip those geometries like it is supposed to. Not sure if it's the case with other CRS but this one is tested and used a lot.

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/feature/AbstractFeatureHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/feature/AbstractFeatureHandler.java
@@ -132,7 +132,8 @@ public abstract class AbstractFeatureHandler extends RestActionHandler {
             }
             JSONArray data = geometries.getJSONArray("data");
             Geometry geometry = getGeometry(geometryType, data, getSrid(srsName, 3067));
-            if(ProjectionHelper.isFirstAxisNorth(crs)) {
+            //TODO: ProjectinHelper doesn't function like it should
+            if(ProjectionHelper.isFirstAxisNorth(crs) || srsName.equals("EPSG:3879")) {
                 // reverse xy
                 ProjectionHelper.flipFeatureYX(geometry);
             }


ZakarFin:
What happens if you add a test for this here https://github.com/oskariorg/oskari-server/blob/master/service-map/src/test/java/fi/nls/oskari/map/geometry/ProjectionHelperTest.java#L231?

```
assertTrue("First axis north EPSG:3879 ", ProjectionHelper.isFirstAxisNorth(CRS.decode("EPSG:3879")));
```

For me it says everything is ok with the projectionhelper.

osqq:
When debugging the content-editor I got this output from `crs.getCoordinateSystem(`). 
```
CartesianCS["Cartesian 2D CS. Axes: northing, easting (N,E). Orientations: north, east. UoM: m.",
  AXIS["Easting", EAST],
  AXIS["Northing", NORTH],
```
So it first says it's (N,E) but ends up giving `AxisDirection[EAST]` from `crs.getCoordinateSystem().getAxis(0).getDirection().absolute()`.

ZakarFin:
Hmm, ok. Well lets not hardcode a single EPSG anyway.

control-base/src/main/java/fi/nls/oskari/control/feature/AbstractFeatureHandler.java
```diff
@@ -132,7 +132,8 @@ protected Feature getFeature(JSONObject jsonObject) throws ActionParamsException
             }
             JSONArray data = geometries.getJSONArray("data");
             Geometry geometry = getGeometry(geometryType, data, getSrid(srsName, 3067));
-            if(ProjectionHelper.isFirstAxisNorth(crs)) {
+            //TODO: ProjectinHelper doesn't function like it should
+            if(ProjectionHelper.isFirstAxisNorth(crs) || srsName.equals("EPSG:3879")) {
```
ZakarFin: Please change this to something like PropertyUtil.getOptional("actionhandler.AbstractFeatureHandler.forceXY", false) and add `actionhandler.AbstractFeatureHandler.forceXY=true` to oskari-ext.properties where needed.

PR #516, opened 2020-01-29
Add logic to delete dataprovider

mikkomustikkamaa:
Pull request contains logic to delete data provider using newly created DataProvider action route. Map layers belonging to the data provider are deleted if mandatory deleteLayers parameter is set to true. With value false, the data provider is removed from map layers. Previously used DeleteOrganization action route is marked as deprecated.

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/layer/DataProviderHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/layer/DataProviderHandler.java
@@ -1,6 +1,8 @@
 package fi.nls.oskari.control.layer;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import org.oskari.log.AuditLog;
 import org.oskari.service.util.ServiceFactory;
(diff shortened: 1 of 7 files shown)


service-control/src/main/java/fi/nls/oskari/control/ActionParameters.java
```diff
@@ -150,6 +150,22 @@ public double getRequiredParamDouble(final String key) throws ActionParamsExcept
...
         }
     }
+    /**
+     * Returns a cleaned up (think XSS) value for the requested parameter
+     * @param key parameter name
+     * @return cleaned up value for the parameter as boolean
+     * @throws ActionParamsException if parameter is not found, is empty or can't be parsed as double
+     */
+    public boolean getRequiredParamBoolean(final String key) throws ActionParamsException {
+        final String errMsg = "Required parameter '" + key + "' missing!";
```
ZakarFin: You don't need to give this string (even as other similar do) since it's the default message for getRequiredParam(key).

service-map/src/main/java/fi/nls/oskari/map/layer/OskariLayerServiceMybatisImpl.java
```diff
@@ -303,6 +310,23 @@ public OskariLayer find(int id) {
         }
         return Collections.emptyList();
     }
+    
+    @Override
+	public List<OskariLayer> findByDataProviderId(int dataProviderId) {
+    	LOG.debug("Find by data provider id: " + dataProviderId);
+        final SqlSession session = factory.openSession();
+        try {
```
ZakarFin: You could do try-with-resources and get rid of the finally statement:
```
try (SqlSession session = factory.openSession()) { ... } 
```

service-base/src/main/java/fi/nls/oskari/domain/map/OskariLayer.java
```diff
@@ -28,7 +28,7 @@
 
     private boolean isBaseMap = false;
     private boolean isInternal = false;
-    private int dataproviderId;
+    private Integer dataproviderId;
```
ZakarFin: I might end up removing the dataproviderId variable completely on the admin refactoring but it's ok for now.

control-base/src/main/java/fi/nls/oskari/control/layer/DataProviderHandler.java
```diff
@@ -17,19 +25,97 @@
...
+
+		if (!dataProviderService.hasPermissionToUpdate(params.getUser(), id)) {
+			throw new ActionDeniedException("Unauthorized user tried to remove data provider id=" + id);
+		}
+
+		if (deleteLayers) {
+			deleteDataProviderAndLayers(id, params);
+		} else {
+			deleteDataProvider(id, params);
+		}
```
ZakarFin: It would be easier to instantly see whats going on if there was a method deleteLayersForProvider()  and removeProviderFromLayers() which would be called based on the boolean. After the conditional statements you could call deleteDataProvider() that would return the removed provider. Then you could write the response based on that in handleDelete.

control-base/src/main/java/fi/nls/oskari/control/layer/DataProviderHandler.java
```diff
@@ -17,19 +25,97 @@
...
+	 * @throws ActionException
+	 */
+	private void deleteDataProviderAndLayers(int id, ActionParameters params) throws ActionException {
+		try {
+			DataProvider organization = dataProviderService.find(id);
+			if (organization == null) {
+				throw new ActionParamsException("Dataprovider not found with id " + id);
+			}
+			// cascade in db will handle that layers are deleted
+			dataProviderService.delete(id);
```
ZakarFin: Wonder if we should write audit log for layers being deleted this way. I'm not sure but maybe the auditlog call below could indicate that layers were removed as well. You can add .withMsg() next to the .withParam() call. Maybe with the handleDelete refactoring we could AuditLog on handleDelete to make it clear we are logging the deletion.

mikkomustikkamaa: Added deleted layer names to audit logging.

service-map/src/main/java/fi/nls/oskari/map/layer/OskariLayerServiceMybatisImpl.java
mikkomustikkamaa: Fixed

service-control/src/main/java/fi/nls/oskari/control/ActionParameters.java
mikkomustikkamaa: Fixed

control-base/src/main/java/fi/nls/oskari/control/layer/DataProviderHandler.java
mikkomustikkamaa: Fixed but little bit different way than suggested.

PR #520, opened 2020-02-04
Get property for flipping coordinates if necessary for content-editor

osqq:
Get property `actionhandler.AbstractFeatureHandler.forceXY` if coordinates are necessary to flip when creating new geometries with content-editor. This happens f.ex. when Geoserver is in the same Jetty as Oskari, and the used CRS is (N,E), like EPSG:3879.


ZakarFin:
Related to #515

PR #521, opened 2020-02-06
Add migration to fix styles

ZakarFin:
Somewhere along the line the admin-functionality has saved null values as "null" strings to styles. This fixes the issue.


ZakarFin:
Assumes no-one actually has a style named "null" but I'll take my chances.

PR #527, opened 2020-02-24
Handle missing locale

ZakarFin:
Fixes an issue where null locale resulted in npe and `{error: "Unhandled exception occured"}` response for frontend. Happens while frontend doesn't validate layer fields perfectly and user hasn't given any end-user/UI name for the layer (locale).


jampukka:
Simple enough

PR #528, opened 2020-02-24
Remove dependency from IOHelper

ZakarFin:
There's bunch of methods on IOHelper handling the same things as this dependency. It's not needed so removing:

            <dependency>
                <groupId>com.github.kevinsawicki</groupId>
                <artifactId>http-request</artifactId>
                <version>4.0</version>
            </dependency>


jampukka:
Looks good, hopefully there aren't any server extensions that depend on this or the IOHelper functions removed. Should mention this in the 1.55 notes for those writing server extensions just in case.

PR #529, opened 2020-02-24
Add parser for querystring

ZakarFin:
Note! Includes commits from #528.

Add methods to IOHelper to parse query string to allow easy manipulation of urls. Use methods to only filter out problematic parameters when validating layer data.


jampukka:
Looks good

PR #558, opened 2020-03-23
Remove fetching namescpaceURL from FeatureWFSTRequestBuilder and inst…

osqq:
…ead handle namespace in InsertFeatureHandler

- remove layer's namespace from InsertFeature
- take namespace name from potential prefix and add it to layerURL

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/feature/InsertFeatureHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/feature/InsertFeatureHandler.java
@@ -53,6 +53,7 @@ public class InsertFeatureHandler extends AbstractFeatureHandler {
             for (int i = 0; i < paramFeatures.length(); i++) {
                 JSONObject featureJSON = paramFeatures.getJSONObject(i);
                 OskariLayer layer = getLayer(featureJSON.optString("layerId"));
+                movePrefixFromNameToURL(layer);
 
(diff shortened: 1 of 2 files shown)


osqq:
closed due to https://github.com/oskariorg/oskari-server/pull/557

control-base/src/main/java/fi/nls/oskari/control/feature/InsertFeatureHandler.java
```diff
@@ -125,5 +126,21 @@ private String parseFeatureIdFromResponse(String response)
...
+
+    /**
+     * Takes workspace prefix from layer name (before ':') and puts it into the layer url after 'geoserver/'
+     * @param  layer  OskariLayer layer
+     */
+    private void movePrefixFromNameToURL(OskariLayer layer){
+        String layerName = layer.getName();
+        String url = layer.getUrl();
+        if(layerName.indexOf(":") != -1){
+            url = url.replace("geoserver/","geoserver/"+(layerName.split(":")[0])+"/");
```
ZakarFin: This is a very specific thing to do. First of all it expects geoserver/ to be in the url. Then with an example layer name of oskari:userlayer
- URL: https://something/geoserver/ows would be changed to https://something/geoserver/oskari/ows
- URL: https://something/geoserver/oskari/wfs would be changed to https://something/geoserver/oskari/oskari/wfs

Maybe this needs some refining?

PR #563, opened 2020-03-30
Pxweb parse data model

okauppinen:
I don't know how these parsers should work for different languages but at least there was a bug in setting datamodel. getDataModel doesn't return null so the model wasn't never populated from table.

https://github.com/oskariorg/oskari-server/blob/develop/service-statistics-common/src/main/java/fi/nls/oskari/control/statistics/data/StatisticalIndicator.java#L120

### DIFF
--- a/service-statistics-pxweb/src/main/java/fi/nls/oskari/control/statistics/plugins/pxweb/parser/PxwebIndicatorsParser.java
+++ b/service-statistics-pxweb/src/main/java/fi/nls/oskari/control/statistics/plugins/pxweb/parser/PxwebIndicatorsParser.java
@@ -82,13 +82,11 @@ public class PxwebIndicatorsParser {
                         ind.setId(indicatorId);
                         indicatorMap.put(indicatorId, ind);
                         indicators.add(ind);
-                    }
-                    ind.addName(lang, item.text);
-                    if (ind.getDataModel() == null) {
                         // only populate model for first (== primary) language as it doesn't support localized labels for variables/selectors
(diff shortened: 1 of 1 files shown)


service-statistics-pxweb/src/main/java/fi/nls/oskari/control/statistics/plugins/pxweb/parser/PxwebIndicatorsParser.java
```diff
@@ -82,13 +82,11 @@ public PxwebIndicatorsParser(PxwebConfig config) {
                         ind.setId(indicatorId);
                         indicatorMap.put(indicatorId, ind);
                         indicators.add(ind);
-                    }
-                    ind.addName(lang, item.text);
-                    if (ind.getDataModel() == null) {
```
ZakarFin: Replace null check with:
```
if (ind.getDataModel().getDimensions().isEmpty()) {
     ind.setDataModel(getModel(table));
}
```

If the null check is removed entirely the indicator params will have the last language that was supported by the API. So for example in paikkatietoikkuna this would mean that the Helsinki datasource would have indicator names in finnish but the selectors/parameters in english or swedish. Since we don't have the selector merging implemented yet the current code tries to have the selectors with the primary language of the geoportal. This leads to indicator name being localized but the selectors always in finnish on paikkatietoikkuna.

ZakarFin: Nevermind, this is ok as the datamodel is only created for new indicators.

PR #574, opened 2020-04-22
Support OAPIF layers in LayerCapabilitiesHelper.updateCapabilities(OskariLayer)

jampukka:
Fix an issue with updating the capabilities of OAPIF layer(s). Currently saving works because it goes through `SaveLayerHandler` which has special handling for OAPIF layers. Move that logic down to LayerCapabilitiesHelper.


ZakarFin:
Fixes an issue with OGC API Features layers with the new layereditor bundle that uses LayerAdmin action route instead of SaveLayer

PR #581, opened 2020-05-08
Use IOHelper for querystring parsing

ZakarFin:
I'll add some tests so draft PR. Not sure if we should accept urls for metadata at this point or normalize the data for db instead.

### DIFF
--- a/service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatter.java
+++ b/service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatter.java
@@ -147,7 +147,7 @@ public class LayerJSONFormatter {
         JSONHelper.putValue(layerJson, "created", layer.getCreated());
         JSONHelper.putValue(layerJson, "updated", layer.getUpdated());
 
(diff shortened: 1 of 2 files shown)


service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatter.java
```diff
@@ -220,29 +219,36 @@ public static JSONObject createStylesJSON(String name, String title, String lege
...
-                }
-            } catch (Exception ignored) {
-                // propably just not valid URL
+        if(!metadataId.toLowerCase().startsWith("http")) {
+            // not a url -> return as is
+            return metadataId;
+        }
+        try {
+            Map<String, List<String>> params = IOHelper.parseQuerystring(metadataId);
+            String idParam = params.keySet().stream()
```
jampukka, line 232: For fun `return IOHelper.parseQuerystring(metadataId).entrySet().stream().filter(e -> "uuid".equalsIgnoreCase(e.getKey()) || KEY_ID.equalsIgnoreCase(e.getKey())).map(e -> e.getValue()).findAny().map(it -> it.stream()).orElse(null);`

PR #584, opened 2020-05-20
UserDataStyle migration to json

okauppinen:
Migrate user_layer_style, analysis_style, categories style columns to layers' options column as json.
Populate base wfs layer (oskari_maplayer) options (clusteringDistance, renderMode, labelProperty).

WFSLayerOptions is mainly used with UserDataLayer which contains only one style named 'default' which contains all style definitions. UserDataLayer injects baselayer options to layer options and if options has labelProperty then default text style is added. I don't know if this is proper solution but at least different layer builders/formatters doesn't need customization.

Normal WFS layers may contain more than one style with partial style definitions. getNamedStyle takes Oskari default style as base definitions. At least print will use this to get style for WFS layers.

Actual usage of json styles will be separated to own pull requests and flyway to remove style columns/tables.

### DIFF
--- a/content-resources/src/main/java/fi/nls/oskari/util/UserDataStyleMigrator.java
+++ b/content-resources/src/main/java/fi/nls/oskari/util/UserDataStyleMigrator.java
@@ -1,6 +1,10 @@
 package fi.nls.oskari.util;
 
+import fi.nls.oskari.db.DatasourceHelper;
+import fi.nls.oskari.domain.map.wfs.WFSLayerOptions;
 import org.json.JSONObject;
+
+import javax.sql.DataSource;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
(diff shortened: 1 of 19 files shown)


ZakarFin:
Not sure how I feel about getters in WFSLayerOptions modifying the data but it's a nice way of working around missing parts for the deep JSON structure. Looks good so far but not merging since there's some more changes that Olli wants to add here.

Also if these migrations are run before having the changes on how/where styles are loaded from/modified to you might get into a situation where modifying the style changes the values in the old table structure and those changes will be rolled back once other parts of the code is updated to match the new structure.

content-resources/src/main/java/fi/nls/oskari/util/UserDataStyleMigrator.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class UserDataStyleMigrator {
+    private static final String CLUSTERING_PROPERTY = ".clustering.distance";
+    private static final String LAYER_ID_PROPERTY = ".baselayer.id";
+    private static final String DEFAULT_RENDER_MODE = "vector";
+
+    public static int migrateStyles (Connection conn, final String layerTable, final String styleTable, final String styleIdColumn) throws SQLException {
+        Map<Long,String> options = getOptions(conn, styleTable);
```
ZakarFin, line 21: Readability of this file would benefit from re-arranging the method order. Maybe move migrateStyles after updateBaseLayerOptions, make updateLayers() private and move it to the end if it's only used from this file?

PR #586, opened 2020-05-22
PrintService use vector layers options to get print style

okauppinen:
To support changes made in:
https://github.com/oskariorg/oskari-server/pull/584
https://github.com/oskariorg/oskari-server/pull/585

Also added support for rgb(0,0,0) and rgba (255,255,255,1) colors. rgba alpha doesn't seem to affect but at least print doesn't fail. Also these aren't documented in Oskari style but frontend supports them.

Removed hard coded text label offsets and added basic support (left and center align) for getting alignment and offsets from style.


service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -735,22 +735,22 @@ private static void draw(PDPageContentStream stream, Geometry g, PDPrintStyle st
...
             c = getLineCentroid ((LineString) g);
-            drawLabel(stream, c , label);
+
         } else if (g instanceof MultiLineString) {
             for (int i = 0; i < g.getNumGeometries(); i++) {
                 c = getLineCentroid ((LineString) g.getGeometryN(i));
-                drawLabel(stream, c, label);
             }
         }
+        if (c == null) return;
```
jampukka, line 752: This could be `if (c != null) { drawLabel(stream, c, align, label); }` instead

service-print/src/main/java/org/oskari/print/util/ColorUtil.java
```diff
@@ -1,16 +1,40 @@
...
 
 public class ColorUtil {
-    public static Color parseColor(String rrggbb) {
-        if (rrggbb == null || rrggbb.isEmpty()) {
+    // rgb(255,0,0)
+    private static final Pattern RGB = Pattern.compile("rgb *\\( *([0-9]+), *([0-9]+), *([0-9]+) *\\)");
+    // rgba(255,0,0,0.5)
+    private static final Pattern RGBA = Pattern.compile("rgba *\\( *([0-9]+), *([0-9]+), *([0-9]+), *([0,1][.]?[0-9]*) *\\)");
+
+    public static Color parseColor(String color) {
```
jampukka, line 13: Please add unit test for this

service-print/src/main/java/org/oskari/print/util/StyleUtil.java
```diff
@@ -222,38 +228,17 @@ private static PDColor createFillPattern(PDResources resources, int fillPattern,
...
-        return json;
+    public static Matrix getMatrixForLabel(Coordinate c, PDPrintStyle.LabelAlign align, PDFont font, float size, String label) throws IOException {
+        float x = align.getX();
+        float y = align.getY();
+        switch(align.getAlign()) {
+            case "center":
+                x -=  PDFBoxUtil.getTextWidth(label, font, size) / 2;
+                break;
+            case "left":
+                x += ICON_SIZE/2;
```
jampukka, line 239: Why is left offset by ICON_SIZE/2 but center is (the center of the text) at the coordinate ?

service-print/src/main/java/org/oskari/print/request/PDPrintStyle.java
```diff
@@ -108,4 +119,19 @@ public boolean hasLineColor () {
     public boolean hasLabels () {
         return labelProperty != null && !labelProperty.isEmpty();
     }
+
+    public static class LabelAlign {
+        private float x;
+        private float y;
+        private String align;
```
jampukka, line 126: Maybe enum instead of String?

service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -735,22 +735,22 @@ private static void draw(PDPageContentStream stream, Geometry g, PDPrintStyle st
         }
     }
 
-    private static void drawLabel(PDPageContentStream stream, Geometry g, String label) throws IOException {
-        Coordinate c;
+    private static void drawLabel(PDPageContentStream stream, PDPrintStyle.LabelAlign align, Geometry g, String label) throws IOException {
+        Coordinate c = null;
         if (g instanceof MultiPoint || g instanceof MultiPolygon ) {
             for (int i = 0 ; i < g.getNumGeometries(); i++){
                 c = g.getGeometryN(i).getCentroid().getCoordinate();
```
jampukka, line 742: This actually makes no sense - this is picking the centroid of the last point or polygon in the multipoint/-polygon.

service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -735,22 +735,22 @@ private static void draw(PDPageContentStream stream, Geometry g, PDPrintStyle st
...
-                drawLabel(stream, c , label);
             }
         } else if (g instanceof LineString) {
             c = getLineCentroid ((LineString) g);
-            drawLabel(stream, c , label);
+
         } else if (g instanceof MultiLineString) {
             for (int i = 0; i < g.getNumGeometries(); i++) {
                 c = getLineCentroid ((LineString) g.getGeometryN(i));
```
jampukka, line 749: Same comment as line 742

service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -735,22 +735,22 @@ private static void draw(PDPageContentStream stream, Geometry g, PDPrintStyle st
         }
     }
 
-    private static void drawLabel(PDPageContentStream stream, Geometry g, String label) throws IOException {
-        Coordinate c;
+    private static void drawLabel(PDPageContentStream stream, PDPrintStyle.LabelAlign align, Geometry g, String label) throws IOException {
```
jampukka, line 738: Having two functions with the same name but similiar parameters in mixed order is confusing:
`void drawLabel(PDPageContentStream stream, PDPrintStyle.LabelAlign align, Geometry g, String label)`
`void drawLabel(PDPageContentStream stream, Coordinate c, PDPrintStyle.LabelAlign align, String label)`
Either:
`void drawLabel(PDPageContentStream stream, PDPrintStyle.LabelAlign align, Geometry g, String label)`
`void drawLabel(PDPageContentStream stream, PDPrintStyle.LabelAlign align, Coordinate c, String label)`
Or:
`void drawLabel(PDPageContentStream stream, Geometry g, PDPrintStyle.LabelAlign align, String label)`
`void drawLabel(PDPageContentStream stream, Coordinate c, PDPrintStyle.LabelAlign align, String label)`
Or rename either one to `drawLabelAt` or something like that.

service-print/src/main/java/org/oskari/print/request/PDPrintStyle.java
```diff
@@ -92,6 +94,15 @@ public void setIcon(PDFormXObject icon) {
...
         this.labelProperty = labelProperty;
     }
+    public LabelAlign getLabelAlign () {
+        return labelAlign;
+    }
+    public void setLabelAlign (LabelAlign labelAlign) {
+        this.labelAlign = labelAlign;
+    }
+    public void setLabelAlign (String align, float x, float y) {
+        this.labelAlign = new LabelAlign(align, x, y);
```
jampukka, line 104: Since LabelAlign only has one public constructor that takes these exact same parameters in the same order this utility setter method adds very little value. Also this is the only utility setter method in the class so it's a bit out of place as the rest of the class is basically a dumb data holder.

service-print/src/main/java/org/oskari/print/request/PDPrintStyle.java
```diff
@@ -108,4 +119,19 @@ public boolean hasLineColor () {
     public boolean hasLabels () {
         return labelProperty != null && !labelProperty.isEmpty();
     }
+
+    public static class LabelAlign {
+        private float x;
```
jampukka, line 124: These are offsets, right? Should they be called `offsetX` and `offsetY` instead of `x` and `y`?

service-print/src/main/java/org/oskari/print/util/StyleUtil.java
```diff
@@ -222,38 +228,17 @@ private static PDColor createFillPattern(PDResources resources, int fillPattern,
...
-        JSONHelper.putValue(stroke, "area", strokeArea);
-        JSONHelper.putValue(json, "stroke", stroke);
-        JSONObject fill = new JSONObject();
-        JSONHelper.putValue(fill, "color", "#FAEBD7");
-        JSONObject fillArea = new JSONObject();
-        JSONHelper.putValue(fillArea, "pattern", -1);
-        JSONHelper.putValue(fill, "area", fillArea);
-        JSONHelper.putValue(json, "fill", fill);
-        return json;
+    public static Matrix getMatrixForLabel(Coordinate c, PDPrintStyle.LabelAlign align, PDFont font, float size, String label) throws IOException {
```
jampukka, line 231: This functions has a lot of duties.
What about adding two functions:
 `float getLabelX(float x, float width);`
 `float getLabelY(float y, float height);` 
to `PDPrintStyle.LabelAlign`? `LabelAlign` class would handle alignment (left, center, offsets) and the caller would control the location (other than aligning) and the size of the label.

service-print/src/main/java/org/oskari/print/PDF.java
jampukka: If that's the desired functionality then do something like `if (g.isEmpty()) { return; } g = g.getGeometryN(g.getNumGeometries() - 1); // Now g is Point, LineString or Polygon (unless it was GeometryCollection, should check that first)`

okauppinen: This was mistake. Looked that there was copy-paste code. Rolled back.

service-print/src/main/java/org/oskari/print/util/StyleUtil.java
okauppinen: Accidentally looked that openlayers sets label outside from icon with 'left' align. Tested again and looked examples. openlayers doesn't consider icon size. Removed this and added offsetX for markers

PR #596, opened 2020-06-11
Revert "Migrations in single transaction"

ZakarFin:
Reverts oskariorg/oskari-server#572

This breaks the empty database initialization. Problematic migrations are:
- flyway/oskari/V1_37_1__publisher2_migration
- flyway/oskari/V1_39_1__migrate_publish_template_ol3

which both can be skipped but after that there's a problem with the json setup files with inserting layers and not finding the referenced dataprovider.

So reverting the change for now.


jampukka:
LGTM

PR #599, opened 2020-06-16
Avoid NPEs in OskariStyle handling in service-print

jampukka:
Current code can't handle incomplete OskariStyle JSON. Add a bit of code to avoid null pointer exceptions.

### DIFF
--- a/service-print/src/main/java/org/oskari/print/util/StyleUtil.java
+++ b/service-print/src/main/java/org/oskari/print/util/StyleUtil.java
@@ -8,7 +8,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.*;
 
-import com.vividsolutions.jts.geom.Coordinate;
 import fi.nls.oskari.util.IOHelper;
 import org.apache.batik.transcoder.TranscoderException;
 import org.apache.batik.transcoder.TranscoderInput;
@@ -18,7 +17,6 @@ import org.apache.pdfbox.cos.COSName;
 import org.apache.pdfbox.multipdf.LayerUtility;
 import org.apache.pdfbox.pdmodel.PDDocument;
 import org.apache.pdfbox.pdmodel.PDPage;
-import org.apache.pdfbox.pdmodel.font.PDFont;
 import org.apache.pdfbox.pdmodel.graphics.form.PDFormXObject;
 import org.apache.pdfbox.pdmodel.PDPatternContentStream;
 import org.apache.pdfbox.pdmodel.PDResources;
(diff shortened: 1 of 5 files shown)


service-print/src/main/java/org/oskari/print/util/StyleUtil.java
```diff
@@ -59,7 +56,7 @@
 
     public static PDPrintStyle getLineStyle (JSONObject oskariStyle) {
         PDPrintStyle style = new PDPrintStyle();
-        JSONObject stroke = JSONHelper.getJSONObject(oskariStyle, "stroke");
+        JSONObject stroke = JSONHelper.optJSONObject(oskariStyle, "stroke", () -> new JSONObject());
```
ZakarFin, line 59: Wonder if this would work/read better if PDPrintStyle had the default values itself and we could just return early instead of creating new objects to make the rest of the code work properly.

service-print/src/main/java/org/oskari/print/util/StyleUtil.java
```diff
@@ -69,11 +66,13 @@ public static PDPrintStyle getLineStyle (JSONObject oskariStyle) {
...
         PDPrintStyle style = new PDPrintStyle();
-        JSONObject area = JSONHelper.getJSONObject(JSONHelper.getJSONObject(oskariStyle, "stroke"), "area");
-        setStrokeStyle(style, area);
+        JSONObject stroke = JSONHelper.optJSONObject(oskariStyle, "stroke", () -> new JSONObject());
+        JSONObject strokeArea = JSONHelper.optJSONObject(stroke, "area", () -> new JSONObject());
+        setStrokeStyle(style, strokeArea);
 
-        JSONObject fill = JSONHelper.getJSONObject(oskariStyle, "fill");
-        int pattern = JSONHelper.getJSONObject(fill, "area").optInt("pattern");
+        JSONObject fill = JSONHelper.optJSONObject(oskariStyle, "fill", () -> new JSONObject());
```
ZakarFin, line 73: You could skip all the lines up till setLabelStyle() if fill is null here? That way it would be more clear that setFillPattern() is not called. Though didn't check if style.setFillColor() does something with null param. optInt() returns 0 for null but since color will be null the setFillPattern() is not called. If the lines are just skipped anyone reading this code in the future wouldn't need to check what  optInt() and parseColors() do for nulls to figure out what is actually set for the style here.

service-print/src/main/java/org/oskari/print/util/StyleUtil.java
```diff
@@ -86,9 +85,10 @@ public static PDPrintStyle getPolygonStyle (JSONObject oskariStyle, PDResources
...
         PDPrintStyle style = new PDPrintStyle();
-        JSONObject image = JSONHelper.getJSONObject(oskariStyle, "image");
+        JSONObject image = JSONHelper.optJSONObject(oskariStyle, "image", () -> new JSONObject());
         int shape = image.optInt("shape", 5); // External icons not supported
-        String fillColor = JSONHelper.optString(JSONHelper.getJSONObject(image, "fill"), "color");
+        JSONObject imageFill = JSONHelper.optJSONObject(image, "fill", () -> new JSONObject());
+        String fillColor = JSONHelper.optString(imageFill, "color");
         style.setFillColor(ColorUtil.parseColor(fillColor));
         int size = image.optInt("size", 3);
         style.setIcon(getIcon(doc, shape, fillColor, size));
```
ZakarFin, line 94: This will receive null as fillColor which is used as a param for `replace("$fill", fillColor)`. Maybe it should use a default color instead or is "null"/unrecognized color black or something like that?

PR #603, opened 2020-06-24
Clean up AbstractWFSFeaturesHandler

jampukka:
Remove `if (!OskariLayer.TYPE_WFS.equals(layer.getType())) { return layer; }` block that always ends up in an exception. Include `AbstractWFSFeaturesHandler#requireWFSLayer` function calls to `AbstractWFSFeaturesHandler#findLayer` so that inherited classes don't have to remember to call it. Split the logic of finding the layer into to more separate branches for improved readability.


ZakarFin:
Lets wait for 2.0/2.1 for this.

PR #604, opened 2020-07-02
Fixes for printout

ZakarFin:
Problems in 1.56.0 (might be in earlier versions as well):

- [x]  Lines are always printed out in default dark color
- [ ]  Layers with mixed geometries in for example my places are ignored by print if multiple geometry types are in the print viewport (works if only one geometry type is on "screen"/the viewport we are printing)
- [x] Myplaces style is cached in printing. If user prints myplaces layer -> changes the style -> prints again the printout will use the earlier style.


ZakarFin:
The mixed geometries hasn't been solved in this pr but we'll keep working on it. Meanwhile it's good enough for fixing 2/3 and get them merged.

PR #621, opened 2020-08-28
Helper refactoring

ZakarFin:
Remove org/oskari/helpers/FlywayHelper as we already had org/oskari/helpers/AppsetupHelper that is all about helping Flyway-migrations regarding app setups. Moved all the functions to AppSetupHelper and renamed them from view-something to app-something.


ZakarFin:
I think the Appsetup and BundleHelpers should be removed and the functions that are generally needed in Flyway-migrations should be added to the actual mybatis services instead. The Mybatis services can usually be used in application-specific Flyway-migrations safely as the core database is migrated before application modules are migrated == database will be in the state Mybatis services in the same code version expect them to be in.

PR #624, opened 2020-09-01
Import Spring dependencies as Bill Of Materials and update.

tsallinen:
Import Spring dependencies as Bill Of Materials (BOM). This allows application specific Oskari Server Extensions to include additional Spring Framework/Security/Session libraries without the need to (re)define version numbers.

Also updated Spring to latest patch versions.


ZakarFin:
Thanks!

PR #642, opened 2020-09-18
PrintHandler get custom styles from payload

okauppinen:
https://github.com/oskariorg/oskari-frontend/pull/1334


control-base/src/main/java/fi/nls/oskari/control/data/GetPrintHandler.java
```diff
@@ -251,14 +250,18 @@ private void setFormat(String formatStr, PrintRequest req)
...
-        // set custom styles
-        if (customStyles != null){
-            printLayers.forEach(l -> {
-                String id = l.getLayerId();
-                if (customStyles.has(id)) {
-                    l.setCustomStyle(JSONHelper.getJSONObject(customStyles, id));
-                }
-            });
+        // set custom stylea
+        if (params.getRequest().getMethod().equals("POST")) {
```
ZakarFin, line 254: Perhaps add a method that accepts both param and payload and returns the first one with value? Then we can merge this before we decide about the frontend.

```
/**
 * Returns custom styles from param or payload or null if not present.
 */
private JSONObject getCustomStyles(ActionParameters params) {
        JSONObject customStyles = params.getHttpParamAsJSON(PARM_CUSTOM_STYLES);
        if (customStyles != null) {
            return customStyles;
        }
        if ("POST".equals(params.getRequest().getMethod())) {
            try {
                JSONObject payload = params.getPayLoadJSON();
                return payload.optJSONObject(PARM_CUSTOM_STYLES);
            } catch(ActionParamsException ignored) {
                LOG.ignored(ignored, "Payload not included in request");
            }
        }
        return null;
}
```

PR #643, opened 2020-09-23
WFSFilter draft version

okauppinen:


service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSFilter.java
```diff
@@ -0,0 +1,132 @@
...
+                    .collect(Collectors.toList());
+            return ff.or(values);
+        }
+        if (property.has(KEY_NOT_IN)) {
+            List <Filter> values = JSONHelper.getArrayAsList(JSONHelper.getJSONArray(property,KEY_NOT_IN))
+                    .stream()
+                    .map(value -> ff.notEqual(name, ff.literal(value), match))
+                    // .map(filter -> property.has(KEY_NOT_IN) ? ff.not(filter):filter)
+                    .collect(Collectors.toList());
+            return ff.or(values);
```
jampukka, line 104: `NOT IN (A, B) <=> NOT (A OR B) <=> NOT A AND NOT B` is not equivalent to `NOT A OR NOT B`!

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSFilter.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+        if (property.has(KEY_GREATER)) {
+            Filter greater = ff.greater(name, ff.literal(JSONHelper.get(property, KEY_GREATER)));
+            if (property.has(KEY_LESS)) {
+                Filter less = ff.less(name, ff.literal(JSONHelper.get(property, KEY_LESS)));
+                return ff.and(greater, less);
+            }
+            if (property.has(KEY_LESS_OR_EQUAL)) {
+                Filter lessEqual = ff.lessOrEqual(name, ff.literal(JSONHelper.get(property, KEY_LESS_OR_EQUAL)));
+                return ff.and(greater, lessEqual);
```
jampukka, line 87: `prop > x AND prop  <= x` is always false. Same value can't be greater than x and less than or equal to x at the same time.

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSFilter.java
```diff
@@ -0,0 +1,132 @@
...
+        boolean match = JSONHelper.getBooleanFromJSON(property,KEY_MATCH_CASE, true);
+        Expression name = ff.property(key);
+        if (property.has(KEY_EQUAL)) {
+            return ff.equal(name, ff.literal(JSONHelper.get(property, KEY_EQUAL)), match);
+        }
+        if (property.has(KEY_GREATER)) {
+            Filter greater = ff.greater(name, ff.literal(JSONHelper.get(property, KEY_GREATER)));
+            if (property.has(KEY_LESS)) {
+                Filter less = ff.less(name, ff.literal(JSONHelper.get(property, KEY_LESS)));
+                return ff.and(greater, less);
```
jampukka, line 83: prop > x AND prop < x is always false. Same value can't be greater than x and less than x at the same time.

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSFilter.java
```diff
@@ -0,0 +1,132 @@
...
+            List <String> values = JSONHelper.getArrayAsList(JSONHelper.getJSONArray(property, KEY_NOT_LIKE));
+            String value = JSONHelper.optString(property, KEY_NOT_LIKE, "");
+            if (!value.isEmpty()) {
+                values = Collections.singletonList(value);
+            }
+            List<Filter> filters = values.stream()
+                    .map(val -> ff.like (name, val))
+                    .map(f -> ff.not(f))
+                    .collect(Collectors.toList());
+            return ff.or(filters);
```
jampukka, line 127: See NOT IN

service-wfs-client/src/main/java/org/oskari/service/wfs/client/WFSFilter.java
```diff
@@ -0,0 +1,213 @@
...
+        if (key != null) {
+            return getPropertyFilter();
+        }
+        if (property !=null) {
+            return property.getPropertyFilter();
+        }
+        if (and != null && !and.isEmpty()) {
+            List<Filter> andFilters = and.stream().map(f -> f.getFilter()).collect(Collectors.toList());
+            if (!andFilters.isEmpty()) {
+                return ff.and(andFilters);
```
ZakarFin, line 153: Shouldn't we still check for existance of or-filters instead of returning here?

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSFilter.java
```diff
@@ -0,0 +1,56 @@
...
+import fi.nls.oskari.domain.map.wfs.WFSLayerCapabilities;
+import fi.nls.oskari.service.ServiceRuntimeException;
+import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.referencing.CRS;
+import org.json.JSONObject;
+import org.opengis.filter.Filter;
+import org.opengis.filter.FilterFactory;
+
+public class OskariWFSFilter {
```
ZakarFin, line 16: Rename this to OskariWFSFilterFactory or similar so it's more clear that this is not a filter but something that is used to read/create filters from JSON.

service-wfs-client/src/main/java/org/oskari/service/wfs/client/WFSFilter.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import fi.nls.oskari.service.ServiceRuntimeException;
+import org.geotools.factory.CommonFactoryFinder;
+import org.opengis.filter.Filter;
+import org.opengis.filter.FilterFactory;
+import org.opengis.filter.expression.Expression;
+
+public class WFSFilter {
```
ZakarFin, line 13: Rename this to OskariWFSFilter so it's clear that this is us and not for example GeoTools. Maybe throw in a comment that this is used for tranlating the Oskari specific JSON filter to a GeoTools filter.

service-wfs-client/src/main/java/org/oskari/service/wfs/client/WFSFilter.java
```diff
@@ -0,0 +1,213 @@
...
+            filters = like.stream()
+                    .map(value -> ff.like(name, value))
+                    .collect(Collectors.toList());
+            return filters.isEmpty() ? null : ff.or(filters);
+        }
+        if (notLike != null) {
+            filters = notLike.stream()
+                    .map(value -> ff.like(name, value))
+                    .collect(Collectors.toList());
+            return filters.isEmpty() ? null : ff.and(filters);
```
ZakarFin, line 209: Should there be a not() call here somewhere?

service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSFilterFactory.java
```diff
@@ -0,0 +1,56 @@
...
+        try {
+            return OBJECT_MAPPER.readValue(json, OskariWFSFilter.class);
+        } catch (Exception ex) {
+            throw new ServiceRuntimeException("Coudn't parse filter from: " + json, ex);
+        }
+    }
+    protected static Filter appendFilter(final Filter main, final Filter toAppend) {
+        if (main == null) {
+            return toAppend;
+        } else {
```
ZakarFin, line 50: Should this have:
```
else if (toAppend == null) {
    return main;
}
```
For consistency and convenience

PR #649, opened 2020-10-07
MIF/MID PLINE with numpts on same line

jampukka:
Solves https://github.com/oskariorg/oskari-docs/issues/229

Add support for PLINE with numpts on the same line as PLINE. Three accepted formats for PLINE are:
```
PLINE 
numpts
x y
...
```
```
PLINE numpts
x y
...
```
```
PLINE MULTIPLE numlines
numpts1
x y
...
```


ZakarFin:
Works towards solving the issue but doesn't solve it completely. Merging and adding my test.zip to the issue to make it easier to debug.

PR #650, opened 2020-10-12
Capabilities parsing improvements

ZakarFin:
Split TileMatrixSet indentifier from the last ':' instead of the first one as the ref might be something like "EPSG:3067:0" so this will work:

Limit of "EPSG:3857:0" on layer:
```
            <TileMatrixLimits>
                <TileMatrix>EPSG:3857:0</TileMatrix>
```
Should map to:
```
        <TileMatrixSet>
            <ows:Identifier>EPSG:3857</ows:Identifier>
            ...
            <TileMatrix>
                <ows:Identifier>0</ows:Identifier>
```

Also adding a documentary test case for WMS-layer with empty <SRS></SRS> tag.


jampukka:
LGTM

PR #653, opened 2020-10-20
Fix MIF parser

jampukka:
`skipOptional(String)` is doing the very opposite of what it is supposed to do. Flip the check.


ZakarFin:
Thanks!

PR #656, opened 2020-10-26
Ask one more than limit so we can deduce if there would have been more

ZakarFin:
When getting search results from OSM Nominatim.


jampukka:
"1" + 1 = "11"

ZakarFin:
Thanks :D

PR #663, opened 2020-11-10
Check response status before trying to read image when printing

ZakarFin:
There's a lot of logging for 404 responses when trying to load images for print output. This should help with that a bit.


ZakarFin:
Previously handling added for proxying layers in #657

PR #664, opened 2020-11-16
announcements-bundle

osqq:
- Create table for announcements bundle and register the bundle.
- Actions and Helpers to handle data to-and-from database (announcements table)
- Currently files are on their own folder for clarity (move to another existing folder?)

TODO (?) : 

- Relocate the flyways ?
- Relocate the files to existing folder, where?
- Remove the bundle from additional modules?


osqq:
501 error fixed with correct package pathing

content-resources/src/main/java/flyway/announcements/V1_0_0__register_announcement_bundles.java
```diff
@@ -0,0 +1,25 @@
...
+        Connection connection = context.getConnection();
+		
+		// BundleHelper checks if these bundles are already registered
+        Bundle bundle = new Bundle();
+        bundle.setName("announcements");
+        BundleHelper.registerBundle(connection, bundle);
+
+        Bundle admin = new Bundle();
+        admin.setName("admin-announcements");
+        BundleHelper.registerBundle(connection, admin);
```
ZakarFin, line 23:
```suggestion
        // BundleHelper checks if these bundles are already registered
        BundleHelper.registerBundle(connection, "announcements");
        BundleHelper.registerBundle(connection, "admin-announcements");
```

ZakarFin: This migration should be moved to the oskari/core-module instead of having a new flyway module just for bundle registration.

content-resources/src/main/resources/flyway/oskari/V2_0_7__create_announcements_tabel.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS oskari_announcements (
+    "id" serial NOT NULL,
+    "title" VARCHAR(255) NOT NULL,
```
ZakarFin, line 3: Could this be typed as `text` as well?

control-announcements/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+        <dependency>
+            <groupId>org.oskari</groupId>
+            <artifactId>control-base</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
```
ZakarFin, line 26: I don't see any tests on this module. Add tests or remove dependency.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsRestActionHandler.java
```diff
@@ -0,0 +1,18 @@
...
+
+import fi.nls.oskari.control.ActionDeniedException;
+import fi.nls.oskari.control.ActionException;
+import fi.nls.oskari.control.RestActionHandler;
+import fi.nls.oskari.util.PropertyUtil;
+
+public class AnnouncementsRestActionHandler extends RestActionHandler {
+    private final static String PROPERTY_MODULES = "db.additional.modules";
+
+    public void requireAnnouncementsConfigured() throws ActionException {
```
ZakarFin, line 11: Why would this need to be checked? The database table is created by the core module anyways so it's always there. This can be removed.

control-announcements/src/main/java/org/oskari/announcements/actions/DeleteAnnouncement.java
```diff
@@ -0,0 +1,29 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+@OskariActionRoute("DeleteAnnouncement")
+public class DeleteAnnouncement extends AnnouncementsRestActionHandler {
+    private static Logger LOG = LogFactory.getLogger(DeleteAnnouncement.class);
+
+    @Override
+    public void handlePost(ActionParameters params) throws ActionException {
+        requireAnnouncementsConfigured();
```
ZakarFin, line 19:
```suggestion
        params.requireAdminUser();
```

control-announcements/src/main/java/org/oskari/announcements/actions/GetAdminAnnouncements.java
```diff
@@ -0,0 +1,30 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+@OskariActionRoute("GetAdminAnnouncements")
+public class GetAdminAnnouncements extends AnnouncementsRestActionHandler{
+    private static Logger LOG = LogFactory.getLogger(GetAnnouncements.class);
+
+    @Override
+    public void handleGet(ActionParameters params) throws ActionException {
+        requireAnnouncementsConfigured();
```
ZakarFin, line 19:
```suggestion
        params.requireAdminUser();
```

control-announcements/src/main/java/org/oskari/announcements/actions/SaveAnnouncement.java
```diff
@@ -0,0 +1,30 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+@OskariActionRoute("SaveAnnouncement")
+public class SaveAnnouncement extends AnnouncementsRestActionHandler{
+    private static Logger LOG = LogFactory.getLogger(SaveAnnouncement.class);
+
+    @Override
+    public void handlePost(ActionParameters params) throws ActionException {
+        requireAnnouncementsConfigured();
```
ZakarFin, line 20:
```suggestion
        params.requireAdminUser();
```

control-announcements/src/main/java/org/oskari/announcements/actions/UpdateAnnouncement.java
```diff
@@ -0,0 +1,30 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+@OskariActionRoute("UpdateAnnouncement")
+public class UpdateAnnouncement extends AnnouncementsRestActionHandler{
+    private static Logger LOG = LogFactory.getLogger(SaveAnnouncement.class);
+
+    @Override
+    public void handlePost(ActionParameters params) throws ActionException {
+        requireAnnouncementsConfigured();
```
ZakarFin, line 20:
```suggestion
        params.requireAdminUser();
```

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementParams.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public boolean isNeedCastVarchar() {
+        return needCastVarchar;
+    }
+
+    public void setNeedCastVarchar(boolean needCastVarchar) {
+        this.needCastVarchar = needCastVarchar;
+    }
+}
```
ZakarFin, line 58: What are these for?

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementParams.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public String getTitle() {
+        return title;
+    }
+
+    public void setTitle(String title) {
+        this.title = title;
+    }
+
+    public Object getValue() {
```
ZakarFin, line 43: Why object? Isn't this always a String?

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsDBHelper.java
```diff
@@ -0,0 +1,367 @@
...
+
+import java.sql.*;
+import java.util.List;
+import java.text.SimpleDateFormat;
+import java.time.*;
+
+/**
+ * Helper for seutumasia db handling.
+ */
+public class AnnouncementsDBHelper {
```
ZakarFin, line 22: This whole file needs to be written as a MybatisService.

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsParser.java
```diff
@@ -0,0 +1,69 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Search helper
+ */
+public class AnnouncementsParser {
```
ZakarFin, line 15: Use this in controller to parse request -> announcement. Not in the "db helper".

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsParser.java
```diff
@@ -0,0 +1,69 @@
...
+import fi.nls.oskari.control.ActionParamsException;
+import org.oskari.announcements.helpers.AnnouncementParams;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Search helper
```
ZakarFin, line 13: Search helper?

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsDBHelper.java
```diff
@@ -0,0 +1,367 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.sql.*;
+import java.util.List;
+import java.text.SimpleDateFormat;
+import java.time.*;
+
+/**
+ * Helper for seutumasia db handling.
```
ZakarFin, line 20: seutumasia?

ZakarFin:
What's the difference with getting announcements for admin or a regular user? If it's just that admin should get them all and regular users only the current ones, maybe combine it in one handler and just return a list based on the user role?

Also you should combine the save/update/delete/get to a single RestController instead of having a "Get" handler for each of them.

content-resources/src/main/java/flyway/announcements/V1_0_0__register_announcement_bundles.java
osqq: What file/folder exactly?

ZakarFin: The same you have for the db table: `flyway/oskari` -> `package flyway.oskari`. The flyway modules are grouped by the folder they are in so with this we would have an `announcements` module with a single migration registering the bundles. If we would want to separate announcements functionality to it's own module the table creation should be in this module as well. But I think it's easier for everyone if we just have both of them in the core-module that can't be skipped.

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementParams.java
```diff
@@ -0,0 +1,49 @@
+package org.oskari.announcements.helpers;
+
+public class AnnouncementParams {
+    private String id;
+    private String title;
+    private Object value;
```
ZakarFin, line 6: Why is value an Object?

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsDBHelper.java
```diff
@@ -0,0 +1,367 @@
...
+     */
+    public static JSONObject getAdminAnnouncements() throws JSONException, ActionParamsException {
+        Connection conn = null;
+
+        JSONArray results = new JSONArray();
+        String sql = "";
+        String sqlWithParams = "";
+
+        try  {
+            conn = getConnection();
```
ZakarFin, line 121: You can do try-with-resources and get rid of the finally-block:
```suggestion
        try (Connection conn = getConnection()) {
```

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsDBHelper.java
```diff
@@ -0,0 +1,367 @@
...
+                }
+            }
+
+            ResultSet rs = pstmt.executeQuery();
+            while (rs.next()) {
+                results.put(rs.getInt("id"));
+            }
+
+        } catch (SQLException e) {
+            LOG.error(e, "Cannot create SQL query, sql=" + sqlWithParams);
```
ZakarFin, line 217: Isn't `sqlWithParams` always empty?

ZakarFin: It would be a much more standard way of detecting errors if you throw an exception if there is one. Now the caller needs to check if it received an id or not from a JSON returned from this method.

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsDBHelper.java
```diff
@@ -0,0 +1,367 @@
...
+        
+            conn = getConnection();
+            StringBuilder sb = new StringBuilder();
+            sb.append("INSERT INTO oskari_announcements(title, content, begin_date, end_date, active) ");
+            sb.append("VALUES (? ::VARCHAR, ? ::VARCHAR, ? ::DATE, ? ::DATE, ? ::BOOLEAN) ");
+            sb.append("RETURNING id;");
+
+            
+
+            sql = sb.toString();
```
ZakarFin, line 188: I can't really see a reason why would you store the sql to this variable instead of calling `prepareStatement(sb.toString())`.

content-resources/src/main/resources/flyway/oskari/V2_0_8__register_announcement_bundles.java
```diff
@@ -0,0 +1,20 @@
+package flyway.oskari;
+
+import java.sql.Connection;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+
+import org.oskari.helpers.BundleHelper;
+import fi.nls.oskari.domain.map.view.Bundle;
+
+public class V1_0_0__register_announcement_bundles extends BaseJavaMigration {
```
ZakarFin, line 10: The class name should match the filename: 
```suggestion
public class V2_0_8__register_announcement_bundles extends BaseJavaMigration {
```

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    public void handleGet(ActionParameters params) throws ActionException {
+
+        try {
+            JSONObject result = new JSONObject();
+            if (params.getUser().isAdmin()) {
+                result = service.getAdminAnnouncements();
+            } else {
+                result = service.getAnnouncements();
+            }
+            System.out.println(result);
```
ZakarFin, line 33: Remove sout.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    private AnnouncementsService service = new AnnouncementsServiceMybatisImpl();
+
+    // Handle get announcements
+    @Override
+    public void handleGet(ActionParameters params) throws ActionException {
+
+        try {
+            JSONObject result = new JSONObject();
+            if (params.getUser().isAdmin()) {
+                result = service.getAdminAnnouncements();
```
ZakarFin, line 29: What is the difference with admin announcements and regular ones? Would you want the admin to see the regular ones but be able to load the admin ones for an admin UI? If so you could use the same getter but add a parameter to toggle which response you would like (and also check that only admin can use that toggle).

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            } else {
+                result = service.getAnnouncements();
+            }
+            System.out.println(result);
+            ResponseHelper.writeResponse(params, 200, result);
+        
+        } catch (Exception e) {
+            LOG.error("Error", e);
```
ZakarFin, line 37: On an exception the frontend doesn't get any response and the message "Error" isn't really helping anyone :) Maybe it should with something like ResponseHelper.writeError()?

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            .withParam("id", announcement.getId())
+            .withParam("title", announcement.getTitle())
+            .withParam("content", announcement.getContent())
+            .withParam("beginDate", announcement.getBeginDate())
+            .withParam("endDate", announcement.getEndDate())
+            .withParam("active", announcement.getActive());
+
+            ResponseHelper.writeResponse(params, result);
+        } catch (JSONException e) {
+            LOG.error("Error updating announcement", e);
```
ZakarFin, line 60: No need to LOG.error() here. Since you are throwing an ActionException it will be logged anyway by the controller. Logging it twice only adds spam to the log files.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
ZakarFin: Or throw an ActionException like you do on update.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            throw new ActionException("Cannot update announcement");
+        }
+    }
+    // Handle delete announcements
+    @Override
+    public void handleDelete(ActionParameters params) throws ActionException {
+        params.requireAdminUser();
+        try {
+            int id = -1;
+            JSONObject jsonParams =  params.getPayLoadJSON();
```
ZakarFin, line 70: If the only parameter is the id, do you need to send it in JSON? Maybe just use an url-parameter and get it with params.getRequiredParamInt("id") so you don't have to worry about if it's not sent or if its not a number etc. Also remove the param getting out of the try-block so you don't wrap the possibly thrown ActionParamsException to a new ActionException spamming the logs.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            if (jsonParams.has("id")) {
+                id = jsonParams.getInt("id");
+            }
+            int deleteId = service.deleteAnnouncement(id);
+            final JSONObject result = new JSONObject();
+            JSONHelper.putValue(result, "id", deleteId);
+            ResponseHelper.writeResponse(params, result);
+        } catch (Exception e) {
+            LOG.error("Error with deleting announcement", e);
```
ZakarFin, line 79: Remove logging when throwing a new exception

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            .withParam("id", announcement.getId())
+            .withParam("title", announcement.getTitle())
+            .withParam("content", announcement.getContent())
+            .withParam("beginDate", announcement.getBeginDate())
+            .withParam("endDate", announcement.getEndDate())
+            .withParam("active", announcement.getActive());
+
+            ResponseHelper.writeResponse(params, result);
+        } catch (Exception e) {
+            LOG.error("Error saving announcement", e);
```
ZakarFin, line 104: Remove logging when throwing a new exception

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            throw new ActionException("Cannot save announcement");
```
ZakarFin, line 105: Attach the thrown exception to the new one so we know why it failed.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            if (jsonParams.has("id")) {
+                id = jsonParams.getInt("id");
+            }
+            int deleteId = service.deleteAnnouncement(id);
+            final JSONObject result = new JSONObject();
+            JSONHelper.putValue(result, "id", deleteId);
+            ResponseHelper.writeResponse(params, result);
+        } catch (Exception e) {
+            LOG.error("Error with deleting announcement", e);
+            throw new ActionException("Cannot delete announcement");
```
ZakarFin, line 80: Attach the thrown exception to the new one so we know why it failed.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public JSONObject getAdminAnnouncements() {
+        LOG.debug("get admin announcements");
+        final SqlSession session = factory.openSession();
+        JSONObject data = new JSONObject();
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            data = parser.parseAnnouncementsMap(mapper.getAdminAnnouncements());
+            session.commit();
```
ZakarFin, line 59: You don't need to commit when only reading.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to get admin announcements", e);
+        } finally {
+            session.close();
+        }
```
ZakarFin, line 64: Use try-with-resources:
```suggestion
        try (final SqlSession session = factory.openSession()) {
            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
            return parser.parseAnnouncementsMap(mapper.getAdminAnnouncements());
        } catch (Exception e) {
            throw new ServiceRuntimeException("Failed to get admin announcements", e);
        }
```

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+        final SqlSession session = factory.openSession();
+        JSONObject data = new JSONObject();
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            data = parser.parseAnnouncementsMap(mapper.getAnnouncements());
+            session.commit();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to get announcements", e);
+        } finally {
+            session.close();
```
ZakarFin, line 79: Use try-with-resources so you don't need to manually close the session.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public JSONObject getAnnouncements() {
+        LOG.debug("get announcements");
+        final SqlSession session = factory.openSession();
+        JSONObject data = new JSONObject();
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            data = parser.parseAnnouncementsMap(mapper.getAnnouncements());
+            session.commit();
```
ZakarFin, line 75: No need to commit on read.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+        final SqlSession session = factory.openSession();
+        int updateId = -1;
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            updateId = mapper.updateAnnouncement(announcement);
+            session.commit();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to update announcement", e);
+        } finally {
+            session.close();
```
ZakarFin, line 95: Use try-with-resources so you don't need to manually close the session.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+        final SqlSession session = factory.openSession();
+        int saveId = -1;
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            saveId = mapper.saveAnnouncement(announcement);
+            session.commit();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to save announcement", e);
+        } finally {
+            session.close();
```
ZakarFin, line 111: Use try-with-resources so you don't need to manually close the session.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+        final SqlSession session = factory.openSession();
+        int deletedId = -1;
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            deletedId = mapper.deleteAnnouncement(id);
+            session.commit();
+        } catch (Exception e) {
+            LOG.error(e, "Couldn't delete announcement with id:", id);
+        } finally {
+            session.close();
```
ZakarFin, line 127: Use try-with-resources so you don't need to manually close the session.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    public int deleteAnnouncement(int id) {
+        LOG.debug("delete announcement with id: " + id);
+        final SqlSession session = factory.openSession();
+        int deletedId = -1;
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            deletedId = mapper.deleteAnnouncement(id);
+            session.commit();
+        } catch (Exception e) {
+            LOG.error(e, "Couldn't delete announcement with id:", id);
```
ZakarFin, line 125: Throw ServiceRuntimeException instead so we know something went wrong. Now it just fails silently.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    public int saveAnnouncement(final Announcement announcement) {
+        LOG.debug("save new announcement");
+        final SqlSession session = factory.openSession();
+        int saveId = -1;
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            saveId = mapper.saveAnnouncement(announcement);
+            session.commit();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to save announcement", e);
```
ZakarFin, line 109: Use ServiceRuntimeExpection instead.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    public int updateAnnouncement(final Announcement announcement) {
+        LOG.debug("update announcement");
+        final SqlSession session = factory.openSession();
+        int updateId = -1;
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            updateId = mapper.updateAnnouncement(announcement);
+            session.commit();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to update announcement", e);
```
ZakarFin, line 93: Use ServiceRuntimeExpection instead.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    public JSONObject getAnnouncements() {
+        LOG.debug("get announcements");
+        final SqlSession session = factory.openSession();
+        JSONObject data = new JSONObject();
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            data = parser.parseAnnouncementsMap(mapper.getAnnouncements());
+            session.commit();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to get announcements", e);
```
ZakarFin, line 77: Use ServiceRuntimeExpection instead.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    public JSONObject getAdminAnnouncements() {
+        LOG.debug("get admin announcements");
+        final SqlSession session = factory.openSession();
+        JSONObject data = new JSONObject();
+        try {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            data = parser.parseAnnouncementsMap(mapper.getAdminAnnouncements());
+            session.commit();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to get admin announcements", e);
```
ZakarFin, line 61: Use ServiceRuntimeExpection instead.

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsParser.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Parse announcement params from json
+     * @param params
+     * @return
+     * @throws JSONException
+     * @throws ActionParamsException
+     */
+    public static Announcement parseAnnouncementParams (ActionParameters params) throws JSONException, ActionParamsException {
+        JSONObject jsonParams =  params.getPayLoadJSON();
+        Announcement Announcement = new Announcement();
```
ZakarFin, line 30: It's newer a good idea to have a variable name the same as the class name. 
```suggestion
        Announcement announcement = new Announcement();
```

control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsParser.java
```diff
@@ -0,0 +1,85 @@
...
+
+        if (jsonParams.has(KEY_ID)) {
+            Announcement.setId(jsonParams.getInt(KEY_ID));
```
ZakarFin, line 33: This looks like you are calling static methods of a class. Use lowercase names for variables to separate them from class names.

content-resources/src/main/resources/flyway/oskari/V2_0_8__register_announcement_bundles.java
osqq: Had some changes that wasn't pushed, including this fix

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        } catch (Exception e) {
+            throw new ServiceRuntimeException("Failed to get announcements", e);
+        }
+    }
+
+    public int updateAnnouncement(final Announcement announcement) {
+        int updateId = -1;
+        try (final SqlSession session = factory.openSession()) {
+            final AnnouncementsMapper mapper = session.getMapper(AnnouncementsMapper.class);
+            updateId = mapper.updateAnnouncement(announcement);
```
ZakarFin, line 76: You'll probably need that commit() call back on these write ops?

control-announcements/src/main/java/org/oskari/announcements/helpers/Announcement.java
```diff
@@ -0,0 +1,69 @@
+package org.oskari.announcements.helpers;
+
+public class Announcement {
+    private Integer id;
+    private String title;
+    private String content;
+    private String begin_date;
+    private String end_date;
```
ZakarFin, line 8: With the Mybatis mapper taking dates as param shouldn't these be dates as well?

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsServiceMybatisImpl.java
osqq: Whoopsie, dropped it on accident, let's add those back

control-announcements/src/main/java/org/oskari/announcements/helpers/Announcement.java
osqq: Those are strings everywhere and then cast as date, unless I'm missing something (?)

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
ZakarFin: Answering myself, the difference is that admin gets all announcements and normal users just the current ones. So when logged in as admin the announcements don't work the same way as for other users. In my opinion this needs a parameter that is only allowed for admin users for getting all announcements.

ZakarFin:
I think the only changes this needs now is to have the get-method to handle a param for getting the full list (for the admin ui) and moving the delete id to a parameter instead of payload json. Then it's good for merging.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
osqq: Is there a possible way to achieve that without turning this into PUT or POST ?

ZakarFin: Yes, you can send URL-parameters in GET requests as well: https://en.wikipedia.org/wiki/Query_string

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,101 @@
...
+public class AnnouncementsHandler extends RestActionHandler {
+    private AnnouncementsService service = new AnnouncementsServiceMybatisImpl();
+
+    // Handle get announcements
+    @Override
+    public void handleGet(ActionParameters params) throws ActionException {
+
+        try {
+            JSONObject result = new JSONObject();
+            if (params.getUser().isAdmin()) {
```
ZakarFin, line 25:
```suggestion
            if (params.getHttpParam("all", false) && params.getUser().isAdmin()) {
```

Then on the frontend when you want to list all of the announcements for admin you can call `/action?action_route=Announcements&all=true`

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -0,0 +1,101 @@
...
+    // Handle delete announcements
+    @Override
+    public void handleDelete(ActionParameters params) throws ActionException {
+        params.requireAdminUser();
+        try {
+            int id = -1;
+            JSONObject jsonParams =  params.getPayLoadJSON();
+            if (jsonParams.has("id")) {
+                id = jsonParams.getInt("id");
+            }
```
ZakarFin, line 68:
```suggestion
            int id = params.getRequiredParamInt("id");
```
And on the frontend create url to call with: `Oskari.urls.getRoute('Announcements', { id: [id to remove]}`

ZakarFin:
Thanks!

PR #668, opened 2020-12-02
Add layer usage check

mhieta:
Add new action route to check if layer is used elsewhere.

Action route takes layer id as parameter and it can be called with:
`/action?id=[layer_id]&action_route=LayerAdminUsageCheck`

Response is in format of:
```
{
   "timeseries": [array of layer ids]
}
```

Currently only timeseries usage is checked.


control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminUsageCheckHandler.java
```diff
@@ -0,0 +1,72 @@
...
+    public void init() {
+        try {
+            mapLayerService = OskariComponentManager.getComponentOfType(OskariLayerService.class);
+        } catch (Exception e) {
+            throw new ServiceRuntimeException("Exception occured while initializing map layer service", e);
+        }
+    }
+
+    @Override
+    public void handleGet(ActionParameters params) throws ActionException {
```
ZakarFin, line 37: Start this off with `params.requireAdminUser()` as first line since this is for admins only. Then we don't really need to care if we expose ids that might require some permission.

control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminUsageCheckHandler.java
```diff
@@ -0,0 +1,72 @@
...
+                    if (metadataLayerId == layerId) {
+                        timeseriesLayerIds.add(layer.getId());
+                    }
+                }
+            } catch (JSONException e) {
+                throw new ActionException("Cannot parse layer metadata options for layer: " +
+                        layer.getName() + ", id: " + layer.getId());
+            }
+
+        }
```
ZakarFin, line 57: Move this to a method that takes that returns the Set.

control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminUsageCheckHandler.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private void writeResponse(ActionParameters params, Object output) {
+        params.getResponse().setCharacterEncoding("UTF-8");
+        params.getResponse().setContentType("application/json;charset=UTF-8");
+        try {
+            ResponseHelper.writeResponse(params, mapper.writeValueAsString(output));
+        } catch (JsonProcessingException e) {
+            ResponseHelper.writeError(params, "Couldn't serialize JSON");
+        }
```
ZakarFin, line 69: I think you can streamline this to:
```suggestion
        JSONObject response = new JSONObject();
        response.put("timeseries", timeseriesLayerIds);
        ResponseHelper.writeResponse(params, response);
```

control-admin/src/test/java/fi/nls/oskari/control/admin/LayerAdminUsageCheckHandlerTest.java
```diff
@@ -0,0 +1,83 @@
...
+        JSONObject metadata = new JSONObject().put("layer", metadataLayerId);
+        JSONObject timeseries = new JSONObject().put("metadata", metadata);
+        wmsLayer.setOptions(new JSONObject().put("timeseries", timeseries));
+    }
+
+    @Test
+    public void testLayerUsage() throws Exception {
+        final Map<String, String> parameters = new HashMap<>();
+        parameters.put("id", METADATA_LAYER_ID.toString());
+        final ActionParameters params = createActionParams(parameters);
```
ZakarFin, line 78: When you add the requirement for admin user you need to do this:
```suggestion
        final ActionParameters params = createActionParams(parameters, getAdminUser());
```

control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminUsageCheckHandler.java
```diff
@@ -0,0 +1,64 @@
...
+            throw new ServiceRuntimeException("Exception occured while initializing map layer service", e);
+        }
+    }
+
+    @Override
+    public void handleGet(ActionParameters params) throws ActionException {
+        params.requireAdminUser();
+        final int layerId = params.getRequiredParamInt(PARAM_LAYER_ID);
+        LOG.info("Checking layer usage in other layers for layerId: ", layerId);
+        Map<String, Set<Integer>> layerUsages = new HashMap<>();
```
ZakarFin, line 38: Change the type here to generic JSONObject since that's when ResponseHelper writes the json-headers: https://github.com/oskariorg/oskari-server/blob/master/service-control/src/main/java/fi/nls/oskari/util/ResponseHelper.java#L32-L35


```suggestion
        JSONObject response = new JSONObject();
```

control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminUsageCheckHandler.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    private Set<Integer> getTimeseriesLayerIds(int layerId) throws ActionException {
+        Set<Integer> timeseriesLayerIds = new HashSet<>();
+        for (OskariLayer layer : mapLayerService.findAll()) {
+            JSONObject options = layer.getOptions();
+            try {
+                if (options != null && options.has("timeseries")) {
+                    JSONObject timeseriesOptions = options.getJSONObject("timeseries");
+                    Integer metadataLayerId = timeseriesOptions.getJSONObject("metadata").getInt("layer");
```
ZakarFin, line 51: This could throw a NPE if options has timeseries but not metadata under it.

ZakarFin:
```suggestion
                    JSONObject timeseriesMetadata = timeseriesOptions.optJSONObject("metadata");
                    if (timeseriesMetadata == null) (
                        continue;
                    }
                    Integer metadataLayerId = timeseriesMetadata.getInt("layer");
```

PR #670, opened 2020-12-04
Add an action route that returns the fields information for WFS layers

mingfeng:
The result is structured as:

```
{
    "attributes": {
        "attr-1": "string",
        "attr-2": "number",
        ...
        "attr-n": "unknown",
    },
    "geometryField": "geometry",
    "locale": {
        "en": {
            "attr-1: "attr-1-en"
            ...
        },
        "fi": {
            "attr-1": "attr-1-fi"
            ...
        },
        ...
    }
}
```

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandler.java
@@ -1,0 +1,73 @@
+package fi.nls.oskari.control.layer;
+
+import fi.nls.oskari.annotation.OskariActionRoute;
+import fi.nls.oskari.control.ActionException;
+import fi.nls.oskari.control.ActionParameters;
+import fi.nls.oskari.control.ActionParamsException;
+import fi.nls.oskari.control.RestActionHandler;
+import fi.nls.oskari.domain.User;
+import fi.nls.oskari.domain.map.OskariLayer;
+import fi.nls.oskari.map.layer.OskariLayerService;
+import fi.nls.oskari.service.OskariComponentManager;
+import fi.nls.oskari.service.ServiceException;
+import fi.nls.oskari.service.ServiceRuntimeException;
+import fi.nls.oskari.util.ResponseHelper;
+import fi.nls.oskari.util.WFSGetLayerFields;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.oskari.permissions.PermissionService;
+
+/**
+ * An action route that returns fields information for WFS layers
+ */
+@OskariActionRoute("GetWFSLayerFields")
+public class GetWFSLayerFieldsHandler extends RestActionHandler {
+    private static final String PARAM_LAYER_ID = "layer_id";
+    private PermissionHelper permissionHelper;
+
+    @Override
+    public void init() {
+        try {
+            final OskariLayerService layerService = OskariComponentManager.getComponentOfType(OskariLayerService.class);
+            final PermissionService permissionService = OskariComponentManager.getComponentOfType(PermissionService.class);
+            permissionHelper = new PermissionHelper(layerService, permissionService);
+        } catch (Exception e) {
+            throw new ServiceRuntimeException("Exception occurred while initializing map layer service", e);
+        }
+    }
+
+    @Override
+    public void handleAction(ActionParameters params) throws ActionException {
+        final int layerId = params.getRequiredParamInt(PARAM_LAYER_ID);
+        final User user = params.getUser();
+        final OskariLayer layer = getLayer(layerId, user);
+        final JSONObject response = getLayerFields(layer);
+        ResponseHelper.writeResponse(params, response);
+    }
+
+    private JSONObject getLayerFields(OskariLayer layer) throws ActionException {
+        try {
+            JSONObject fields = WFSGetLayerFields.getLayerFields(layer);
+            JSONObject locale = getFieldsLocale(layer);
+            fields.putOpt("locale", locale);
+            return fields;
+        } catch (ServiceException ex) {
+            throw new ActionException("Error getting layer fields", ex);
+        } catch (JSONException ex) {
+            throw new ActionException("Invalid attribute locale", ex);
+        }
+    }
+
+    private JSONObject getFieldsLocale(OskariLayer layer) {
+        JSONObject data = layer.getAttributes().optJSONObject("data");
+        return data != null ? data.optJSONObject("locale") : null;
+    }
+
+    private OskariLayer getLayer(int layerId, User user) throws ActionException {
+        final OskariLayer layer = permissionHelper.getLayer(layerId, user);
+        if (!OskariLayer.TYPE_WFS.equals(layer.getType())) {
+            throw new ActionParamsException("Only WFS supported. Wrong layer type: " + layer.getType());
+        }
+        return layer;
+    }
+}
--- a/control-base/src/main/java/fi/nls/oskari/util/WFSGetLayerFields.java
+++ b/control-base/src/main/java/fi/nls/oskari/util/WFSGetLayerFields.java
@@ -1,0 +1,172 @@
+package fi.nls.oskari.util;
+
+import fi.nls.oskari.domain.map.OskariLayer;
+import fi.nls.oskari.service.ServiceException;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.oskari.service.wfs3.OskariWFS3Client;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.util.*;
+
+public class WFSGetLayerFields {
+    private static final String STRING = "string";
+    private static final String NUMBER = "number";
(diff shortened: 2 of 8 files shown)


control-base/src/main/java/fi/nls/oskari/util/WFSGetLayerFields.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * Return fields information for WFS 3.x layer
+     * @param layer
+     * @return
+     * @throws ServiceException
+     */
+    private static JSONObject getCollectionFields(OskariLayer layer) throws ServiceException {
+        final String url = layer.getUrl() + "collections/" + layer.getName() + "/queryables";
```
ZakarFin, line 48: The queryables endpoint is used for describing for example attributes that can be used to search items. It isn't a comprehensive list of attributes that the items in the collection have. The best we can do here instead is querying the /items endpoint for the first 10 or so features and gather the set of attribute names detected from those features. That should be reasonably accurate.

mingfeng: Makes sense

control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        try {
+            layerService = OskariComponentManager.getComponentOfType(OskariLayerService.class);
+        } catch (Exception e) {
+            throw new ServiceRuntimeException("Exception occurred while initializing map layer service", e);
+        }
+    }
+
+    @Override
+    public void handleAction(ActionParameters params) throws ActionException {
+        final int layerId = ConversionHelper.getInt(params.getRequiredParam(PARAM_LAYER_ID), -1);
```
ZakarFin, line 37:
```suggestion
        final int layerId = params.getRequiredParamInt(PARAM_LAYER_ID)
```

control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    private JSONObject getFieldsLocale(OskariLayer layer) {
+        JSONObject data = layer.getAttributes().optJSONObject("data");
+        return data != null ? data.optJSONObject("locale") : null;
+    }
+
+    private OskariLayer getLayer(int layerId) throws ActionException {
+        final OskariLayer layer = layerService.find(layerId);
```
ZakarFin, line 62: Run this through the PermissionHelper as in https://github.com/oskariorg/oskari-server/blob/master/control-base/src/main/java/fi/nls/oskari/control/layer/GetLayerTileHandler.java#L67

The helper will throw a ActionDeniedException if the user doesn't have permission to view the layer -> they shouldn't get to see the attributes/configurations we have on that layer either.

control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        if (layer == null) {
+            throw new ActionException("Layer not found: " + layerId);
+        }
+        return layer;
```
ZakarFin, line 66:
```suggestion
        if (!OskariLayer.TYPE_WFS.equals(layer.getType()) {
            throw new ActionParamsException("Wrong layer type");
        }
        return layer;
```

control-base/src/main/java/fi/nls/oskari/util/WFSGetLayerFields.java
```diff
@@ -0,0 +1,157 @@
...
+     * The fields are collected by requesting 10 features from the
+     * collection and derive the attribute type from the attribute
+     * value, which is either NUMBER or STRING. If the attribute
+     * type cannot be derived from any feature, e.g. all 10 features
+     * have null values on the attribute, then the attribute will
+     * have an "unknown" type
+     *
+     */
+    private static JSONObject getCollectionFields(OskariLayer layer) throws ServiceException {
+        String url = layer.getUrl() + "collections/" + layer.getName() + "/items";
```
ZakarFin, line 59: Use https://github.com/oskariorg/oskari-server/blob/master/service-wfs3/src/main/java/org/oskari/service/wfs3/OskariWFS3Client.java#L229 (make the static method public on the client) so we see where it's used. I don't think we should use the existing client for getting the features since it's probably doing too much for what we need here. We could use `OskariWFS3Client.validateResponse(conn, CONTENT_TYPE_GEOJSON);` though for sanity checking the response.

control-base/src/main/java/fi/nls/oskari/util/WFSGetLayerFields.java
```diff
@@ -0,0 +1,157 @@
...
+        final Set<String> booleanTypes = new HashSet<>(Arrays.asList("boolean"));
+        try {
+            final JSONObject propertyTypes = response.getJSONObject("propertyTypes");
+            final Iterator<?> attributeNames = propertyTypes.keys();
+            final JSONObject attributes = new JSONObject();
+            final JSONObject result = new JSONObject();
+            while (attributeNames.hasNext()) {
+                String attributeName = (String) attributeNames.next();
+                String attributeType = propertyTypes.getString(attributeName);
+                attributeType = attributeType.split(":")[1];  // remove xml namespace prefix
```
ZakarFin, line 138: We should make sure there is a ':' character included and not assume there is. Most probably have it but there's always the chance that XML default namespace is set based on the featuretype.

control-base/src/test/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandlerTest.java
```diff
@@ -0,0 +1,92 @@
...
+        OskariLayerService mockLayerService = getMockLayerService();
+        PowerMockito.mockStatic(OskariComponentManager.class);
+        when(OskariComponentManager.getComponentOfType(OskariLayerService.class)).thenReturn(mockLayerService);
+    }
+
+    private static OskariLayerService getMockLayerService() throws JSONException {
+        OskariLayerService mockLayerService = mock(OskariLayerServiceMybatisImpl.class);
+        OskariLayer wfsLayer = new OskariLayer();
+        wfsLayer.setType("WFS");
+        final String rawJsonStr = "{\"data\": {\"locale\": {\"fi\": {\"field-en\": \"field-fi\"}}}}";
```
ZakarFin, line 63: "field-en" might be a bit misleading here. Rename to something like `"test-attribute": "label for fi"` or similar.

control-base/src/test/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandlerTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private static JSONObject getMockFields() throws JSONException {
+        final String rawFieldsStr = "{\"attributes\": {\"attr-name\": \"string\"}, \"geometryField\": \"geometry\"}";
+        return new JSONObject(rawFieldsStr);
+    }
+
+    private static void mockOskariComponentManager() throws JSONException {
+        OskariLayerService mockLayerService = getMockLayerService();
+        PowerMockito.mockStatic(OskariComponentManager.class);
+        when(OskariComponentManager.getComponentOfType(OskariLayerService.class)).thenReturn(mockLayerService);
```
ZakarFin, line 56: I think we've mostly used setters on handlers for setting up mock services with init() checking if the services have been set before overwriting them with values from the component manager. But this is ok since it looks like it's not leaking the mocks between tests.

control-base/src/test/java/fi/nls/oskari/util/WFSGetLayerFieldsTest.java
```diff
@@ -0,0 +1,91 @@
...
+        propertyTypes.put("attr-5", "prefix:complex");
+        propertyTypes.put("geom", "gml:GeometryPropertyType");
+        final JSONObject propertyTypesResponse = new JSONObject();
+        propertyTypesResponse.put("propertyTypes", propertyTypes);
+        when(WFSDescribeFeatureHelper.getWFSFeaturePropertyTypes(layer, String.valueOf(layer.getId()))).thenReturn(propertyTypesResponse);
+    }
+
+    private OskariLayer getWFSLayer(String version) {
+        OskariLayer layer = new OskariLayer();
+        layer.setType("WFS");
```
ZakarFin, line 82:
```suggestion
        layer.setType(OskariLayer.TYPE_WFS);
```

ZakarFin:
For anyone following Oskari development/PRs: future plans include using this as a replacement and unifying at least these two:
- https://github.com/oskariorg/oskari-server/blob/master/control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerGeometryTypeHandler.java
- https://github.com/oskariorg/oskari-server/blob/master/control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSDescribeFeatureHandler.java

PR #671, opened 2020-12-08
Cleanup timeseries metadata after layer removal

mhieta:
Cleanup referenced layers metadata after layer removal.
Refactor layer metadata handling to be done in one place with Helper-class.
Create own test class for Layer Admin tests.
Refactor existing Layer Admin Usage tests to use a common base class with Layer Admin tests.

### DIFF
--- a/control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminHandler.java
+++ b/control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminHandler.java
@@ -27,7 +27,6 @@ import org.oskari.maplayer.admin.LayerValidator;
 import org.oskari.maplayer.model.MapLayer;
 import org.oskari.maplayer.model.MapLayerAdminOutput;
 import org.oskari.log.AuditLog;
-import org.oskari.service.util.ServiceFactory;
 
 import java.util.*;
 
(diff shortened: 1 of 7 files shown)


control-admin/src/test/java/fi/nls/oskari/control/admin/LayerAdminUsageCheckHandlerTest.java
```diff
@@ -88,13 +100,21 @@ public void testLayerUsageCheckWhenUserIsNotAdmin() {
...
-            handler.handleAction(params);
+            adminUsageCheckHandler.handleAction(params);
             fail("ActionDeniedException should have been thrown");
         } catch (ActionException e) {
             assertEquals("Admin only", e.getMessage());
         }
     }
 
+    @Test
+    public void testCleanupLayerReferences() throws Exception {
```
ZakarFin, line 111: I think this should be on it's own test file but it would require similar mocking for both tests. Maybe sharing a helper class so that mocks only need to be written one time?

control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminHandler.java
```diff
@@ -123,6 +140,32 @@ public void handleDelete(ActionParameters params) throws ActionException {
         }
     }
 
+    public List<OskariLayer> cleanupLayerReferences(int layerId) throws ActionException {
+        List<OskariLayer> layers = mapLayerService.findAll();
```
ZakarFin, line 144: There's duplicated code here from the usage check handler. I'm wondering if we should have a helper that could be used in both. Just report report usage on the other handler and LayerAdminHandler could use that same helper to get a list of layers with reference the input layer id. Thoughts on this?

ZakarFin: I mean it's not bad but with nothing to tie these two together it's pretty easy to miss updating the logic in one when updating it on the other one.

control-admin/src/test/java/fi/nls/oskari/control/admin/LayerAdminUsageCheckHandlerTest.java
mhieta: Yeah, a helper class or a shared new common superclass could do the trick here.

control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminHandler.java
mhieta: Agreed. It's better to do that layer looping only once in one place. I'll check this.

PR #676, opened 2021-01-28
Update GeoTools (requires JTS upgrade as well)

ZakarFin:
Mainly to resolve issues with Shapefile parsing:
- https://osgeo-org.atlassian.net/browse/GEOT-6729
- https://osgeo-org.atlassian.net/browse/GEOT-6740

GeoTools 23.2 -> 23.4
JTS 1.16.1 -> 1.17.0


ZakarFin:
Jumping straight to 24.2 as looks like it's not causing any issues.

PR #680, opened 2021-01-28
Remove unused dependency: jsonld

ZakarFin:
Doesn't appear to be used. Some more testing required.


ZakarFin:
This was probably for transport webapp and serializing feature geometries.

PR #684, opened 2021-02-02
Layerlist hierarchical

osqq:
Set layerlist hierarchy depth as optional param with default value of 2

Set value in oskari-ext.properties with "layerlist.hierarchy.maxdepth"


ZakarFin:
Thanks!

PR #690, opened 2021-03-03
GFIType default value

okauppinen:
Use value from capabilities if gfi type isn't set for layer. Default to first SUPPORTED_GET_FEATURE_INFO_FORMATS.

### DIFF
--- a/service-map/src/main/java/fi/nls/oskari/map/data/domain/GFIRequestParams.java
+++ b/service-map/src/main/java/fi/nls/oskari/map/data/domain/GFIRequestParams.java
@@ -1,13 +1,19 @@
 package fi.nls.oskari.map.data.domain;
 
 import fi.nls.oskari.domain.map.OskariLayer;
+import fi.nls.oskari.util.JSONHelper;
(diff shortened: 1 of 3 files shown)


service-map/src/main/java/fi/nls/oskari/map/data/domain/GFIRequestParams.java
```diff
@@ -158,5 +159,26 @@ private String getBaseQueryURL() {
...
+        if (!value.isEmpty()) {
+            return value;
+        }
+        List <String> available = JSONHelper.getArrayAsList(JSONHelper.getJSONArray(formats, KEY_AVAILABLE));
+        for(String format : SUPPORTED_GET_FEATURE_INFO_FORMATS) {
+            if (available.contains(format)) {
+                return format;
+            }
+        }
+        return DEFAULT_FORMAT;
```
ZakarFin, line 181: I'm wondering if we might as well cancel the request if any of the available are not supported or should we fallback to using the first from available and hope for the best?

PR #691, opened 2021-03-10
Wfs attributes data

okauppinen:
Add filter and geometryType to GetWFSLayerFields response. I don't know if its better to return [names] instead of {default: [names]} if filter is array.

Move type checks and conversions to WFSConversionHelper to use them also in userlayer and analysis json formatter. 

Parse fields to attributes.data for userlayers and analysis. Add test for userlayer.

WFSLayerAttributes.getSelectedAttributes() find selected attribute with default key if language specific isn't available. I don't know if default lang should be last fallback like lang, default, defaultLang,[]

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandler.java
@@ -11,8 +11,10 @@ import fi.nls.oskari.map.layer.OskariLayerService;
 import fi.nls.oskari.service.OskariComponentManager;
 import fi.nls.oskari.service.ServiceException;
 import fi.nls.oskari.service.ServiceRuntimeException;
+import fi.nls.oskari.util.JSONHelper;
 import fi.nls.oskari.util.ResponseHelper;
 import fi.nls.oskari.util.WFSGetLayerFields;
+import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.oskari.permissions.PermissionService;
@@ -50,6 +52,7 @@ public class GetWFSLayerFieldsHandler extends RestActionHandler {
             JSONObject fields = WFSGetLayerFields.getLayerFields(layer);
             JSONObject locale = getFieldsLocale(layer);
             fields.putOpt("locale", locale);
+            fields.putOpt("filter", getFieldsFilter(layer));
             return fields;
         } catch (ServiceException ex) {
             throw new ActionException("Error getting layer fields", ex);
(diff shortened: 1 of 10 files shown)


control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandler.java
```diff
@@ -50,6 +52,7 @@ private JSONObject getLayerFields(OskariLayer layer) throws ActionException {
             JSONObject fields = WFSGetLayerFields.getLayerFields(layer);
             JSONObject locale = getFieldsLocale(layer);
             fields.putOpt("locale", locale);
+            fields.putOpt("filter", getFieldsFilter(layer));
```
ZakarFin, line 55: Should we rename this to something like `visibleAttributes` now that we also have filter in another meaning?

service-base/src/main/java/fi/nls/oskari/domain/map/wfs/WFSLayerAttributes.java
```diff
@@ -123,7 +123,7 @@ public boolean hasFilter() {
 
     public List<String> getSelectedAttributes(String lang) {
         return params.getOrDefault(lang,
-                params.getOrDefault(PropertyUtil.getDefaultLanguage(), Collections.emptyList()));
+                params.getOrDefault("default", Collections.emptyList()));
```
ZakarFin, line 126: For using this we wouldn't need to do anything for the frontend if we just use the instances default language (Could use Oskari.getLocalized() in the frontend). If we change this we will need to add handling for the "default".

service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatterUSERLAYER.java
```diff
@@ -34,49 +27,50 @@ public JSONObject getJSON(final OskariLayer baseLayer, UserLayer ulayer, String
...
+                JSONHelper.putValue(locale.get(lang), name, locales.get(lang));
+            });
+            // attribute types, like WFSGetLayerFields
+            JSONHelper.putValue(types, name, WFSConversionHelper.getSimpleType(type.toLowerCase()));
         }
+
+        JSONHelper.putValue(attributes, "data", data);
+        JSONHelper.putValue(data, "filter", filter);
+        JSONHelper.putValue(data, "types", types);
+        JSONHelper.put(filter, "default", filteredFields);
```
ZakarFin, line 70: This could be `lang` instead of `default` since it's always the same we can just use the current users language (or the instances default lang is fine too).

control-base/src/main/java/fi/nls/oskari/control/layer/GetWFSLayerFieldsHandler.java
okauppinen: Renamed `filter` to `selection`. Don't know if it's better but I think that it's more like featuredata bundle to decide that properties which aren't selected is hidden for default. User then may click all visible and array isn't updated. Also it's used for ordering. Also renamed `attributes` to `types`. Maybe action route could be renamed as `GetWFSPropertiesInfo`, `GetWFSMetadata`, `GetWFSLayerInfo` etc.

service-base/src/main/java/fi/nls/oskari/domain/map/wfs/WFSLayerAttributes.java
okauppinen: Reverted

service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatterUSERLAYER.java
okauppinen: Removed `default` and changed `filter` to be array

PR #692, opened 2021-03-11
Add support for importing GeoPackages

jampukka:
Here's my suggestion to https://github.com/oskariorg/oskari-docs/issues/246


service-userlayer/src/main/java/org/oskari/map/userlayer/input/GPKGParser.java
```diff
@@ -0,0 +1,66 @@
...
+                        .getCoordinateReferenceSystem();
+                if (crs != null) {
+                    sourceCRS = crs;
+                }
+                collection = FeatureCollectionParsers.read(source, sourceCRS, targetCRS);
+                if (!collection.isEmpty()) {
+                    break;
+                }
+            }
+            return collection;
```
jampukka, line 52: Throw an exception with approriate error message if this is null instead of returning null causing a weird NPE in `CreateUserLayerHandler` - probably an `UserLayerException`

ZakarFin:
Nice, thanks!

PR #696, opened 2021-03-22
Add handler for timeseries parameter

ZakarFin:
If timeseries URL-parameter has a value that can be split to two parts with `/` it will inject a `time` key to state with an array value. For other values it will just pass the parameter value as-is to frontend as `timeseries` bundle state.


mingfeng (approved):
LGTM :shipit:

PR #697, opened 2021-03-23
MyPlaces export features as geojson

okauppinen:
Add new action route for exporting myplaces features as geojson.

Not sure if we have to use Jackson ObjectMapper instead of JSONObject.toString() and BufferedWriter. Also don't know what defaults for prettify or indent should be.

### DIFF
--- a/control-myplaces/src/main/java/fi/nls/oskari/control/myplaces/handler/ExportMyPlacesFeaturesHandler.java
+++ b/control-myplaces/src/main/java/fi/nls/oskari/control/myplaces/handler/ExportMyPlacesFeaturesHandler.java
@@ -1,17 +1,11 @@
 package fi.nls.oskari.control.myplaces.handler;
 
 import fi.nls.oskari.annotation.OskariActionRoute;
-import fi.nls.oskari.control.ActionDeniedException;
 import fi.nls.oskari.control.ActionException;
 import fi.nls.oskari.control.ActionParameters;
-import fi.nls.oskari.control.RestActionHandler;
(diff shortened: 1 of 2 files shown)


control-myplaces/src/main/java/fi/nls/oskari/control/myplaces/handler/ExportMyPlacesFeaturesHandler.java
```diff
@@ -0,0 +1,64 @@
...
+    private static final DateTimeFormatter TIME_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+
+
+    @Override
+    public void handleGet(ActionParameters params) throws ActionException {
+        final User user = params.getUser();
+        final String srs = params.getHttpParam(PARAM_SRS, PropertyUtil.get("oskari.native.srs", "EPSG:4326"));
+        final String layerId = params.getRequiredParam(PARAM_LAYER_ID);
+        final int indent = params.getHttpParam(PARAM_INDENT, DEFAULT_INDENT);
+        final boolean prettify = params.getHttpParam(PARAM_PRETTIFY, true);
```
ZakarFin, line 38: I think we don't need a default indent value as if we remove it we don't need prettify. If the user gives an indent value between 1-8(?) we use that for indentation. I think we should make some sanity check for the value but if we do get one we can consider prettify == true.

ZakarFin:
Lets remove the prettify param and just use indent to detect if we need to format the response. Other than that looks good.

PR #699, opened 2021-03-23
Minor changes on TimeseriesParamHandler

mingfeng:
- Change the object key to "value" to match the one used in frontend
- Convert the values to integers


mingfeng:
We want to use the same TimeseriesParamHandler for player ui also, in which case a timestamp value is used

PR #703, opened 2021-03-29
Process additional "params" parameter in GFI requests

mingfeng:
params parameter is a json object includes additional params for each layer
with layer ids as object keys

### DIFF
--- a/service-map/src/main/java/fi/nls/oskari/map/data/domain/GFIRequestParams.java
+++ b/service-map/src/main/java/fi/nls/oskari/map/data/domain/GFIRequestParams.java
@@ -1,13 +1,16 @@
 package fi.nls.oskari.map.data.domain;
 
 import fi.nls.oskari.domain.map.OskariLayer;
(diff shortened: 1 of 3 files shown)


service-map/src/main/java/fi/nls/oskari/map/data/domain/GFIRequestParams.java
```diff
@@ -140,11 +146,21 @@ private String getAsQueryString() {
...
                 + getCurrentStyle() + "&LAYERS=" + wmsName;
+
+        if (additionalParams != null) {
+            Iterator<String> keys = additionalParams.keys();
+            while (keys.hasNext()) {
+                String key = keys.next();
+                String value = additionalParams.optString(key);
+                query += "&" + key.toUpperCase() + "=" + value;
+            }
+        }
```
ZakarFin, line 162: Use a method in IOHelper instead that will take care of urlencoding the param values etc.

PR #704, opened 2021-04-01
Add max results workaround to SearchService

kasparia:


service-search/src/main/java/fi/mml/portti/service/search/SearchService.java
```diff
@@ -36,9 +35,12 @@ public void init() {
...
 	 * Returns a generic maximum results instruction for search functions.
 	 * SearchChannels/implementations may opt to use this to
-	 * @return
+	 * @return maxCount  - Search results max count
      */
 	public int getMaxResultsCount() {
+		if (maxCount == -1) {
+			maxCount = ConversionHelper.getInt(PropertyUtil.getOptional("search.max.results"), maxCount);
+		}
 		return maxCount;
```
ZakarFin, line 44: Having maxCount at -1 when this is not configured can be problematic. Return the previous default when not configured with one of these. Pref the first if it works as intended:
```suggestion
		if (maxCount == -1) {
			maxCount = ConversionHelper.getInt(PropertyUtil.getOptional("search.max.results"), 100);
		}
		return maxCount;
```
If there's a timing issue with the first one we could also do this:
```suggestion
		if (maxCount == -1) {
			maxCount = ConversionHelper.getInt(PropertyUtil.getOptional("search.max.results"), -1);
		}
		if (maxCount == -1) {
		    // default to 100 when not configured to workaround timing issues
 		    return 100;
 		}
		return maxCount;
```

ZakarFin:
This is a workaround for timing issue where properties are not loaded before the SearchService.init() is called resulting in maxCount initializing to 100 instead of the value configured in oskari-ext.properties file.

service-search/src/main/java/fi/mml/portti/service/search/SearchService.java
kasparia: First one seemed to work.

PR #710, opened 2021-04-16
WFSLayer WPS params migration to JSONObject

okauppinen:
WFS layers' WPS params json is stored as String in db (inside attributes column). Frontend getWpsLayerParams gets WPS params from attributes object instead of wps_params (created JSONObject from String to own key to layerjson) which causes problems if attributes contains String wps params.

To fix the issue and clarify code:
migrate attributes.wpsParams to under attributes.data so it's not "analyze-specific"
Update WFSLayerAttributes to handle commonId and noDataValue
wpsInputType isn't added to WFSLayerAttributes because it's analysis specific but changed analysis to use WFSLayerAttributes.getData to link code together
Remove unused wps_params from wfslayer json formatter

Didn't update wpsInputType parsing/handling in AnalysisParser. Only updated to get value from layer.attributes.data.wpsInputType and added comment how it could be handled.
Method -> wpsReferenceType
AnalysisLayer -> wpsInputType
possible values: wfs, gs_vector, geojson
User own data layers' uses gs_vector and temp layers uses geojson.

### DIFF
--- a/content-resources/src/main/java/flyway/oskari/V2_3_1__migrate_wps_params.java
+++ b/content-resources/src/main/java/flyway/oskari/V2_3_1__migrate_wps_params.java
@@ -12,6 +12,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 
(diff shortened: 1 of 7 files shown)


ZakarFin:
After an internal meeting -> let's migrate stuff currently in attributes.wpsParams to under attributes.data so it's not "analyze-specific".

PR #713, opened 2021-04-23
Remove unnecessary layers

kuosman:
Remove unnecessary layers (my_places_categories and user_layer_data_style) for geoserver setup.


ZakarFin:
We could remove pushing the SLDs as well and remove them from code.

PR #715, opened 2021-04-29
GeoTools 25.1

okauppinen:
Update GeoTools to 25.x  when https://osgeo-org.atlassian.net/browse/GEOT-6850 is fixed

FIxed in: 25.1


ZakarFin:
Related links:
- #714 
- geotools/geotools#3256

ZakarFin:
Fixed in geotools/geotools#3482 (might get a backport to 25.x as well)

PR #724, opened 2021-05-14
Allow OAPIF requests for properties (without geometries)

ZakarFin:
Configuring a layer in oskari_maplayer.params with `{"properties": "id"}` will now include "properties=id" to requests for OGC API features requests. This will return the requested feature property without geometry. This can be used to optimize usage of these layers for example on scattered timeseries metadata.

Base branch is master since we might want to make 2.3.1 "hotfix" for enabling this.


jampukka:
Note that https://github.com/opengeospatial/ogcapi-features/tree/master/proposals/property-selection is still in proposal stage and therefore rather far from actually getting standardized. Consider adhering to the spec when it gets finalized in the future. However, as this is completely opt-in, admin-only functionality I don't see any harm in having it.

PR #731, opened 2021-06-02
Cache layerlist results

ZakarFin:
If the whole layerlist is requested (most cases in geoportal) -> cache the result by lang/projection/user roles. 

On most instances the result will be the same for guest users and another set for logged in "normal users" but without caching the result is always calculated from the ground up which is expensive due to the amount of processing required especially when there are a lot of layers in the instance.

Draft towards master atm as we might want to patch this in for 2.3.x still.

Caching in GetMapLayerGroupsHandler.java. Other changes are related to flushing the cached results when admin does changes to the db.


ZakarFin:
Merged to develop

PR #732, opened 2021-06-02
Clean up GFI error logging

ZakarFin:
Reduce logging where it's pretty common but doesn't really give any insight towards debugging. Stack trace for error is no longer logged normally but is available for logging on debug level.

Draft towards master atm as we might want to patch this in for 2.3.x still.


ZakarFin:
Merged to develop

PR #735, opened 2021-06-09
MIF/MID: handle projection type codes over 2000

jampukka:
Projection codes in MapInfo software may be modified by addition of 1000 or 2000 (or both) to indicate that the used system has affine transformation parameters or explicit bounds (respectively) defined. Also not all projection codes require the same number of parameters so move logic parsing double from MIFHeader to MIFProjection.

This fixes codes over 2000 but doesn't really address Affine Transforms, which we probably have to fix sometime in the future.


ZakarFin:
Added the missing imports but the test still fails for this.

ZakarFin:
Awesome, thanks!

PR #736, opened 2021-06-10
Remove layer json parts that are no longer used

ZakarFin:
- Removes "admin-block" from layers.
- Uses booleans for permissions instead of string values


ZakarFin:
Related frontend change: oskariorg/oskari-frontend#1599

PR #744, opened 2021-06-24
Add GitHub action for running mvn test (verify)

ZakarFin:


ZakarFin:
Java 11 fails with things like: 
```
java.lang.IllegalAccessError: class javax.naming.spi.NamingManager (in unnamed module @0x700f356b) cannot access class jdk.internal.loader.ClassLoaderValue (in module java.base) because module java.base does not export jdk.internal.loader to unnamed module @0x700f356b
```

Need to check these before merging. Wasn't a problem in Travis based builds but Travis used an old "update version" of 11.

PR #770, opened 2021-12-10
WMS GFI styles now come in additional params

kuosman:
This PR cleans server side from issue https://github.com/oskariorg/oskari-docs/issues/261


ZakarFin:
Thanks! Previous implementation assumed that just having a style in one parameter was enough since it works when requesting for single layer. In oskariorg/oskari-frontend#1709 the frontend was changed to pass styles per layer in `additionalParams` making it work when there's multiple layers on the map. This cleans up server logic so style is no longer read from that single param.

PR #780, opened 2021-12-23
MVT encoder changes

ZakarFin:
Currently there's an issue with the MVT encoder we use with the JTS dependency version mismatch between GeoTools and the MVT encoding library com.wdtinc.mapbox-vector-tile. This results in exceptions being thrown for line/polygon features when encoding MVTs as the JTS API has slightly changed:
```
Caused by: java.lang.NoSuchMethodError: 'org.locationtech.jts.geom.LineString org.locationtech.jts.geom.Polygon.getExteriorRing()'
	at com.wdtinc.mapbox_vector_tile.adapt.jts.JtsAdapter.toFeature(JtsAdapter.java:357)
	at com.wdtinc.mapbox_vector_tile.adapt.jts.JtsAdapter.toFeatures(JtsAdapter.java:300)
	at org.oskari.service.mvt.SimpleFeaturesMVTEncoder.encode(SimpleFeaturesMVTEncoder.java:54)
	at 
```

GeoServer (that also uses GeoTools) uses no.ecc.vectortile.java-vector-tile for encoding MVTs so using that instead would remove the JTS version mismatch problem. However there is a concern that the performance is not as good (some say the wdtinc is 10x faster) so we need to do some testing for this. Another option would be to fork the wdtinc library, update the JTS version/code to match like here (https://github.com/jampukka/mapbox-vector-tile-java) and host a version of it on the Maven repo on oskari.org so the updated version could be used in Oskari.

Anyways, this needs more testing including performance testing before merging.


ZakarFin:
Perfomance seems to be comparable to wdtinc (though did only light testing). The file sizes of tiles seem to be bigger for polygons and lines with no.ecc encoder, but the reason is unclear at this point. However generating MVTs in Oskari is not a critical feature and this fixes the JTS version mismatch problem by using the same library as GeoServer. So GeoTools/JTS version are now in line with the MVT encoder.

PR #789, opened 2022-01-13
Allow environment variables to take precedence over properties files

jampukka:
Here's my rather simple solution to https://github.com/oskariorg/oskari-docs/issues/262


ZakarFin:
Great, thanks!

PR #701, opened 2021-03-26
UserDataLayer localized names

okauppinen:
Add locale column for analysis, categories and user_layer tables. Populate locale object with name and available languages. 

Add all locale names for layerjson to allow user to edit them. Frontend doesn't allow to edit localized name for now but layer.getName can handle localized object.

Add LayerJSONFormatterANALYSIS to parse analysis like other UserDataLayers.

PR #794, opened 2022-01-20
Userlayer locale name, desc and source

okauppinen:
Migrage desc and source to locale with default language (name was migrated in https://github.com/oskariorg/oskari-server/pull/701).

Use locale to store values and return locale in maplayer json. Name is also returned as common to userdatalayer. Maybe return locale could be moved to common userdatalayer (and name removed) and override layer.getname in frontend.

Changed UserDataLayer to extend JSONLocalizedName because subtitle isn't used.

### DIFF
--- a/control-userlayer/src/main/java/org/oskari/control/userlayer/EditUserLayerHandler.java
+++ b/control-userlayer/src/main/java/org/oskari/control/userlayer/EditUserLayerHandler.java
@@ -21,9 +21,7 @@ import org.oskari.map.userlayer.service.UserLayerException;
 @OskariActionRoute("EditUserLayer")
 public class EditUserLayerHandler extends RestActionHandler {
 
-    private static final String PARAM_DESC = "desc";
-    private static final String PARAM_NAME = "name";
-    private static final String PARAM_SOURCE = "source";
+    private static final String PARAM_LOCALE = "locale";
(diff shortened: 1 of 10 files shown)


control-userlayer/src/main/java/org/oskari/control/userlayer/EditUserLayerHandler.java
```diff
@@ -37,11 +35,9 @@ public void init() {
     public void handlePost(ActionParameters params) throws ActionException {
         String mapSrs = params.getHttpParam(ActionConstants.PARAM_SRS);
         final UserLayer userLayer = UserLayerHandlerHelper.getUserLayer(userLayerDbService, params);
-        userLayer.setName(PropertyUtil.getDefaultLanguage(), params.getRequiredParam(PARAM_NAME)); // FIXME: userLayer.setLocale(locale);
-        userLayer.setLayer_desc(params.getHttpParam(PARAM_DESC, userLayer.getLayer_desc()));
-        userLayer.setLayer_source(params.getHttpParam(PARAM_SOURCE, userLayer.getLayer_source()));
+        userLayer.setLocale(params.getHttpParamAsJSON(PARAM_LOCALE));
```
ZakarFin, line 38: Not that getRequiredParam can't be used per field we should check that there's a name for the layer inside locale.

PR #811, opened 2022-02-17
Send headers before getting response code

ZakarFin:
So we don't get IllegalStateException with "Already connected".

### DIFF
--- a/service-base/src/main/java/fi/nls/oskari/util/IOHelper.java
+++ b/service-base/src/main/java/fi/nls/oskari/util/IOHelper.java
@@ -419,6 +419,7 @@ public class IOHelper {
 
             log.info("Following redirect to", location);
             HttpURLConnection newConnection = getConnection(location, user, pass, query, headers);
+            IOHelper.addIdentifierHeaders(newConnection);
             return followRedirect(newConnection, user, pass, query, headers, redirectLatch);
         } else {
             return conn;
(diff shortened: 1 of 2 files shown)


service-base/src/main/java/fi/nls/oskari/util/IOHelper.java
```diff
@@ -678,8 +679,12 @@ public static void setContentType(final HttpURLConnection con,
      * @throws IOException
      */
     public static void addIdentifierHeaders(final HttpURLConnection con) throws IOException {
-        con.setRequestProperty(HEADER_USERAGENT, getUserAgent());
-        con.setRequestProperty(HEADER_REFERER, getMyDomain());
+        try {
+            con.setRequestProperty(HEADER_USERAGENT, getUserAgent());
+            con.setRequestProperty(HEADER_REFERER, getMyDomain());
+        } catch (IllegalStateException e) {
+            log.warn("Couln't write ident headers:", e.getMessage());
```
jampukka, line 686: Couldn't

ZakarFin: Thanks :D

ZakarFin:
```suggestion
            log.warn("Couldn't write ident headers:", e.getMessage());
```

PR #813, opened 2022-02-22
WCS client TIFFReader: add short[] variants of readStrip and readTile

jampukka:
Add short[] variants of `readStrip` and `readTile` methods to TIFFReader.
Add simple tests for TIFFReader checking we can correctly read tiled TIFFs of type `Int16`, `UInt16`, `Float32`.

Instead of returning void return the `dst` array to allow for nicer api, allowing for example:
`float[] tile = myMap.computeIfAbsent(tileIndex, __ -> tiffreader.read(0, tileIndex, new float[256 * 256]));`
instead of 
`float[] tile = myMap.computeIfAbsent(tileIndex, __ -> { float[] tmp = new float[256 * 256]; tiffreader.read(0, tileIndex, tmp); return tmp; }));`


ZakarFin:
Nice, thanks!

PR #814, opened 2022-02-24
Add optional download basket email authentication

WTurunen:
Your SMTP host might need authentication. This commit checks properties
for user/password pair. If it finds them, it uses them for email
authentication.


ZakarFin:
Nice, thanks!

PR #816, opened 2022-03-02
Add OAPIF Core filtering capabilities to OskariWFS3Client

jampukka:
Work in progress


jampukka:
Add support for limited filtering capabilities to OGC API Features layers. This is currently limited to feature properties as described in [OGC API Features - Part 1 Core](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_parameters_for_filtering_on_feature_properties)

service-wfs3/src/main/java/org/oskari/service/wfs3/FilterToOAPIFCoreQuery.java
```diff
@@ -0,0 +1,443 @@
...
+
+        SimpleAndFlatteningFilterVisitor flattener = new SimpleAndFlatteningFilterVisitor(CommonFactoryFinder.getFilterFactory());
+        filter = (Filter) filter.accept(flattener, null);
+
+        OAPIFCoreFilterSplittingVisitor splitter = new OAPIFCoreFilterSplittingVisitor(capabilities, queryables);
+        filter.accept(splitter, null);
+
+        Filter preFilter = splitter.getFilterPre();
+        preFilter = (Filter) preFilter.accept(flattener, null);
+        preFilter.accept(this, query);
```
ZakarFin, line 92: What's the logic behind sometimes assigning variable with object returned by accept() and sometimes not? Does `accept()` mutate it's input or return a new instance or mutates and returns the mutated instance? I guess it both mutates and returns the original instance? It would be easier to read if the return value is either always or never assigned. If there's a reason maybe add a comment about it?

service-wfs3/src/main/java/org/oskari/service/wfs3/OskariWFS3Client.java
```diff
@@ -140,29 +145,49 @@ public static SimpleFeatureCollection getFeatures(OskariLayer layer,
...
+            return new PaginatedFeatureCollection(pages, schema, "FeatureCollection", hardLimit);
         } catch (IOException e) {
             throw new ServiceRuntimeException("IOException occured", e);
         } catch (MismatchedDimensionException | TransformException e) {
             throw new ServiceRuntimeException("Projection transformation failed", e);
         }
+    }
+
+    private static int clamp(int value, int min, int max) {
+        return value > max ? max : value < min ? min : value;
```
ZakarFin, line 166: Would be more lines but easier to read in my opinion:
```suggestion
        if (value > max) {
            return max;
        } else if (value < min) {
            return min;
        }
        return value;
```

service-wfs3/src/main/java/org/oskari/service/wfs3/FilterToOAPIFCoreQuery.java
jampukka: The behaviour depends on the implementation of FilterVisitor. I'll just get rid of `SimpleAndFlatteningFilterVisitor` as it's not readable and is too complex for the use-case which boils down to one recursive functions with couple of instanceof checks.

jampukka: See https://github.com/oskariorg/oskari-server/pull/816/commits/b2951d8788daf364fc60fed847ed80f434f84829

PR #828, opened 2022-04-11
Print-Service fill pattern fix

okauppinen:
Update fill patterns to use same patterns than frontend:
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/mapping/mapmodule/oskariStyle/generator.ol.js#L356-L382
Don't know why pdfbox width looks bolder than ol/canvas (ppi differs?) but tested that frontend 4 and pdfbox 3 looks the same.

Update icon stroke to black.
In frontend style generator doesn't handle icon stroke (and it's not documented) but mapmodule handles in getSvg:
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/mapping/mapmodule/oskariStyle/generator.ol.js#L473-L475
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/mapping/mapmodule/AbstractMapModule.js#L1919-L1924

### DIFF
--- a/service-print/src/main/java/org/oskari/print/util/StyleUtil.java
+++ b/service-print/src/main/java/org/oskari/print/util/StyleUtil.java
@@ -41,7 +41,7 @@ import fi.nls.oskari.util.JSONHelper;
 public class StyleUtil {
 
     private static final String SVG_MARKERS_JSON = "svg-markers.json";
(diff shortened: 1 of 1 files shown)


jampukka:
96 ("default screen" DPI / 72 (PDF DPI) = 4/3 so that would make sense

PR #830, opened 2022-04-12
Announcements bundle localization

osqq:
- Alter announcements table structure to accommodate localization and migrate existing data
- Change title and content into locale, return locale as json object


ZakarFin:
Looks good but split the migration into three migration scripts:

1) create the new `locale` column
2) migrate content from `title` and `content` to `locale`
3) drop `title` and `content` columns

This way we make sure that content isn't lost if the server goes down or an exception occurs etc.

ZakarFin:
Just for clarification: This is the server side change for oskariorg/oskari-frontend#1800. They need to be merged at the same time.

PR #835, opened 2022-04-22
Processing of Helsinki metadata

willberthos:
Parsing of metadata-uuid is now dependent of allowed domains list that is created from the properties.

### DIFF
--- a/service-capabilities/src/main/java/org/oskari/capabilities/LayerCapabilities.java
+++ b/service-capabilities/src/main/java/org/oskari/capabilities/LayerCapabilities.java
@@ -86,6 +86,8 @@ public class LayerCapabilities {
     }
 
     public void addCapabilityData(String key, Object value) {
-        typeSpecific.put(key, value);
+        if (value != null ) {
+            typeSpecific.put(key, value);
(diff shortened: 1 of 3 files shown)


service-capabilities/src/main/java/org/oskari/capabilities/LayerCapabilities.java
```diff
@@ -86,6 +86,8 @@ public Map<String, Object> getTypeSpecific() {
     }
 
     public void addCapabilityData(String key, Object value) {
-        typeSpecific.put(key, value);
+        if (value != null ) {
```
ZakarFin, line 89:
```suggestion
        if (value != null) {
```

ZakarFin:
Adds handling for property `service.metadata.domains` as comma-separated list in oskari-ext.properties. If the property is defined the metadata uuid parsing is only done for the listed domains. This helps with cases where layer capabilities point to other metadata catalogues than the one configured for the Oskari instance. Also allows allowing multiple domains like `paikkatietohakemisto.fi, geonetwork.nls.fi` that point to the same service.

PR #845, opened 2022-05-05
Refactor Capabilities for WMTS

ZakarFin:
Shorten JSON serialization `identifier` -> `id`, `projection` -> `crs` etc requiring new update migration for database (included).

Printing now uses WMTS capabilities from `oskari_maplayer` instead of querying them at runtime.
GetLayerCapabilities also changed (requires frontend change as well).

Longer term goal is to have `OskariLayer.getCapabilities()` return `LayerCapabilities` object instead of `JSONObject`. This is an initial step towards that.


jampukka:
Very nice

PR #848, opened 2022-05-13
Feature/mydata date fields

Creaticide:
Migrations etc for created/updated fields for mydata bundle in front end.


content-resources/src/main/resources/flyway/oskari/V2_8_2__oskari_appsetup_updated_field.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE oskari_appsetup
+add "updated" TIMESTAMP WITH TIME ZONE;
```
ZakarFin, line 2:
```suggestion
    ALTER COLUMN created TYPE TIMESTAMP WITH TIME ZONE
    ALTER COLUMN used TYPE TIMESTAMP WITH TIME ZONE
    ADD "updated" TIMESTAMP WITH TIME ZONE;
```

content-resources/src/main/resources/flyway/oskari/V2_8_3__oskari_appsetup_update_trigger.sql
```diff
@@ -0,0 +1,15 @@
+DROP TRIGGER IF EXISTS trigger_appsetup_update ON oskari_appsetup;
```
ZakarFin, line 1: Don't use a db trigger. Update value in code instead.

ZakarFin, line 1: Don't use a db trigger. Update value in code instead.

PR #849, opened 2022-05-17
Announcements refactor

okauppinen:
Use jackson to parse json to announcement. 
Migrate date to timestamptz and active to options. Add new model for migrated data.

Maybe endDate should be + 24 hours
and getAnnouncements -> getAllAnnouncements


control-announcements/src/main/java/org/oskari/announcements/helpers/AnnouncementsParser.java
```diff
@@ -12,6 +12,7 @@
 /**
  * Announcements parser
  */
+@Deprecated
```
ZakarFin, line 15: This could be removed if it's not used by this functionality anymore.

control-announcements/src/main/java/org/oskari/announcements/helpers/Announcement.java
```diff
@@ -1,5 +1,6 @@
 package org.oskari.announcements.helpers;
 
+@Deprecated
```
ZakarFin, line 3: Remove this as well in favor of the new model.

control-announcements/src/main/java/org/oskari/announcements/actions/AnnouncementsHandler.java
```diff
@@ -21,13 +23,17 @@ public class AnnouncementsHandler extends RestActionHandler {
...
+            if (params.getUser().isAdmin()) {
+                announcements = service.getAnnouncements();
             } else {
-                result = service.getAnnouncements();
+                announcements = service.getActiveAnnouncements();
+            }
+            JSONArray result = new JSONArray();
+            for (Announcement announcement : announcements) {
+                result.put(announcement.asJSON());
             }
```
ZakarFin, line 35: Could we use AnnouncementsHelper.writeJSON() instead and get rid off asJSON() in announcement?

PR #868, opened 2022-08-05
Show indicators without data.

Creaticide:
Show indicators without data.

### DIFF
--- a/control-statistics/src/main/java/fi/nls/oskari/control/statistics/GetIndicatorListHandler.java
+++ b/control-statistics/src/main/java/fi/nls/oskari/control/statistics/GetIndicatorListHandler.java
@@ -78,7 +78,7 @@ public class GetIndicatorListHandler extends ActionHandler {
     }
 
     private Optional<JSONObject> toJSON(StatisticalIndicator indicator, String language) {
-        if(indicator == null || indicator.getLayers() == null || indicator.getLayers().isEmpty()) {
+        if(indicator == null) {
             return Optional.empty();
         }
         try {


control-statistics/src/main/java/fi/nls/oskari/control/statistics/GetIndicatorListHandler.java
```diff
@@ -78,7 +78,7 @@ JSONObject getIndicatorsListJSON(long datasourceId, User user, String language)
     }
 
     private Optional<JSONObject> toJSON(StatisticalIndicator indicator, String language) {
-        if(indicator == null || indicator.getLayers() == null || indicator.getLayers().isEmpty()) {
+        if(indicator == null) {
```
jampukka, line 81: if `indicator.getLayers()` returns `null` then the whole `toJSON(StatisticalIndicator indicator, String language)` throws and catches the NullPointerException and still returns `Optional.empty()`. It might be that that `StatisticalIndicator#getLayers` never returns null (maybe an empty collection is stored in the database always) but that's not something I'd trust.

ZakarFin:
I'm not sure why would we want to show/list indicators without data? This will probably mean that there can be a lot of indicators that are listed but will feel broken since the user can't use them in anyway on some instances (like the arctic-sdi geoportal)?

Maybe find some alternative solution to this for user generated indicators if that is the reason for this change?

ZakarFin:
More user indicator related approach in #881. Closing this as the other implementation does the same thing and this one is opening more flood gates than what is needed.

PR #870, opened 2022-08-16
Fix main maplayer group creation

kuosman:
Now when saving new parent theme group the parentId is saved incorrectly saved as 0 (and for this reason the group is visible in the front only during the session).

The bug can be reproduced at [demo.oskari.org](https://demo.oskari.org/):
* [ ] login admin/oskari user
* [ ] add and save a new main theme group in the layerlist
* [ ] now the newly created group is shown in the layrlist
* [ ] refresh page and group is now visible anymore because new group gets parentId as 0

This PR fixes at main group parentId is -1.


ZakarFin:
Awesome, thanks!

PR #878, opened 2022-09-05
Ignore timeseries value when there is no resolution present

ZakarFin:
Some services have timeseries described like this:
```
<Extent name="time" default="9999-99" nearestValue="0">1978-11/2015-12</Extent> 
```
If we added code to detect a resolution we might be able to make an educated guess that the above could be P1Y or P1M. But for now lets skip the timeseries when we need to do guessing. If there is only single value in the times array the frontend breaks visually so filtering them out.


jampukka:
Agree with not guessing. Better to ask user (admin) adding the layer for the information, even better the user asks the service provider to add the information.

PR #883, opened 2022-09-22
WFSSearchChannel: get region from a property

jampukka:
Allow feature property to be used for the value of `SearchResultItem.region` as discussed in https://github.com/oskariorg/oskari-docs/issues/280.

Stored in database table `oskari_wfs_search_channels` inside column `config`:
```
{
   ... rest of the config ...,
  "region-property": "foobar",
}
```

Example SQL for adding the value to existing config value (there's no UI for this yet)
`UPDATE oskari_wfs_search_channels SET config = (config::jsonb || '{"region-property": "my_prop"}'::jsonb) where id = :id`


ZakarFin:
Nice 👍

PR #887, opened 2022-10-10
Set max total for Redis pool

ZakarFin:


jampukka:
This was never set before? Nice job

ZakarFin:
Kinda hard to believe so I checked like 5 times, but still couldn't find it set. After a bit of searching it looks like max total should default to 8, but I'm not sure if there's other logic involved since with our defaults the `maxIdle` should be 15. Anyway, noticed this while trying to figure out why when running Oskari in Azure App Service the Redis pool gets exhausted (during the nightly UNStats syncing scheduled job) while in any other environment it doesn't.

ZakarFin:
The problem with Azure that I'm having is probably because we should use JedisCluster with the Azure env instead of JedisPool. However these classes in Jedis don't share an interface that could be used to run commands on with 3.8.0 of Jedis that we use.

Jedis 4.0 seems to have a fix for this issue with a new class JedisPooled that looks compatible with JedisCluster so we could have code that initializes the shared Redis client based on setup and use the same client reference to run commands on both kind of environments. However the Spring Data 2.x version that we use is built using Jedis 3.8 and looking at the release notes of Jedis 4 it seems obvious that using it instead of 3.x with Spring will not work.

PR #900, opened 2022-11-28
WMTS layer null legend fix

okauppinen:
If style doesn't have legend, capabilities styles for wmts layer is parsed as:
`"styles": [{
        "default": true,
        "legend": null,
        "name": "default",
        "title": "default"
 }]`
String for JSONObject.null is 'null'. Changed to use JSONHelper which handles null more properly. This fixes issue where wmts layer could have 'null' or proxyUrl in frontend and layer is handled as it has legend url and tried to get legend image for layer.


jampukka:
Looks good

ZakarFin:
Related to oskariorg/oskari-server#880, but 880 handles it "too late" for the frontend to work properly.

PR #901, opened 2022-12-08
Add a helper for instance customization options

ZakarFin:
Previously there was an almost identical `svg-markers.json` in both `control-base` and `service-print` modules. One in `control-base` had hardcoded colors and print used placeholders. These can now be queried from a new `org.oskari.util.Customization` helper class (in `service-base`): `JSONArray svgMarkers = Customization.getMarkers();`. This returns the markers with the placeholders so GetAppSetup needs to replace those with the existing hard coded `#000000` values.

The service logo can now be fetched as byte array with: `Customization.getLogo()`. This is used by printing to draw the service logo on the printout. This also adds an action route to get the logo with http://service.org/action/Logo. This enables using the same logo on the frontend with for example the "logo plugin" on the map or even the browser favicon. This makes customizing an Oskari instance easier.

The logo for printing can still be configured with oskari-ext.properties like before with the `print.logo.path`, but also with `logo.path.print` and it also enables configuring a global logo with `logo.path` and adding different kinds of logo images for different purposes with other qualifiers if needed. The logo defaults to the Oskari logo found under `service-base` resources.

This could also be improved to enable a service to override marker SVGs more easily since they are fetched from single source to both frontend and printing.


service-base/src/main/java/org/oskari/util/Customization.java
```diff
@@ -0,0 +1,96 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class Customization {
+    private static final String SVG_MARKERS_JSON = "svg-markers.json";
+    private static Cache<String> CUSTOMIZATION = CacheManager.getCache("Customization");
```
ZakarFin, line 21: This is a bit of an overkill if we can't get the logo byte[] here as well. Maybe even if we could store the logos here.

PR #906, opened 2023-01-09
Coordinate information to print page

willberthos:
Printed map now shows ETRS-TM35FIN coordinates on corners

### DIFF
--- a/service-print/src/main/java/org/oskari/print/PDF.java
+++ b/service-print/src/main/java/org/oskari/print/PDF.java
@@ -126,17 +126,18 @@ public class PDF {
     private static final float LOGO_HEIGHT = PDFBoxUtil.mmToPt(9);
 
     private static final float OFFSET_SCALE_BOTTOM = PDFBoxUtil.mmToPt(5);
+    private static final float OFFSET_SCALE_BOTTOM_Y = PDFBoxUtil.mmToPt(3);
 
     private static final float OFFSET_TIMESERIES_RIGHT = PDFBoxUtil.mmToPt(50);
     private static final float OFFSET_TIMESERIES_LABEL_BOTTOM = PDFBoxUtil.mmToPt(10);
     private static final float OFFSET_TIME_IN_TIMESERIES_BOTTOM = PDFBoxUtil.mmToPt(5);
     
-    private static final float OFFSET_CROSS_UPPER = PDFBoxUtil.mmToPt(7);
-    private static final float OFFSET_CROSS_LOWER = PDFBoxUtil.mmToPt(7);
+    private static final float OFFSET_CROSS_TOP = PDFBoxUtil.mmToPt(7);
+    private static final float OFFSET_CROSS_BOTTOM = PDFBoxUtil.mmToPt(7);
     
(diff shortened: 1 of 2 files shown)


service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -137,6 +147,9 @@ public class PDF {
     private static final String MARKER_COORD_SEPARATOR= "_";
 
     private static final FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2(null);
+    
+    private static final ResourceBundle rb = ResourceBundle.getBundle("messages");
```
ZakarFin, line 151: We should rename "messages" to something else to avoid confusion when theres another localization file set called messages. Maybe just rename to "print-messages"?

service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -321,6 +336,79 @@ private static void drawTimeseriesTexts(PDPageContentStream stream,
...
+    }
+
+    private static void drawCross(PDPageContentStream stream, float x1, float x2, float y1, float y2) throws IOException {
+        stream.moveTo(x1, y1);
+        stream.lineTo(x1, y2);
+        stream.lineTo(x2, y2);
+        stream.stroke();
+    }
+    
+    private static int[] getCornerCoords(PrintRequest request) {
```
ZakarFin, line 401: We shouldn't round to int for all projections. Maybe just keep as is without rounding for now?

service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -321,6 +336,79 @@ private static void drawTimeseriesTexts(PDPageContentStream stream,
...
+        float y1 = y2 - OFFSET_CROSS_BOTTOM;
+        
+        drawCross(stream, x1, x2, y1, y2);
+        
+        float textX = x1 - OFFSET_TEXT_BOTTOM_X;
+        float textY = y1 + OFFSET_TEXT_BOTTOM_Y;
+        float textY2 = textY - OFFSET_TEXT_Y;
+        
+        String coords = rb.getString("coordinates.lower") + ": " + bottomLat + ", " + bottomLon;
+        String projection = "ETRS-TM35FIN" + " " + rb.getString("coordinates");
```
ZakarFin, line 388: We shouldn't hard-code "ETRS-TM35FIN". Get it from the messages based on projection with fallback to the EPSG-code.

service-print/src/main/java/org/oskari/print/PDF.java
```diff
@@ -321,6 +336,79 @@ private static void drawTimeseriesTexts(PDPageContentStream stream,
                 x, OFFSET_TIME_IN_TIMESERIES_BOTTOM);
     }
 
+    private static void drawCornerCoords(PDPageContentStream stream, PrintRequest request, float x, float y, float mapHeight, float mapWidth, PDRectangle pageSize)
```
ZakarFin, line 339: Page size isn't used so we can remove it from params:

```suggestion
    private static void drawCornerCoords(PDPageContentStream stream, PrintRequest request, float x, float y, float mapHeight, float mapWidth)
```

ZakarFin:
Included in #1035

PR #918, opened 2023-02-02
PrintService create icons fix

okauppinen:
PrintService requires xmlns in svg. This resolves issue with printing vector layers or markers.

Added simple tests for creating icons and styles. Mostly to catch exceptions.

org.apache.batik.transcoder.TranscoderException
Root element namespace does not match that requested:
Requested: http://www.w3.org/2000/svg
Found: null

TODO: had to remove xmlns from EnvHelper to run GetAppSetupHandler tests.


control-base/src/main/java/fi/nls/oskari/util/EnvHelper.java
```diff
@@ -125,8 +127,10 @@ private static JSONArray getMarkers() {
                 JSONObject marker = svgMarkers.optJSONObject(i);
                 String svg = marker.optString("data");
                 String updated = svg
-                        .replace("'$fill'", "'#000000'")
-                        .replace("'$stroke'", "'#000000'");
+                        .replace(PLACEHOLDER_FILL, "#000000")
+                        .replace(PLACEHOLDER_STROKE, "#000000")
+                        // TODO: GetAppSetupHandlerTest fails with xmlns. Should we update json resources??
```
jampukka, line 132: This is weird

okauppinen: I agree. Maybe we should replace placeholders in frontend or add Customization.getMarkers(stroke, fill) and fix/modify GetAppSetupHandler test resources to match svg-markers.json. Maybe by using placeholders in tests and replacing them by svg-markers.json.

For now both EnvHelper and PrintService is using same resource file:
https://github.com/oskariorg/oskari-server/blob/develop/service-base/src/main/resources/org/oskari/util/svg-markers.json

jampukka: Using same resource file should be the goal, which tests are failing?

ZakarFin: We previously had the markers in two resource files so I broke this in #901. The printing used placeholders on it's resource files and the one used be EnvHelper had those hard-coded values. So this was more of a backwards compatibility thing for EnvHelper than anything else. I _think_ the frontend code might replace the fills etc anyways but I was lazy to test it so just had them work as they did before.

It's the same thing with the namespace. The printing resource had them while the EnvHelper one didn't. We could just update the expected results from GetAppSetup related JUnit tests to have the namespace in the expected results if it works on the frontend with the namespace included (it should but they might get wrapped in some other SVG currently for scaling purposes on the frontend so would need to test it).

ZakarFin: I guess we have atleast two options here:

1) Pass marker namespace and color placeholders to frontend

Requires testing that it works on frontend and updating tests for GetAppSetup to either include the marker changes in expected test results or skip checking markers or make an additional check for them that checks that they match the built-in marker SVG.

2) Just accept doing a weird thing here

I'm fine with this option since we don't really test nor document how to override the markers to begin with and I would like to return to this and validate the markers in a some meaningful way (checking for placeholders etc) once we actually work on customizing them.

ZakarFin:
Merging this to fix issues with vector features. We have an upcoming refactoring planned for getting rid of the replace() calls etc.

PR #922, opened 2023-02-13
Layerlist view modifier

okauppinen:
Check if backendstatus bundle is available for layerlist bundle. Add possibility to use groupToggleLimit via properties.

https://github.com/oskariorg/oskari-frontend/pull/2123

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/view/modifier/bundle/LayerlistHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/view/modifier/bundle/LayerlistHandler.java
@@ -1,0 +1,27 @@
+package fi.nls.oskari.control.view.modifier.bundle;
+
+import fi.nls.oskari.annotation.OskariViewModifier;
+import fi.nls.oskari.util.JSONHelper;
+import fi.nls.oskari.util.PropertyUtil;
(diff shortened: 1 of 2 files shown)


control-base/src/main/java/fi/nls/oskari/control/view/modifier/bundle/LayerlistHandler.java
```diff
@@ -0,0 +1,27 @@
...
+import fi.nls.oskari.util.JSONHelper;
+import fi.nls.oskari.util.PropertyUtil;
+import fi.nls.oskari.view.modifier.ModifierException;
+import fi.nls.oskari.view.modifier.ModifierParams;
+import org.json.JSONObject;
+
+@OskariViewModifier("layerlist")
+public class LayerlistHandler extends BundleHandler {
+
+    private static final String BACKEND_STATUS_BUNDLE = "backendstatus";
```
ZakarFin, line 13: Move the bundle reference to ViewModifier as that is where all the other are listed: https://github.com/oskariorg/oskari-server/blob/develop/service-control/src/main/java/fi/nls/oskari/view/modifier/ViewModifier.java#L19-L43

PR #924, opened 2023-02-14
Allow hyphen/dash in email domain name

koivunen:
Hyphen (dash) is the only spacing character allowed in a domain name


ZakarFin:
Thanks!

PR #933, opened 2023-03-13
Admin vector styles

okauppinen:
Handle vector styles. Added own methods for admin to check that creator is handled correctly	(overreacted ?).

For layer update, it's little confusing to catch changes for vector styles. Proposal:
- Add vector style status -> Had to extend maplayer (MapLayerAdminInput)
- Handle update by status

https://github.com/oskariorg/oskari-frontend/pull/2167

### DIFF
--- a/service-map/src/main/java/org/oskari/maplayer/model/MapLayerAdminInput.java
+++ b/service-map/src/main/java/org/oskari/maplayer/model/MapLayerAdminInput.java
@@ -1,0 +1,18 @@
+package org.oskari.maplayer.model;
+
+import java.util.Map;
+
+/**
+ * Used as input data model for layers in the admin UI.
+ */
+public class MapLayerAdminInput extends MapLayer {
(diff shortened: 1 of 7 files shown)


service-map/src/main/java/org/oskari/maplayer/model/MapLayerAdminInput.java
```diff
@@ -0,0 +1,18 @@
+package org.oskari.maplayer.model;
+
+import java.util.Map;
+
+/**
+ * Used as input data model for layers in the admin UI.
+ */
+public class MapLayerAdminInput extends MapLayer {
```
ZakarFin, line 8: Maybe we should refer to this as MapLayerVector instead of MapLayerAdminInput as all of these are more of an extension for vector layers than input from MapLayerAdmin? Or should we just add these to the MapLayer class directly? These are used for initializing layers from migrations as well.

PR #936, opened 2023-03-14
Fix/admin vector style

okauppinen:
Changes:
- don't ignore layerId, created and updated from json as UserStyles tab needs them. Added time formatter because Jackson doesn't support OffsetDateTime by default (requires jackson-datatype-jsr310 module)
- safety checks for inserting null layerId and limit 1 for getdefaultstyle

Bug fixes:
- add updated time
- remove layerId and creator from update query. only save/insert set layerId so in update layerId was removed
- if style exist and wasn't modified => status caused NPE on admin layer save

### DIFF
--- a/service-base/src/main/java/fi/nls/oskari/domain/map/style/VectorStyle.java
+++ b/service-base/src/main/java/fi/nls/oskari/domain/map/style/VectorStyle.java
@@ -36,7 +36,6 @@ public class VectorStyle implements Serializable {
         this.id = id;
     }
 
-    @JsonIgnore
     public Integer getLayerId() {
         return layerId;
     }
@@ -88,7 +87,6 @@ public class VectorStyle implements Serializable {
         this.style = new JSONObject(style);
     }
 
-    @JsonIgnore
     public OffsetDateTime getCreated() {
         return created;
     }
@@ -97,7 +95,6 @@ public class VectorStyle implements Serializable {
         this.created = created;
     }
 
-    @JsonIgnore
     public OffsetDateTime getUpdated() {
         return updated;
     }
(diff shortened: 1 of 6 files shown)


service-base/src/main/java/fi/nls/oskari/domain/map/style/VectorStyle.java
```diff
@@ -18,6 +19,7 @@ public class VectorStyle implements Serializable {
     public static final String TYPE_OSKARI = "oskari";
     public static final String TYPE_MAPBOX = "mapbox";
     public static final String TYPE_3D = "cesium";
+    private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ISO_DATE_TIME;
```
jampukka, line 22: From [JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_DATE_TIME): `DateTimeFormatter.ISO_DATE_TIME` The ISO-like date-time formatter that formats or parses a date-time with the offset and zone if available, such as '2011-12-03T11:15:30', '2011-12-03T11:15:30+01:00' or '2011-12-03T11:15:30+01:00[Europe/Paris]'. 

Consider `DateTimeFormatter.ISO_INSTANT` if UTC can be used (`"2011-12-03T10:15:30Z"`) or `DateTimeFormatter.ISO_OFFSET_DATE_TIME` (`"2011-12-03T12:15:30+02:00"`) if we want offset to appear in the string. ZoneId as string (`[Europe/Paris]`) would be weird in the output.

jampukka: In addition, I wouldn't oppose to adding jackson-datatype-jsr310 as a dependency to avoid all this

okauppinen: Changed Announcement to use ISO_INSTANT also. Times are formatted with Date in frontend which seems to fail with ZoneId string.

ZakarFin: Updated this to match: https://github.com/oskariorg/oskari-docs/commit/af020afbb257db0d51339d23d8a944ecd7b69e7b

PR #938, opened 2023-03-15
remove styles from options

okauppinen:
Remove `styles` and `externalStyles` from wfs, tiles3d and vectortile layer options.
Remove temporal solution where VectorStyles were added to layer options.
Remove named style handling from WFSLayerOptions and handle only 'default' named feature style for UserData layers.
Move create default/empty category to MyPlacesService and fallback to WFSLayerOptions default style in VectorStyleService


ZakarFin:
Note that https://github.com/oskariorg/oskari-server/pull/947 used the version number 2.11.3 so this will need to be updated.

PR #940, opened 2023-03-24
Coords to printout

ZakarFin:
Adding a work-in-progress branch for finishing up: #906 "Printed map now shows ETRS-TM35FIN coordinates on corners"

TODOs from other pr:

- We should rename "messages.properties" to something else to avoid confusion when theres another localization file set called messages. Maybe just rename to "print-messages"?
- We shouldn't round coordinates to int for all projections. Maybe just keep as is without rounding for now?
- We shouldn't hard-code "ETRS-TM35FIN". Get it from the messages based on projection with fallback to the EPSG-code.
- Page size isn't used in drawCornerCoords() so we can remove it from params.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -1,5 +1,14 @@
 # Release Notes
 
+## 2.10.1
+
+For a full list of changes see: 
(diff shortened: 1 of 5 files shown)


ZakarFin:
Included in #1035

PR #942, opened 2023-03-30
Vector style handler

okauppinen:
Add handleDelete and rename permission check.
Return updated/saved style from db. At least saved style needs created date from db.
Fix mapper deleteStyle.

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/style/VectorStyleHandler.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/style/VectorStyleHandler.java
@@ -39,10 +39,11 @@ public class VectorStyleHandler extends RestActionHandler {
             throw new ActionDeniedException("Session expired");
         }
         final long userId = params.getUser().getId();
+        final VectorStyle style = VectorStyleHelper.readJSON(params.getPayLoad());
         try {
-            final VectorStyle style = VectorStyleHelper.readJSON(params.getPayLoad());
             style.setCreator(userId);
             long id = getService().saveStyle(style);
+            style.setId(id);
             AuditLog.user(params.getClientIp(), params.getUser())
                     .withParam("id", id)
                     .withParam("name", style.getName())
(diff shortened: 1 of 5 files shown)


service-map/src/main/java/fi/nls/oskari/map/style/VectorStyleMapper.java
```diff
@@ -36,7 +36,7 @@ public interface VectorStyleMapper {
     @Select("SELECT * FROM oskari_maplayer_style WHERE layer_id = #{layerId} AND (creator IS NULL OR creator=#{user})")
     List<VectorStyle> getStyles(@Param("user") long user, @Param("layerId") int layerId);
 
-    @Delete("DELETE FROM oskari_maplayer_style WHERE id = #{id} RETURNING id")
+    @Select("DELETE FROM oskari_maplayer_style WHERE id = #{id} RETURNING id")
```
ZakarFin, line 39: I don't think we really need the "returning id" here. Is that why Select-annotation is used instead of Delete? I would prefer the Delete and without the returning id if that is the reason for using Select?

okauppinen: You are right that id isn't used/needed because handler returns true if everything goes right. Changed delete to select because mapper didn't delete row from db. Not sure if returning id causes this or how @Delete should work at all. Checked from Announcements mapper and there select is used for delete.

okauppinen: Missing commit caused that delete statement/annotation didn't work. Maybe MyBatis notices that @Select delete updates data in db so it auto-commits.

control-base/src/main/java/fi/nls/oskari/control/style/VectorStyleHandler.java
```diff
@@ -39,32 +39,33 @@ public void handlePost(final ActionParameters params) throws ActionException {
...
             throw new ActionException("Error when trying add vector style", e);
         }
-        ResponseHelper.writeResponse(params, true);
+
     }
     public void handlePut(final ActionParameters params) throws ActionException {
         User user = params.getUser();
         if (user.isGuest()) {
             throw new ActionDeniedException("Session expired");
         }
```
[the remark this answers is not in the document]
okauppinen: changed

PR #948, opened 2023-04-14
VectorStyle migrate layer default style

okauppinen:
Migrate layer default style if layer has style. If default style isn't found then first style is marked as default. 

BEFORE:
image
AFTER:
image
STYLES:
image

Note! not sure if internal = false should be added to getMaplayers. I think that internal layers doesn't have styles so style (default) isn't updated for them.


ZakarFin:
Internal layers are mainly used for the baselayers for myplaces etc. Maybe for region sets for statistical data as well, but neither really have anything to migrate so it should be ok to have them included in the migration. And even if they do have styles it makes more sense to do the migration for them as well.

PR #969, opened 2023-05-30
UserService role handling

okauppinen:
Changes:
- updated logging: info => debug
- ensureRolesInDB before linking roles and update to protected to enable override
- get guest username from properties like logged-in and admin
- move "guest" names from User to GuestUser

DatabaseUserService gets role for guest user from db (is_guest == true => first)
https://github.com/oskariorg/oskari-server/blob/develop/service-users/src/main/java/fi/nls/oskari/user/DatabaseUserService.java#L27-L31
https://github.com/oskariorg/oskari-server/blob/develop/service-users/src/main/java/fi/nls/oskari/user/MybatisRoleService.java#L166-L183

Maybe we should unify username. For now:
DB: user_name, UsersHandler.user2Json(): user, User: screenname, User.toJson(): nickName


ZakarFin:
Agree on having the same name for screen/nick/username. Username would be good, but it would require changing the User class which might break things. We should also add timestamp fields for users so we could change these at the same time.

PR #973, opened 2023-06-16
New routes for layer permissions bundle

ZakarFin:
`LayerPermissionHandler` would be replacement for `GetPermissionsLayerHandlers` and `SaveLayerPermissionHandler`. Also `GetAllRolesHandler` can be removed since the newer `ManageRoles` does basically the same thing.

The response from LayerPermissionHandler for GET request would be layers and their permissions for each role (that have permissions):
```
{
    "names": {
        "VIEW_PUBLISHED": "Katseluoikeus",
        "EDIT_LAYER_CONTENT": "Muokkaa tasoa",
        "DOWNLOAD": "Latausoikeus",
        "PUBLISH": "Julkaisuoikeus",
        "VIEW_LAYER": "Listausoikeus"
    },
    "layers": [{
            "id": 234,
            "name": "Layer name 1",
            "permissions": {
                "12": [
                    "VIEW_PUBLISHED",
                    "DOWNLOAD",
                    "VIEW_LAYER"
                ],
                "43": [
                    "VIEW_PUBLISHED",
                    "DOWNLOAD",
                    "VIEW_LAYER"
                ],
                "23": [
                    "VIEW_PUBLISHED",
                    "DOWNLOAD",
                    "PUBLISH",
                    "VIEW_LAYER"
                ]
            }
        },
        {
            "id": 76,
            "name": "Layer name 35",
            "permissions": {
                "12": [
                    "VIEW_PUBLISHED",
                    "DOWNLOAD",
                    "VIEW_LAYER"
                ],
                "43": [
                    "VIEW_PUBLISHED",
                    "DOWNLOAD",
                    "PUBLISH",
                    "VIEW_LAYER"
                ],
                "23": [
                    "VIEW_PUBLISHED",
                    "DOWNLOAD",
                    "PUBLISH",
                    "VIEW_LAYER"
                ]
            }
    }]
}
```
Where "12", "43" and "23" are role ids corresponding to roles like "Guest" or "User" or "Admin" and the array lists permissions that the role has.

Sending id-parameter with a layer id can be used to get permissions for single layer.

### DIFF
--- a/control-admin/src/main/java/fi/nls/oskari/control/admin/AbstractLayerAdminHandler.java
+++ b/control-admin/src/main/java/fi/nls/oskari/control/admin/AbstractLayerAdminHandler.java
@@ -78,7 +78,7 @@ public abstract class AbstractLayerAdminHandler extends RestActionHandler {
         }
         return roles;
     }
-    private Set<String> getAvailablePermissions() {
+    protected Set<String> getAvailablePermissions() {
         if (availablePermissionTypes != null) {
             return availablePermissionTypes;
         }
--- a/control-admin/src/main/java/fi/nls/oskari/control/admin/GetAllRolesHandler.java
+++ b/control-admin/src/main/java/fi/nls/oskari/control/admin/GetAllRolesHandler.java
@@ -15,9 +15,11 @@ import org.json.JSONObject;
 
 /**
  * Route returning all roles in the system.
+ * @deprecated Use ManageRoles instead
  * @author SMAKINEN
  */
 @OskariActionRoute("GetAllRoles")
+@Deprecated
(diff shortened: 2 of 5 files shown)


ZakarFin:
Might need to tune the JSON format still. This is about 8megs worth of json with ~2500 layers and ~20 roles/5 permission types.

ZakarFin:
After https://github.com/oskariorg/oskari-server/pull/973/commits/9023a50499a86b551856cc71c9161fb09a628c0d the same result is ~660kb which seems more doable.

ZakarFin:
I'll merge this for now. The saving functionality still need to be added to this route (based on SaveLayerPermissions, but we should take a look at what format we want to use for the payload etc).

PR #974, opened 2023-06-20
register new featuredata bundle in database

DenverCoder544:


content-resources/src/main/java/flyway/oskari/V2_12_1__register_featuredata_bundle.java
```diff
@@ -0,0 +1,18 @@
+package flyway.oskari;
+
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.oskari.helpers.BundleHelper;
+
+import java.sql.Connection;
+
+/**
+ * Register mydata bundle as intended replacement for personaldata bundle.
```
ZakarFin, line 10:
```suggestion
 * Register featuredata bundle (React impl) as intended replacement for featuredata2 bundle (jQuery impl).
```

PR #976, opened 2023-06-22
VectorFeatureWriterHandler: fix axisOrder issue when CRS is North/East

jampukka:
VectorFeatureWriterHandler correctly reads GeoJSON as lon/lat to JTS x/y order. However, when generating WFS 1.1.0 `<wfs:Transaction>` it fails to respect the axis order defined by the EPSG definition.

No changes to any current `public static` function (header or functionality).

### DIFF
--- a/control-base/src/main/java/fi/nls/oskari/control/feature/FeatureWFSTRequestBuilder.java
+++ b/control-base/src/main/java/fi/nls/oskari/control/feature/FeatureWFSTRequestBuilder.java
@@ -2,8 +2,6 @@ package fi.nls.oskari.control.feature;
 
 import fi.nls.oskari.domain.map.Feature;
 
-import fi.nls.oskari.log.LogFactory;
-import fi.nls.oskari.log.Logger;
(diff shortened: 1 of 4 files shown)


jampukka:
Yeah, uhm, this sort of works, but only if `ProjectionHelper.isFirstAxisNorth(crs);` returns the _correct_ answer. And that doesn't happen if a GeoServer with default settings (forceXY=true) is running in the same servlet container, which is not that uncommon for Oskari installations.

Currently we are outputting `srsName="http://www.opengis.net/def/crs/EPSG/0/3879"` which should mean longitude/latitude (at least when [GeoServer outputs geometries](https://docs.geoserver.org/main/en/user/services/wfs/axis_order.html)). However, when reading the geometry from a WFS-T request GeoServer honors the WFS 1.1.0 spec of respecting the axis order of the EPSG definition and handles it as latitude/longitude. With `srsName="EPSG:3879"` GeoServer falls back to the assumption of forceXY=true and reads it as longitude/latitude.

PR #981, opened 2023-08-08
Describe layer wfs

okauppinen:
Sort and add hidden if layer has filtered/selected properties. Add format for properties. Tried list.add(index, prop) but had some problems with it so added json ignored order field for props and sorting.

Add required values from options and attributes to response and remove them from JSON formater.

For: https://github.com/oskariorg/oskari-frontend/pull/2371

image

UPDATED:
- `data` => `controlData`. Attributes, options, params, metadata etc. is reserved for other purposes. So hard to figure out naming
- removed wpsInputType
- Changed UserData formatters return `describeLayer` object instead of creating attributes or options objects. didn't remove parseAttributes() (maybe needed for migrations)
- describeLayer can be handled in frontend like DescribeLayerHandler response as predeined info
- update/add test for formatters
- geometryType -> styleType. actual geometryType is inside properties (type 'geometry' and rawType)

Fixes after update:
- Add attention_text and geometry (GeometryPropertyType -> styleType collection) for MyPlaces props
- Add default geometry type to Analysis props. Analysis doesn't have rawType (number -> int, double,..)
- Fix userlayer styleType, type -> rawType and add (simple) type
- Had to change WFSConversionHelper.getSimpleType to use lower case comparison as UserLayer uses upper case.

UPDATED II:
- Don't remove attributes (or data). not sure how smart it was to remove options and gather values (capa, attr, opts) to new controlData. Idea was to reduce data to be loaded on appstart and use one variable in frontend (no need to parse data from many places, no need for setters, getters only) to have all data needed for rendering layer. Forget that options doesn't have styles anymore and attributes are used elsewhere (contrib or others).
- add isDefault and localized name (wfs type) for MyPlaces
- test isDefault, locale and name

image

image

image

image

BTW: using underscore as separator:  `analysis_userlayer_1_13` -> analysis 13 based on userlayer_1

### DIFF
--- a/control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java
+++ b/control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java
@@ -164,13 +164,14 @@ public class DescribeLayerHandler extends RestActionHandler {
         data.put(WFSLayerAttributes.KEY_COMMON_ID, attr.getCommonId());
         data.put(WFSLayerAttributes.KEY_ID_PROPERTY, attrData.optString(WFSLayerAttributes.KEY_ID_PROPERTY, null));
 
-        String geometryType = attrData.optString(WFSLayerAttributes.KEY_GEOMETRY_TYPE, null);
-        if (geometryType == null) {
+        // TODO: should admin store geometryType => styleType
+        String styleType = attrData.optString(WFSLayerAttributes.KEY_GEOMETRY_TYPE, null);
+        if (styleType == null) {
             String geomName = caps.getGeometryField();
             FeaturePropertyType fpt = caps.getFeatureProperty(geomName);
-            geometryType = WFSConversionHelper.getStyleType(fpt.type);
+            styleType = WFSConversionHelper.getStyleType(fpt.type);
         }
-        data.put(WFSLayerAttributes.KEY_GEOMETRY_TYPE, geometryType);
+        data.put(WFSLayerAttributes.KEY_STYLE_TYPE, styleType);
 
         data.put(WFSLayerOptions.KEY_RENDER_MODE, opts.getRenderMode());
         data.put(WFSLayerOptions.KEY_CLUSTER, opts.getClusteringDistance());
--- a/control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java
+++ b/control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java
@@ -111,26 +111,28 @@ public class DescribeLayerHandler extends RestActionHandler {
         return output;
     }
     private void setDetailsForWFS (LayerExtendedOutput output, OskariLayer layer, String lang) {
+        // UserDataLayers are handled in frontend by WFS plugin and embedded myplaces is using WFS type
+        // so LayerJSONFormatterUSERDATA gathers values from options and attributes in same way than this
         LayerCapabilitiesWFS caps = CapabilitiesService.fromJSON(layer.getCapabilities().toString(), layer.getType());
         WFSLayerAttributes attr = new WFSLayerAttributes(layer.getAttributes());
         WFSLayerOptions opts = new WFSLayerOptions(layer.getOptions());
         output.properties = getProperties(caps, attr, lang);
-        output.data = getData(caps, attr, opts);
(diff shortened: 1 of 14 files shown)


control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java
```diff
@@ -102,53 +100,83 @@ private LayerExtendedOutput getLayerDetails(ActionParameters params, OskariLayer
...
-        JSONObject labelsForLang = locale.optJSONObject(lang);
-        if (labelsForLang == null) {
-            return null;
+    private Map<String, Object> getData (LayerCapabilitiesWFS caps, WFSLayerAttributes attr, WFSLayerOptions opts) {
+        Map<String, Object> data = new HashMap<>();
+
+        JSONObject attrData = attr.getAttributesData();
+        data.put(WFSLayerAttributes.KEY_NO_DATA_VALUE, attr.getNoDataValue());
+        data.put(WFSLayerAttributes.KEY_COMMON_ID, attr.getCommonId());
+        data.put(WFSLayerAttributes.KEY_WPS_TYPE, attrData.optString(WFSLayerAttributes.KEY_WPS_TYPE, null));
```
ZakarFin, line 164: I think we should skip/not handle the WPS_TYPE for now since it's only for analysislayers and DescribeLayer only works with layers in oskari_maplayer table currently.

control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java
```diff
@@ -102,53 +100,83 @@ private LayerExtendedOutput getLayerDetails(ActionParameters params, OskariLayer
...
-        }
-        JSONObject labelsForLang = locale.optJSONObject(lang);
-        if (labelsForLang == null) {
-            return null;
+    private Map<String, Object> getData (LayerCapabilitiesWFS caps, WFSLayerAttributes attr, WFSLayerOptions opts) {
+        Map<String, Object> data = new HashMap<>();
+
+        JSONObject attrData = attr.getAttributesData();
+        data.put(WFSLayerAttributes.KEY_NO_DATA_VALUE, attr.getNoDataValue());
+        data.put(WFSLayerAttributes.KEY_COMMON_ID, attr.getCommonId());
```
ZakarFin, line 163: What's the difference between common id and id property? I would assume they both reference a property that should be used as an id?

control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java
okauppinen: removed

control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java
okauppinen: commonId is used to compare two layers and idProperty is workaround for admin.

have to rename idProperty
admin: "Use feature property as identifier" select: https://github.com/oskariorg/oskari-frontend/pull/2368 
fast test for usage: https://github.com/oskariorg/oskari-frontend/pull/2374

service-map/src/main/java/fi/nls/oskari/map/layer/formatters/LayerJSONFormatterWFS.java
```diff
@@ -18,6 +18,9 @@ public JSONObject getJSON(OskariLayer layer,
 
         final JSONObject layerJson = getBaseJSON(layer, lang, isSecure, crs);
         JSONHelper.putValue(layerJson, KEY_ISQUERYABLE, true);
+        // getBaseJSON adds these but model builder isn't using them. Frontend uses DescribeLayer response
+        layerJson.remove(KEY_OPTIONS);
+        layerJson.optJSONObject(KEY_ATTRIBUTES).remove("data");
```
ZakarFin, line 23: We might want to keep this still in the layer response because of https://github.com/oskariorg/oskari-frontend/blob/master/bundles/mapping/mapmodule/AbstractMapModule.js#L2346-L2360

okauppinen: removed

PR #987, opened 2023-08-15
Replace deprecated XMLHelper with new one

ZakarFin:
Requires a test case


ZakarFin:
It should, but I haven't had the time to test it really :)

ZakarFin:
I removed the try/catch since any uncaught exception is caught in https://github.com/oskariorg/oskari-server/blob/master/service-control/src/main/java/fi/nls/oskari/control/ActionControl.java#L146 and that throws the same ActionException as this one and includes the cause on it or in case of an ActionException, throws it as is. These will be logged here https://github.com/oskariorg/oskari-server/blob/master/servlet-map/src/main/java/fi/nls/oskari/AjaxController.java#L44-L51

PR #989, opened 2023-08-15
Update GeoTools 28.2 -> 29.2

ZakarFin:


ZakarFin:
Seems we can't upgrade just yet: "GeoTools 29.x requires Java 11 as the minimum version. If you are still using Java 8, you will have to remain on GeoTools 28.x or ealier."

PR #996, opened 2023-08-28
param for announcement api to send no-cache headers on response

ArttuPietarinen:
ability to send parameter to announcement api to put Cache-Control, no-cache and no-store headers to response of announcement GET and POST api calls.


ZakarFin:
Thanks! I'm not sure if there's need to use the param to do this but it doesn't hurt either.

PR #1000, opened 2023-09-06
User created and last login timestamps

DenverCoder544:


ZakarFin:
Looks good to me, but lets wait for 2.13

PR #1010, opened 2023-10-04
implement search methods

DenverCoder544:
-fetch features without wfs-t straight from db
-still missing some properties in the resulting geojson - bbox / crs etc.

### DIFF
--- a/service-myplaces/src/main/resources/fi/nls/oskari/myplaces/MyPlaceMapper.xml
+++ b/service-myplaces/src/main/resources/fi/nls/oskari/myplaces/MyPlaceMapper.xml
@@ -25,6 +25,21 @@
         <result property="attentionText" column="attention_text" />
     </resultMap>
 
+    <resultMap id="MyPlacesResult" type="MyPlace">
+        <result property="id" column="id" />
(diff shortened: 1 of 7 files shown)


service-myplaces/src/main/resources/fi/nls/oskari/myplaces/MyPlaceMapper.xml
```diff
@@ -25,6 +25,21 @@
         <result property="attentionText" column="attention_text" />
     </resultMap>
 
+    <resultMap id="MyPlacesResult" type="MyPlace">
```
ZakarFin, line 28: Should we just add this to the mapper-class instead in an effort to some day getting rid of the xml? It's always a bit trickier to find these when they can be on the Java-file with some parts in the xml.

ZakarFin: But we can do it later when we try combining myplaces and userlayers. Then the MyPlace-class will probably be replaced with something else and we'll have to re-do the mappers as well.

PR #1012, opened 2023-10-05
Check permissions when querying GFI

ZakarFin:
Formatting is in it's own commit so checking changes is easier to see when seeing changes per commit. Mainly this one: https://github.com/oskariorg/oskari-server/commit/6e448ed29993b778ee55ed7944d2dad91eb39641


okauppinen (approved):
Looks good but not sure what `acceptInternal` is for. For later use? However it works like before.

ZakarFin:
Currently permissionHelper allowed access to internal layers without further checks since access to these layers is handled differently. I could have used PermissionHelper without changes and check if the layer is internal (myplaces baselayer etc) in the calling function, but since GFI functions should never call internal layers (as most if not all are vector layers) I decided to add a new param for the PermissionHelper function and added an overloaded version with the previous signature on the class so everything calling it now is working like before (getting references to internal layers).

okauppinen:
Sorry, my mistake. Didn't notice that GFI uses false.

PR #1013, opened 2023-10-12
Myplaces use database directly

DenverCoder544:
-implement fetching features to map directly from db (bypasssing wfs client) for myPlaces
-make userLayerWfs / analysisWfs db-direct-ready
-able to work with a featurecollection that has varying geometry-types


jampukka:
Fantastic

PR #1015, opened 2023-10-23
Userlayer data db direct

DenverCoder544:
-fetch user layer data to map directly from db and lose the wfs-t dependency

### DIFF
--- a/service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
+++ b/service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
@@ -181,15 +181,4 @@ public class WKTHelper {
         log.debug("BBOX: " + bbox);
         return bbox;
     }
-
-    /**
-     * Get the numeric SRID of an EPSG-code
-     * @param srsName The String representation of the EPSG-code (e.g. EPSG:3857)
-     * @return the (int) id part of the string e.g. 3857
-     */
-     public static int getSRID(String srsName) {
-        String srid = srsName.substring(srsName.indexOf(':') + 1);
-        return Integer.parseInt(srid);
-    }
-
 }
--- a/service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
+++ b/service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
@@ -1,5 +1,9 @@
 package fi.nls.oskari.map.geometry;
 
+import fi.nls.oskari.log.LogFactory;
+import fi.nls.oskari.log.Logger;
+import org.geotools.geometry.jts.JTS;
+import org.geotools.referencing.CRS;
 import org.locationtech.jts.geom.CoordinateSequence;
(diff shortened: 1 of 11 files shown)


service-userlayer/src/main/resources/org/oskari/map/userlayer/service/UserLayerMapper.xml
```diff
@@ -19,6 +19,19 @@
...
+
+    <resultMap id="UserLayerDataResult" type="UserLayerData">
+        <result property="id" column="id" />
+        <result property="user_layer_id" column="user_layer_id" />
+        <result property="uuid" column="uuid" />
+        <result property="feature_id" column="feature_id" />
+        <result property="property_json" column="property_json" />
+        <result property="wkt" column="wkt" />
+        <result property="databaseSRID" column="srid"/>
+        <result property="created" column = "created" jdbcType="TIMESTAMP" javaType="java.time.OffsetDateTime" />
```
ZakarFin, line 31: Should this have updated as well?

service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
```diff
@@ -181,4 +181,15 @@ public static String getBBOX(final double westBoundLongitude, final double south
         log.debug("BBOX: " + bbox);
         return bbox;
     }
+
+    /**
+     * Get the numeric SRID of an EPSG-code
+     * @param srsName The String representation of the EPSG-code (e.g. EPSG:3857)
+     * @return the (int) id part of the string e.g. 3857
+     */
+     public static int getSRID(String srsName) {
```
ZakarFin, line 190: The ProjectionHelper has similar methods already: https://github.com/oskariorg/oskari-server/blob/master/service-map/src/main/java/fi/nls/oskari/map/geometry/ProjectionHelper.java#L133-L194

We should add this there instead so they are not spread out. Should we check for null param here as well? Maybe throw a ServiceRuntimeException telling that srs was not available instead of the NullPointerException that this would throw?

service-userlayer/src/main/resources/org/oskari/map/userlayer/service/UserLayerMapper.xml
DenverCoder544: yes.

service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
DenverCoder544: moved and added the runtime exception.

service-map/src/main/java/fi/nls/oskari/map/geometry/WKTHelper.java
```diff
@@ -181,4 +181,15 @@ public static String getBBOX(final double westBoundLongitude, final double south
...
+        String srid = srsName.substring(srsName.indexOf(':') + 1);
```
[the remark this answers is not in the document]
DenverCoder544: good point. changed.

PR #1016, opened 2023-10-24
Analysis db direct

DenverCoder544:


control-base/src/main/java/fi/nls/oskari/control/feature/AnalysisWFSHelper.java
```diff
@@ -172,10 +180,13 @@ private boolean isVisibleProperty(String name) {
...
-            Filter filter = this.getWFSFilter(layerId, bbox);
-            SimpleFeatureCollection sfc = wfsClient.getFeatures(layer, bbox, crs, filter);
-            return this.postProcess(sfc);
+            JSONObject featureCollectionJSON = analysisDbService.getFeatures(parseId(layerId), bbox, crs);
+            SimpleFeatureCollection featureCollection = featureCollectionJSON == null ?
+                new EmptyFeatureCollection(null) :
+                GeoJSONReader.toFeatureCollection(featureCollectionJSON);
+            return postProcess(featureCollection);
         } catch(Exception e) {
+            LOG.error("Failed to get features. ", e);
```
ZakarFin, line 189: Switch the params on this as Logger takes throwable as first param to print stack trace.
```suggestion
            LOG.error(e, "Failed to get features.");
```
We'll need to remove this or limit logging at some point but for now it's ok to both throw the exception and log. Usually it would be a bad idea since the error will get logger twice.

ZakarFin: Or actually. Lets log this on the wfsclient so we don't need to add this to all helpers.

ZakarFin: So we could do this in here:
https://github.com/oskariorg/oskari-server/blob/develop/service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariFeatureClient.java#L82

DenverCoder544: Moved the logging to oskarifeatureclient for all three implementations. We currently need to do the logging in featureclient, because later on, when the exception gets caught and rethrown the stack trace is lost by design (ActionCommonException). But I added ad disclaimer that we might wanna lose this debuglogging at a later time.

PR #1017, opened 2023-10-30
Add 'loose bbox' query for testing performance on user layers

ZakarFin:
Trying out queries on prod this gives a small performance boost from 80ms to 50ms on my dataset for the query. 

However we should take a better look at the code that is run after the query. It looks like we are:
1) querying for WKT
2) transforming the coordinates and writing out geojson features
3) building a GeoTools SimpleFeatureCollection from the geojson features
4) transforming the coordinates again
5) writing out geojson.

We could return the SimpleFeatureCollection from db service directly and reduce the amount of code that needs to run and handle the features before we write it out as a geojson string for the client.


ZakarFin:
On further studies on this. It is the db query that is taking up the time and in current server env using the "loose bbox" drops select time from ~1.2s to ~50ms.

PR #1025, opened 2023-11-14
userlayerdata: return simpleFeatureCollection from mybatis

DenverCoder544:
Simplify implementation and just return simplefeaturecollection from the results of db. Got rid of many unnecessary transformations (format, srs).

### DIFF
--- a/service-userlayer/src/main/java/org/oskari/map/userlayer/service/UserLayerDbServiceMybatisImpl.java
+++ b/service-userlayer/src/main/java/org/oskari/map/userlayer/service/UserLayerDbServiceMybatisImpl.java
@@ -318,7 +318,6 @@ public class UserLayerDbServiceMybatisImpl extends UserLayerDbService {
         SimpleFeatureType simpleFeatureType = featureTypeBuilder.buildFeatureType();
         SimpleFeatureBuilder featureBuilder = new SimpleFeatureBuilder(simpleFeatureType);
         featureBuilder.set(GEOM_ATTRIBUTE, geom);
-        featureBuilder.set("geometry_name", GEOM_ATTRIBUTE);
         featureBuilder.set("id", feature.getId());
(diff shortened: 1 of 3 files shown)


jampukka:
Could we get by with single FeatureType with nullable Geometry.class?

service-userlayer/src/main/java/org/oskari/map/userlayer/service/UserLayerDbServiceMybatisImpl.java
```diff
@@ -266,74 +277,107 @@ private UserLayerMapper getMapper(SqlSession session) {
...
-        } catch(JSONException ex) {
-            log.warn("Failed to convert UserLayerData to GeoJSONFeature");
+    private SimpleFeature toSimpleFeature(UserLayerData feature) {
+
+        Geometry geom = parseWKT(feature.getWkt());
+        SimpleFeatureTypeBuilder featureTypeBuilder = getFeatureTypeBuilder(geom);
+
+        SimpleFeatureType simpleFeatureType = featureTypeBuilder.buildFeatureType();
+        SimpleFeatureBuilder featureBuilder = new SimpleFeatureBuilder(simpleFeatureType);
+        featureBuilder.set(GEOM_ATTRIBUTE, parseWKT(feature.getWkt()));
```
jampukka, line 320: Don't parse WKT again, use the already parsed `geom` instead.

DenverCoder544: You're right about these. Unfortunately it seems that a generic featuretype with Geometry.class does not work. It results in an exception when trying to create the actual feature.

jampukka: Thanks for trying!

PR #1029, opened 2023-11-17
Sotkanet region sets mapping for Hyvinvointialue

ZakarFin:
This allows us to show data for Hyvinvointialueet from Sotkanet. However we make an educated guess about which regionsets _can be used_ for some of the indicators (mostly the ones that do have data for hyvinvointialueet). Sotkanet has a mechanism to tell which region sets they have data for:

- Hyvinvointialue: https://sotkanet.fi/rest/1.1/indicators/3815
 
Notice the missing/empty array of classifications.region.values
```
"classifications": {
    "region": {
        "title": {
            "fi": "Hyvinvointialue",
            "en": "The wellbeing services county",
            "sv": "Välfärdsområde"
        },
        "values": []
    }
    
}
```

- Non-hyvinvointialue indicator: https://sotkanet.fi/rest/1.1/indicators/900
```
"classifications": {
    "region": {
        "title": {
            "fi": "Kunta, seutukunta, maakunta, hyvinvointialue, sairaanhoitopiiri, yhteistyöalue, aluehallintoviraston alue, suuralue, Manner-Suomi/Ahvenanmaa, erityisvastuualue, koko maa",
            "en": "Municipality, sub-region, region, the wellbeing services county, hospital district, collaborative area, area for the regional state administrative agency, major region, university hospital special responsibility area, Mainland Finland/Åland, whole country",
            "sv": "Kommun, ekonomisk region, landskap, välfärdsområde, sjukvårdsdistrikt, samarbetsområde, området för regionförvaltningsverket, storområde, specialupptagningsområde, Fastlandsfinland/Åland, hela landet"
        },
        "values": [
            "Suuralue",
            "Maa",
            "Sairaanhoitopiiri",
            "Aluehallintovirasto",
            "Nuts1",
            "Kunta",
            "Erva",
            "Seutukunta",
            "Maakunta"
        ]
    }
}
```

We rely on that values array to filter out regionsets that can't be used with the indicator, but the ones with Hyvinvointialue data seem to have the empty array. First tried assuming no region values means there's data for all the region sets. However this doesn't appear to be very user-friendly as it looks like there's no data in these indicators for any other region set that isn't Hyvinvointialue.

Next added logic to detect region values for missing values from the region title:
```
    "region": {
        "title": {
            "fi": "Hyvinvointialue",
            ...
```
This makes more sense as this (case-insensitively) seems to be the same as the category for regions in https://sotkanet.fi/rest/1.1/regions. We use the category values to match the classification region values already.
```
{
    "id": 961,
    "code": "01",
    "category": "HYVINVOINTIALUE",
    "title": {
        "fi": "Itä-Uudenmaan hyvinvointialue",
        "en": "East Uusimaa wellbeing services county",
        "sv": "Östra Nylands välfärdsområde"
    }
}
```
Needed to change the logic where the indicator parameters/region sets are parsed. It was previously parsed from the response of /indicators listing, BUT for these Hyvinvointialue indicators the values are missing from that one. So changed all indicators to parse parameters from the /indicators/[id] response since that _should_ be the correct place for these (even if it's sometimes conflicting with the /indicators response).

### DIFF
--- a/service-statistics-sotka/src/main/java/fi/nls/oskari/control/statistics/plugins/sotka/parser/SotkaIndicatorParser.java
+++ b/service-statistics-sotka/src/main/java/fi/nls/oskari/control/statistics/plugins/sotka/parser/SotkaIndicatorParser.java
@@ -27,12 +27,7 @@ public class SotkaIndicatorParser {
 
     public StatisticalIndicator parse(JSONObject json, Map<String, Long> sotkaLayersToOskariLayers) {
         try {
-            StatisticalIndicator indicator = createIndicator(json, sotkaLayersToOskariLayers);
-            if(indicator == null) {
-                return null;
-            }
-            setupMetadata(indicator);
-            return indicator;
+            return createIndicator(json, sotkaLayersToOskariLayers);
         } catch (Exception e) {
             LOG.error("Error in mapping Sotka Indicators response to Oskari model: " + e.getMessage(), e);
         }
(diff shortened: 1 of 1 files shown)


ZakarFin:
Another problem arises with https://sotkanet.fi/rest/1.1/indicators/3803 having the classification.region.title.fi BUT the same indicator in https://sotkanet.fi/rest/1.1/indicators has the structure BUT doesn't have the value...

ZakarFin:
Also added parsing for additional description for indicators ("interpretation", "limits", "legislation" parts are appended to "description"):

image

Before:
image

PR #1045, opened 2024-02-09
Remove geoserver

DenverCoder544:
-removed setup module
-removed some myplaces / mydata / analysis related stuff that were used by setupmodule creating said layers to geoserver
-can now survive without geoserver url/username/pw in props


ZakarFin:
GeoServer is no longer required for user generated content. This PR removes code that was used to initialize config for included GeoServer.

PR #1058, opened 2024-03-18
Make sure we don't write duplicated indicators

ZakarFin:
When processing indicators from a statistical datasource. It breaks the UI completely when the user scrolls to such indicator:

aineistohaku

This might happen when we have multiple servers that are processing the indicator list at the same time. One server might save the indicator as processed while another still has it's on its work queue. When the other server saves it gets the processed one from redis and adds a duplicate from its work queue.

https://github.com/ant-design/ant-design/issues/33864


ZakarFin:
Maybe we should just enable restricting the update to be done on a single node on clustered env. Nothing here really addresses the race condition issue where two nodes get a common list of saved indicators from Redis, but might add different sets when saving the changes at the same time (overwriting the other nodes processed data). This might lead to some indicators not being listed after the processing has been done.

ZakarFin:
Disabling this might work with just adding this empty value prop: `oskari.scheduler.job.StatisticsDatasources=`. The scheduler will log a warning with "Available scheduled job StatisticsDatasources needs the cronLine configuration parameter" but it should not run the update when the line is empty.

ZakarFin:
Documenting here that configuring `oskari.scheduler.job.StatisticsDatasources=` did not disable the scheduler to run the update.

PR #1067, opened 2024-08-26
Refactor permission queries to joins

ZakarFin:
On local dev with some 3000 layers. Loading layers from db (with heavy logging so it's not properly measured):

Before:
- Amount of select queries for permissions: 6097
- Time from loading layers: 11,6s

After:
- Amount of select queries for permissions: 15
- Time from loading layers: 3,9s

Time measured from log entries between:
- from "GetAppSetupHandler - View ID: 1 created by user -1 is public, access granted for user with id -1 " 
- to "OskariLayerWorker - Returning 2712 / 2928 layers "
But with all the sql logging happening which I'm sure adds a lot to it.


jampukka:
Looks good

PR #1072, opened 2024-10-30
layerspecific additional whitlisted urlparams not found in caps

DenverCoder544:
Some layers require apikeys or referers or whatnot to function, and those did not get added in printservice.

This solution is based on paikkis and one provider specifically, so the whitelist probably should reside somewhere else than hardcoded here and in paikkis' case we might also consider a few other whitelisted params as well.

### DIFF
--- a/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
+++ b/service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
@@ -1,31 +1,35 @@
 package org.oskari.print.loader;
 
-import java.awt.Graphics2D;
-import java.awt.image.BufferedImage;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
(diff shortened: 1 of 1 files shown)


service-print/src/main/java/org/oskari/print/loader/CommandLoadImageWMTS.java
```diff
@@ -201,6 +211,43 @@ private LayerCapabilitiesWMTS getLayerCapabilities() throws IllegalArgumentExcep
         throw new IllegalArgumentException("Could not find layer from Capabilities");
     }
 
+    private String getLayerAdditionalRequestParams() {
+        String[] layerUrlParams = layer.getOskariLayer().getUrl().split("\\?");
```
ZakarFin, line 215: OskariLayer has getParams(). We should have the apikey there instead of parsing it from the URL. Also IOHelper has bunch of methods parsing and constructing URLs.

PR #1081, opened 2024-11-13
WMS time dimension and interval parsing

ZakarFin:
The frontend assumes to get the time series as an array of times. This changes the time series array value that present an interval to a list of times for the capabilities JSON.


jampukka:
See https://docs.oracle.com/javase/9/docs/api/java/time/Period.html#parse-java.lang.CharSequence-

PR #1083, opened 2024-11-13
Remove restricted params from urls

ZakarFin:
If a WFS-layer is registered to the application with an URL containing request, version or service parameters some things will break for example the capabilities parser that tries to call request=DescribeFeatureType. This results in calls having something like request=GetCapabilities&request=DescribeFeatureType when fetching the featuretype documentation.

This is first attempt with what I consider the "right places" for doing this:
- MapLayer is used when inserting layers from JSON at the startup and when admin UI is used.
- ServiceConnectInfo is used for fetching capabilities and handles the issue when there are layers in the database that has the problematic parameters on the URL.

However, we might need to support resources:// protocol for urls (used in thematic maps) and regenerating the url is pretty nasty looking at this point (generating arrays and strings for each param).

### DIFF
--- a/service-base/src/test/java/org/oskari/util/LayerUrlHelperTest.java
+++ b/service-base/src/test/java/org/oskari/util/LayerUrlHelperTest.java
@@ -1,0 +1,15 @@
+package org.oskari.util;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class LayerUrlHelperTest {
+
+    @Test
+    public void getSanitizedUrl() {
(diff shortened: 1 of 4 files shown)


service-base/src/test/java/org/oskari/util/LayerUrlHelperTest.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class LayerUrlHelperTest {
+
+    @Test
+    public void getSanitizedUrl() {
+        assertEquals("https://jee.org/?apikey=woot&proxy=not", LayerUrlHelper.getSanitizedUrl("https://jee.org/?apikey=woot&request=GetWheel&proxy=not"));
```
jampukka, line 11: GetWheel 🙊

PR #1085, opened 2024-11-22
Support mixed geometry types when reading region sets from resources / statistical data

ZakarFin:
Use the same logic to read GeoJSON in WFSClient and RegionSetHelper to workaround this: https://gis.stackexchange.com/questions/316134/reading-a-geojson-via-java-geotools-api-issue

When GeoTools is used to read the feature collection it will use the geometry type from the first feature. This results in region sets that have both Polygon and Multipolygon features look like this on the map (Multipolygons forced to Polygons if the first feature has a Polygon geometry):
image

Note that there's already a GeoJSONReader on geotools-ext Maven module, but it has very little dependencies. To keep it that way I added the GeoJSONStringReader to service-map that already has Jackson-databind as dependency.


ZakarFin:
After running with this code:
image

PR #1089, opened 2024-12-10
Java 11/17 produces slightly different coordinates than Java 8

ZakarFin:
Remove Java version detection from test and updated expected coordinates.


jampukka:
Very slight differences, it seems to round to the nearest 14th decimal point? That's pretty small as 7th decimal point corresponds to 1 cm.

PR #1094, opened 2025-01-07
Remove axiom from transforming GFI responses

ZakarFin:
Part of https://github.com/oskariorg/oskari-documentation/issues/70


ZakarFin:
Should be ok, but still searching for a test case for this.

PR #1098, opened 2025-01-09
Remove deprecated proxying code related to GeoServer/user generated content

ZakarFin:
Related to: https://github.com/oskariorg/sample-configs/pull/29

The PR is a bit larger than I hoped but any of the code is no longer by Oskari. If someone has built something app specific using this, it should be pretty straight forward to include in the app itself.


ZakarFin:

PR #1099, opened 2025-01-09
Remove unused code - SLDStore

ZakarFin:
Continuing (and including commits from) #1098 with removal of SLDStore and related resource files. We haven't supported rendering raster images with SLD for some years so this was unused as well.


ZakarFin:

PR #1100, opened 2025-01-09
Remove code related to creating new analysislayers

ZakarFin:
Continuing  #1098 and #1099.

Removal for code that relates to creating new analyses. The implementation has been deemed too fiddly to work with. We are considering creating more simple and generic analysis functionalities later on.


ZakarFin:

PR #1101, opened 2025-01-09
Remove WFST code related to myplaces

ZakarFin:
Most of the code relates to myplaces being saved using WFST by calling the "internal" GeoServer. This has since been migrated to direct db calls and the code can be removed as unused.

OskariLayoutWorker might have had something to do with how statistical data table was added to printouts, but it hasn't been updated to work with the current printout implementation.


ZakarFin:

PR #1107, opened 2025-01-15
Remove service-statistics-eurostat for statistics functionality

ZakarFin:
The plugin never went to production and the APIs have changed on eurostat so it doesn't really make sense to keep the code. It's easy enough to add back if we ever need to use these.


ZakarFin:

PR #1108, opened 2025-01-15
Upgrade PDFBox

osqq:
Upgrade PDFBox 2.0.32 to 3.0.3 and fix migration issues


service-print/pom.xml
```diff
@@ -37,6 +37,7 @@
         <dependency>
             <groupId>org.apache.pdfbox</groupId>
             <artifactId>pdfbox</artifactId>
+            <version>${pdfbox.version}</version>
```
ZakarFin, line 40: Version should be removed since we have pdfbox in parent-pom dependencyManagement.
```suggestion
```

PR #1109, opened 2025-01-17
Remove axiom

okauppinen:
Remove axiom, fi.nls.oskari.XmlHelper and use org.oskari XmlHelper

`Parsing with xpaths don't work with the "future" parser so we can't use it yet`
=> didn't find problems with xpath but setting sax/features/namespaces to false for nsAware builder caused that ns/uri wasn't parsed and CSWService got null for responseElement.getNamespaceURI

`CSWISORecordNamespaceContext.GMDNS.equals(responseElement.getNamespaceURI()` => Invalid response exception

Without axiom-impl some control-base tests fails. Removing:
`@PowerMockIgnore({"com.sun.org.apache.xalan.*", "com.sun.org.apache.xerces.*", "javax.xml.*", "org.w3c.dom.*", "org.xml.*", "com.sun.org.apache.xml.*"})` 
will fix tests. 

For now left axiom-impl in pom.xml and used it in control-base with comments because junit will be udated also and there will be changes  for powermock usage. I will remove axiom-impl after junit update.


control-base/pom.xml
```diff
@@ -90,5 +90,10 @@
             <artifactId>gt-epsg-hsql</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <!-- some powermock tests ignores xml packages and tests fails without this -->
+            <groupId>org.apache.ws.commons.axiom</groupId>
+            <artifactId>axiom-impl</artifactId>
```
ZakarFin, line 96: If it only affects tests we could use <scope>test</scope> here?

ZakarFin:
```suggestion
            <artifactId>axiom-impl</artifactId>
            <scope>test</scope>
```

PR #1111, opened 2025-01-20
Ignore initializing wfs search channels during tests

ZakarFin:
Prevent unncessary noise (SQL errors) when running unit tests. GetAppSetupHandler tests initialize all ViewModifiers including `SearchHandler`. SearchHandler initializes SearchService which tries to find all  ChannelProviders that inject new SearchChannels as available channels. One (and only) ChannelProvider that we currently have is the `WFSChannelProvider` that tries to query the database for additional search channels. As there is no database nor tables initialized for these unit tests, it triggers a couple of screens worth of stacktrace about an SQL error and there not being a database table to query.

So this is a workaround to prevent that unnecessary noise. Could be used in production but don't really care to document it as it's doesn't really make sense to use this configuration.


ZakarFin:
Solves the rest of https://github.com/oskariorg/oskari-documentation/issues/76

ZakarFin:
Maybe this should have been something like `search.providers.ignored=WFSChannelProvider` with a list as value that would have been handled by SearchService instead.

PR #1123, opened 2025-01-29
migrate infoplugin colourScheme to theme->map->infobox and remove red…

DenverCoder544:
…undant keys

this will move the old way of configuring infobox colours on published maps ( colouScheme: { ... } } as part of GetInfoPlugin's config to infobox-section under theme->map.

Frontend changes for infobox using coloring from theme was implemented in
https://github.com/oskariorg/oskari-frontend/pull/2777

### DIFF
--- a/content-resources/src/main/java/flyway/oskari/V2_15_1__infobox_colourscheme_to_theme.java
+++ b/content-resources/src/main/java/flyway/oskari/V2_15_1__infobox_colourscheme_to_theme.java
@@ -1,0 +1,157 @@
+package flyway.oskari;
+
+import fi.nls.oskari.log.LogFactory;
+import fi.nls.oskari.log.Logger;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+public class V2_15_1__infobox_colourscheme_to_theme  extends BaseJavaMigration {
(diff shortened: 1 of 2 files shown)


content-resources/src/main/java/flyway/oskari/V2_15_1__infobox_colourscheme_to_theme.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private JSONArray updateColourSchemeToTheme(JSONArray mapModules) throws JSONException {
+        JSONArray returnValue = new JSONArray();
+
+        for (int i = 0; i < mapModules.length(); i++) {
+            JSONObject module = mapModules.getJSONObject(i);
+            JSONObject config = new JSONObject(module.getString("config"));
+            JSONObject metadata = new JSONObject(module.getString("metadata"));
+
+            if(metadata == null || config == null) {
```
ZakarFin, line 98: Pretty sure metadata nor config are null at this point. They are either an empty JSONObject or JSONObject(module.getString("config")) throws an exception (which is bad since one bundle without a config or metadata will break the whole loop). Maybe check before giving the string to JSONObject. 

We could also just create a local class instead of wrapping the whole stuff on JSONObject and JSONArrays. The arrays are bit of a pain to loop through.

ZakarFin: Mildly concerned there's some instance with null configs/metadata out there that will break. Otherwise this looks good to me.

DenverCoder544: Fixed the nullchecks to actually make sense. In retrospect, a local class would've been a cleaner solution than fiddling with the nested hierarchies of JSONObjects / JSONArrays which _are_ truly a pain...

PR #1135, opened 2025-02-26
Replace Hystrix with Resilience4j

okauppinen:
Refactor print and wfs  Hystrix commands to Resilience4j loaders.

Resilience4j throws exceptions:
- CircuitBreaker: Open/Half-Open => CallNotPermittedException
- TimeLimiter:  timeout => TimeoutException
- Bulkhead: full => BulkheadFullException
In command chain exception is wrapped with CompletionException => getCause to inspect actual exception

Print uses fallback so Resilience4j doesn't throw exceptions for print loader.

Resilience4j config:
- https://resilience4j.readme.io/docs/circuitbreaker
- https://resilience4j.readme.io/docs/bulkhead
- https://resilience4j.readme.io/docs/timeout
- https://resilience4j.readme.io/docs/retry

Hystrix:
- https://github.com/netflix/hystrix/wiki/configuration


control-admin/src/main/java/fi/nls/oskari/control/admin/SearchWFSChannelActionHandler.java
```diff
@@ -71,7 +72,7 @@ public void handleGet(ActionParameters params)
                 }
             }
         } catch (Exception ex){
-            log.error(org.apache.commons.lang.exception.ExceptionUtils.getStackTrace(ex));
+            log.error(ExceptionUtils.getStackTrace(ex));
             throw new ActionParamsException("Couldn't get WFS search channels");
```
ZakarFin, line 76: Just add the cause exception instead? I know only import changed here, but I don't think we need the ExceptionUtils here?
```suggestion
            throw new ActionParamsException("Couldn't get WFS search channels", ex);
```

PR #1137, opened 2025-03-05
Digitransit v2

DenverCoder544:
Should be considered work in progress (for instance unit tests for service impl are not yet updated).

Read from v2 api and map to v1 format response to make it appear as if nothing's changed. Although some things have changed.
Request parameters:
-showIntermediateStops & maxWalkDistance in the request do not do anything anymore (=they do not have a place in the new planConnection - query).

In the response types:
-itinerary:
* transittime is calculated by us and is not readily available in the api
* walkLimitExceeded & tooSloped are not available in the new api
-Leg:
* property "pathway" could not be found
* property "route" -> equivalent not found in the new api -> using route.shortName here

Stop:
* stopIndex & stop sequence not (yet) available. I'll still have to take a looksie at Stop.stopPosition and see if stuff found under that can be made match these.

### DIFF
--- a/service-routing/src/test/resources/fi/nls/oskari/routing/digitransit-response-error-v2.json
+++ b/service-routing/src/test/resources/fi/nls/oskari/routing/digitransit-response-error-v2.json
@@ -1,0 +1,28 @@
+{
+  "errors": [
+    {
+      "message": "Validation error (SubselectionRequired@[planConnection/edges/node/legs/start]) : Subselection required for type 'LegTime!' of field 'start'",
+      "locations": [
+        {
+          "line": 48,
+          "column": 13
+        }
+      ],
+      "extensions": {
+        "classification": "ValidationError"
+      }
+    },
+    {
+      "message": "Validation error (SubselectionRequired@[planConnection/edges/node/legs/end]) : Subselection required for type 'LegTime!' of field 'end'",
+      "locations": [
+        {
+          "line": 49,
+          "column": 13
+        }
+      ],
+      "extensions": {
+        "classification": "ValidationError"
+      }
+    }
+  ]
+}
--- a/service-routing/src/test/resources/fi/nls/oskari/routing/digitransit-response-success-v2.json
+++ b/service-routing/src/test/resources/fi/nls/oskari/routing/digitransit-response-success-v2.json
@@ -1,0 +1,11717 @@
+{
+    "data": {
+        "planConnection": {
+            "searchDateTime": "2025-03-07T12:29:00+02:00",
+            "edges": [
+                {
+                    "node": {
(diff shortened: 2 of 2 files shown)


service-routing/src/main/java/fi/nls/oskari/routing/RouteParser.java
```diff
@@ -170,12 +194,15 @@ public JSONObject generatePlan(Route route, RouteParams params){
...
      * @param params
      * @return from JSON
      */
-    private JSONObject getFromJSON(Plan plan, RouteParams params){
-        final From from = plan.getFrom();
+    private JSONObject getFromJSON(PlanConnection planConnection, RouteParams params){
+        // first node -> first leg -> from
+        Edge firstEdge = planConnection.getEdges().get(0);
+        fi.nls.oskari.routing.pojo.Leg lastLeg = firstEdge.getNode().getLegs().get(0);
+        final Place from = lastLeg.getFrom();
```
ZakarFin, line 205: lastLeg -> firstLeg?

service-routing/src/main/java/fi/nls/oskari/routing/RouteParser.java
```diff
@@ -285,189 +283,319 @@ public JSONObject getItinerariesGeoJSON(Itinerary itinerary, RouteParams params)
...
-                    if(stepJSON.has(PARAM_LEGS_STEPS_LON) && stepJSON.has(PARAM_LEGS_STEPS_LAT)){
-                        String stepLon = stepJSON.getString(PARAM_LEGS_STEPS_LON);
-                        String stepLat = stepJSON.getString(PARAM_LEGS_STEPS_LAT);
+                legsJSON.put(legJSON);
+            }
+        } catch (JSONException ex){
+            LOG.error("Cannot get itineraries legs JSON", ex);
+        } catch(Exception e) {
+            // TODO: remove this before going to develop -> keep for ease of debugging.
+            LOG.error("LegsJSON major fail.", e);
```
ZakarFin, line 375: Do we want to remove this before merging?

ZakarFin:
We'll continue working on making the code more readable, but it's kinda time critical to get this available for testing so lets go with this version for now.

In hindsight it might have been easier to test if we had kept the old one and created a new V2Impl instead of updating the old one. Could generated some expected results from v1 and checked that the new version produces similar enough output/document differences on unit tests.

service-routing/src/main/java/fi/nls/oskari/routing/RouteParser.java
ZakarFin: I'll merge it for testing. We can remove it later.

PR #1147, opened 2025-03-13
Add config property to configure flyway migration naming validation

tuomari:


tuomari:
The default in flyway is false, ie the current functionality is not changed by this MR. 
It could be argued that default should be set to true. It is extremely annoying to debug why stuff is failing, only to finally figure out the problem is the migration starting with lowercase `v` instead of an uppercase one..

tuomari:
If the validation is off, the migration with invalid file formatting is not executed, but the system will start.
If the validation is on, the system will fail starting.

IMO it is better to fail visibly and hard with a valid error message than continue running, with one or more failed migrations that might cause some strange behaviour.

The current error message flyway gives:
```
org.flywaydb.core.internal.resource.ResourceNameValidator - 1 SQL migrations were detected but not run because they did not follow the filename convention.
org.flywaydb.core.internal.resource.ResourceNameValidator - Set 'validateMigrationNaming' to true to fail fast and see a list of the invalid file names.
```

ZakarFin:
On that I totally agree. Lets change the default to true so the developer gets a message about attempted migration that would be skipped and merge this.

tuomari:
Modified.

ZakarFin:
Thanks!

PR #1163, opened 2025-04-07
only log error message for given content types

DenverCoder544:
only log error message for whitelisted content types & limit number of chars to log


ZakarFin:
Cherry-picked 117dde64fe06c7e873aa85b3a048bb9847c31e56 and 62a822cb77fd606b27eef554cb69192ce05da049 to hotfix/3.0.2

PR #1165, opened 2025-04-08
Try reconnecting to Redis for cluster comms

ZakarFin:
Previously if the Redis connection was broken (redis-server restarted etc) the cluster communications did not reconnect. With this change the code tries to reconnect when a new message would be sent and the connection has been broken. It's not perfect but it's better. Otherwise we would need some sort of retry after a given time and gradual backoff etc.


ZakarFin:
Cherry-picked ebbf181b9f414f217b1bc96eac4e2d39c3edfdc6 to hotfix/3.0.2

PR #1167, opened 2025-04-16
Force targetCRS to always have longitude first on GetWfFeaturesHandler

tuomari:
Oskari seems to internally always handle longitude first, but GetWfsFeatureHandler tries to read the coordinate order from the actual coordinatesystem. This causes lat-lon coordinate systems ( for example ETRS:3879 ) to fail while fetching user defined layers.


ZakarFin:
Uh, coordinate order stuff I'm not keen changing without serious testing, but lets merge it to develop so there will be plenty of time to test it before it goes to a release.

PR #1168, opened 2025-04-17
Remove duplicate deploy

tuomari:
When running mvn deploy, the deploy phase gets executed twice, which causes problems for example with github repositories where artifacts are immutable, ie. the second deploy fails.

If artifacts are deployed with mvn deploy command, there should not be any functional changed. Both sources and javadocs get created, they are just deployed with all other artifacts, instead of built and sent in a separate phase


ZakarFin:
Cool, changed the target to develop so we don't mess up the upcoming hotfix if there's something that doesn't work in our usual process with this change.

PR #1177, opened 2025-06-27
Bump geotools 32.3 -> 33.1

jampukka:
There's some edge case that upsets an assert statement in GeoTools 33.x versions. The issue is related to forcing longitude on both sourceCrs and targetCrs and the changes of the datum on which Finnish national projections rely (and the EPSG registry update done in GeoTools 33). When creating the operation step from EPSG:4258 to EPSG:10690 (which is NOP/identity matrix as per self) the two forced axis order swaps cancel out each other and [this assert](https://github.com/geotools/geotools/blob/0ae99132d94cf5bd80aff98aecbf2868fb11c03e/modules/library/referencing/src/main/java/org/geotools/referencing/operation/AuthorityBackedFactory.java#L516) fails. With assertations disabled the code still works fine and the test passes. I'll try to formulate some sort of bug report to GeoTools about this, but the topic is quite advanced.

I expect no one to run their Oskari installations with assertations enabled (-ea JVM argument), based on the fact that (my) top Google result for "java enable assertations" is a Stack overflow thread asking "How to enable assertations?". As such this shouldn't cause any issues when actually running Oskari and we should be fine with moving to GeoTools 33.x.


jampukka:
Debugged this a bit further. EUREF-FIN (EPSG:10690) replaced ETRS89 (EPSG:4258) as the Geodetic CRS of most Finnish projections with the recent EPSG database update. When combined with longitudeFirst=true the dual xy-swaps from `EPSG:4258, true` to `EPSG:10690, true` cancel each other out, resulting in the assertation failure.

I could not figure out how to actually disable Java asserts (tried to disable it for `AuthorityBackedFactory` at runtime which worked when running a single test, tried setting enableAssertations in maven-surefire-plugin configurations to false) so I decided it's just easier if I alter the test so that it passes. The point about this not causing issues when actually running Oskari still stands.

PR #1191, opened 2025-09-29
Journey towards MyFeatures continues

jampukka:


service-myfeatures/src/main/java/org/oskari/map/myfeatures/service/MyFeaturesServiceMybatisImpl.java
```diff
@@ -236,6 +238,19 @@ public void swapAxisOrder(UUID layerId) {
...
+    public void deleteFeaturesByLayerId(UUID layerId) {
+        layerId = Objects.requireNonNull(layerId);
+        try (SqlSession session = factory.openSession()) {
+            MyFeaturesMapper mapper = getMapper(session);
+            mapper.deleteFeaturesByLayerId(layerId);
+            mapper.refreshLayerMetadata(layerId);
+            session.commit();
+        }
+
+        throw new UnsupportedOperationException("Unimplemented method 'deleteAllFeatures'");
```
ZakarFin, line 251: This just always throws an exception after commiting the single layer delete?

PR #1225, opened 2025-11-27
Add LayerProvider concept

jampukka:
Add new concept: `LayerProvider`
- `LayerProvider` provides information about available Layers in unifrom way
- Add two implementations of `LayerProvider`
  - `OskariMapLayerProvider` provides information about "default" layers `~OskariLayer`
  - `MyFeaturesLayerProvider` provides information about "MyFeatures" layers
- Refactor LayerList and DescribeLayer to use available `LayerProvider`s.


ZakarFin:
This would allow us to get rid of OskariLayerWorker and LayerJSONFormatters etc. Also it would allow an easy'ish extension point for layers that come from an external source.

ZakarFin:
Let's continue with this one on Oskari 4.0+ when we are removing myplaces/userlayers (replaced by myfeatures).

PR #1227, opened 2025-12-09
Myfeatures content editor

DenverCoder544:
minor mods to make front play with add/modify/delete

### DIFF
--- a/control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesWFSHelper.java
+++ b/control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesWFSHelper.java
@@ -48,6 +48,8 @@ public final class MyFeaturesWFSHelper extends UserLayerService {
     public static final String CREATED_PROP_NAME = "created";
     public static final String UPDATED_PROP_NAME = "updated";
 
+    public static final String FID_PROP_NAME = "fid";
+
     private MyFeaturesService service;
 
     private MyFeaturesService getService() {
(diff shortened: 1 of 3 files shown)


control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesWFSHelper.java
```diff
@@ -113,6 +116,7 @@ private static SimpleFeatureType createType(List<MyFeaturesFieldInfo> fields) {
         b.add(GEOM_PROP_NAME, Geometry.class);
         b.add(CREATED_PROP_NAME, OffsetDateTime.class);
         b.add(UPDATED_PROP_NAME, OffsetDateTime.class);
+        b.add(FID_PROP_NAME, String.class);
```
ZakarFin, line 119: This adds the id as feature property?

At some point we had something like called oskariId, oId, oid, OID or similar that was attached to features as internal id, but couldn't find it right now. Not sure if it's something that is still being used or if it has been removed from the codebase. I think we should use the same one here if we add the feature id as one of the feature properties. Will look into it more thoroughly.

jampukka: This is good. `fid` in MyFeatures is the original `id` when imported from a file. But it might be missing and/or generated when exported (`FeatureType.1` etc). `id` should be the Oskari managed internal/technical id of type `long` for both simplicity and performance reasons.

jampukka:
Looks good to me. I'll rework the layerId parsing in `MyFeaturesFeature` POST/PUT actions in a later PR.

PR #1232, opened 2026-01-21
style type change

DenverCoder544:
fix parsing of myfeatures layer style when creating new layer from scratch

### DIFF
--- a/control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesLayerHandler.java
+++ b/control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesLayerHandler.java
@@ -96,8 +96,7 @@ public class MyFeaturesLayerHandler extends RestActionHandler {
 
         JSONObject jsonStyle;
         try {
-            ObjectMapper jsonMapper = new ObjectMapper();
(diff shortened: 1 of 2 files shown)


control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesLayerHandler.java
```diff
@@ -94,7 +93,16 @@ public void handlePost(ActionParameters params) throws ActionException {
         layer.setOwnerUuid(params.getUser().getUuid());
         layer.setLayerFields(createLayer.getLayerFields());
         layer.setLocale(new JSONObject(createLayer.getLocale()));
-        layer.getLayerOptions().setDefaultFeatureStyle(createLayer.getStyle());
+
+        JSONObject jsonStyle;
+        try {
+            ObjectMapper jsonMapper = new ObjectMapper();
```
jampukka: Use existing ObjectMapper (`om`) instead of creating new one for each request.

jampukka:
Other than what I commented Looks good to me

control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesLayerHandler.java
DenverCoder544: good catch 👌

PR #1233, opened 2026-01-27
Update Layer extent/coverage from CSW metadata when saving layer

jampukka:
**Add support for LayerAdminHandler to upsert the metadata information when saving the layer.** Currently layer metadata is updated only via `CSWCoverageUpdateService` so there's a large delay after saving the layer before the information is updated. In addition there's also no immediate feedback available if there's something wrong with fetching the metadata and/or the extent based on the metadata (maybe the metadata id was wrong?).

**Add new class `CSW` that combines shared logic regarding CSW handling**. GetCSWDataHandler added support for overriding the CSW service base URL via layer attributes, but CSWCoverageUpdateService did not support this functionality. Additionally, GetCSWDataHandler also supported metadataid to exist in layers capabilities whereas CSWCoverageUpdateService ignored this as well.


ZakarFin:
Changes look good to me, just need to fix the test to workaround not having a datasource to operate on.

PR #1235, opened 2026-02-05
Spring 7

DenverCoder544:
spring 7.n.n update

This has not yet been tested with an actual instance of redis, thus the draft.

changes to apps overriding stuff might be necessary too, here's what changed in pti:
https://github.com/nls-oskari/kartta.paikkatietoikkuna.fi/pull/250


ZakarFin:
This is targeted for Oskari 4.0 as we need to upgrade Tomcat packaging to 11 and possibly update Jackson to 3.x as well.

PR #1237, opened 2026-02-12
Add support for GeoJSON imports

jampukka:
Add support for importing GeoJSON (.json or .geojson) files to userlayer/myfeatures.

GeoJSON files are a bit different compared to other currently supported files which either support WGS84 only or have a standard way of extracting the CRS information.

If user provided CRS is not available we try to read the obsolete Named CRS object from the GeoJSON file (we write Named CRS object when exporting `myplaces`).


service-userlayer/src/main/java/org/oskari/map/userlayer/input/FeatureCollectionParsers.java
```diff
@@ -51,6 +54,7 @@ public static FeatureCollectionParser getByFileExt(String fileExt) {
         case MIFParser.SUFFIX: return new MIFParser();
         case SHPParser.SUFFIX: return new SHPParser();
         case GPKGParser.SUFFIX: return new GPKGParser();
+        case GeoJSONParser.SUFFIX: return new GeoJSONParser();
```
ZakarFin, line 60: Should we add `.geojson` as well?

jampukka: Sure, I'm not that happy with the whole "detection by file extension" decision but I'll add .geojson.

jampukka:
Added support for .geojson

PR #1238, opened 2026-02-16
handle attributes in myfeatures

DenverCoder544:
save & restore attributes to & fro myfeatures

### DIFF
--- a/control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesWFSHelper.java
+++ b/control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesWFSHelper.java
@@ -1,6 +1,7 @@
 package org.oskari.control.myfeatures;
 
 import fi.nls.oskari.annotation.Oskari;
+import org.json.JSONArray;
 import org.oskari.user.User;
 
 import fi.nls.oskari.domain.map.OskariLayer;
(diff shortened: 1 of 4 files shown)


control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesWFSHelper.java
```diff
@@ -218,19 +221,60 @@ private List<FeatureProperties> getProperties(MyFeaturesLayer layer, String lang
...
             p.type = field.getType().getSimpleType();
             p.rawType = field.getType().getOutputBinding().getName();
-            p.label = field.getName();
-            p.hidden = false;
-            p.format = null;
-            p.order = i++;
+            p.label = getLocalizedName(field, attributes, lang);
+            p.hidden = getHidden(field, attributes);
+            p.format = getFieldFormat(field, attributes);
+            p.order = getFieldOrder(field, attributes, i);
```
ZakarFin, line 234: I guess the previous ones were just placeholders for these. I thought that the frontend should do all the handling for these, but since they already have a placeholder it's good to have the correct value in it 🤷 Not sure if they could be removed though?

DenverCoder544: This is the same way it's done for wfs layers. I don't think the front is interested or involved in parsing these at all. https://github.com/oskariorg/oskari-server/blob/develop/control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java#L157

ZakarFin: Sorry, I thought this was modifying all the features going through the server, but it's just for the DescribeLayer call. This is exactly right.

PR #1239, opened 2026-02-19
Myfeatures export geojson

DenverCoder544:
geojson export + add srid for inserted features

### DIFF
--- a/control-myfeatures/src/main/java/org/oskari/control/myfeatures/ExportMyFeaturesHandler.java
+++ b/control-myfeatures/src/main/java/org/oskari/control/myfeatures/ExportMyFeaturesHandler.java
@@ -3,26 +3,26 @@ package org.oskari.control.myfeatures;
 import fi.nls.oskari.annotation.OskariActionRoute;
 import fi.nls.oskari.control.ActionException;
 import fi.nls.oskari.control.ActionParameters;
-import fi.nls.oskari.domain.map.myfeatures.MyFeaturesFeature;
+import fi.nls.oskari.control.RestActionHandler;
 import fi.nls.oskari.domain.map.myfeatures.MyFeaturesLayer;
 import fi.nls.oskari.log.LogFactory;
 import fi.nls.oskari.log.Logger;
-import fi.nls.oskari.service.ServiceException;
+import fi.nls.oskari.service.OskariComponentManager;
 import fi.nls.oskari.util.PropertyUtil;
 import jakarta.servlet.http.HttpServletResponse;
 import org.json.JSONObject;
+import org.oskari.map.myfeatures.service.MyFeaturesService;
 import org.oskari.user.User;
-
 import java.io.BufferedWriter;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
-import java.util.List;
+import java.util.Objects;
 import java.util.UUID;
 
 @OskariActionRoute("ExportMyFeaturesLayer")
-public class ExportMyFeaturesHandler extends MyFeaturesFeatureHandler {
+public class ExportMyFeaturesHandler extends RestActionHandler {
     private final static Logger LOG = LogFactory.getLogger(ExportMyFeaturesHandler.class);
     private static final String PARAM_SRS = "srs";
     private static final String PARAM_LAYER_ID = "layerId";
@@ -31,6 +31,18 @@ public class ExportMyFeaturesHandler extends MyFeaturesFeatureHandler {
     private static final String FILE_TYPE = "application/json";
 
     private static final DateTimeFormatter TIME_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+    private MyFeaturesService service;
+
+    @Override
+    public void init() {
+        if (service == null) {
+            setService(OskariComponentManager.getComponentOfType(MyFeaturesService.class));
+        }
+    }
+
+    void setService(MyFeaturesService myFeaturesService) {
+        this.service = Objects.requireNonNull(myFeaturesService);
+    }
 
     @Override
     public void handleGet(ActionParameters params) throws ActionException {
@@ -62,6 +74,11 @@ public class ExportMyFeaturesHandler extends MyFeaturesFeatureHandler {
             LOG.warn(e);
             throw new ActionException("Failed to export features");
         }
+    }
+
+
+    protected MyFeaturesService getService() {
+        return service;
     }
 
 }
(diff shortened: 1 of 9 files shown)


control-myfeatures/src/main/java/org/oskari/control/myfeatures/ImportMyFeaturesHandler.java
```diff
@@ -143,14 +145,14 @@ public void handlePost(ActionParameters params) throws ActionException {
             myFeaturesMaxFileSize = PropertyUtil.getOptional(PROPERTY_MYFEATURES_MAX_FILE_SIZE_MB, 10) * MB;
             unzippiedFileSizeLimit = 15 * myFeaturesMaxFileSize; // Max size of unzipped data, 15 * the zip size
         }
-
-        String sourceEPSG = params.getHttpParam(PARAM_SOURCE_EPSG_KEY);
+        String sourceEPSG = PropertyUtil.get(NATIVE_SRS, FALLBACK_NATIVE_SRS);
```
jampukka: Not sure about this change. sourceEPSG should be null (default case) or provided by the user (override). FeatureCollectionParsers will detect sourceESPG from the underlying file when possible, except if non-null sourceEPSG is provided ("trust me bro, just use this instead").

control-myfeatures/src/main/java/org/oskari/control/myfeatures/ExportMyFeaturesHandler.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.BufferedWriter;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.List;
+import java.util.UUID;
+
+@OskariActionRoute("ExportMyFeaturesLayer")
+public class ExportMyFeaturesHandler extends MyFeaturesFeatureHandler {
```
jampukka: I like the logic of extending MyFeaturesFeatureHandler, but this exposes the inherited handlePost, handlePut, handleDelete methods when it probably shouldn't. You could create a common base class for the both of them but it might be easier to just copy the common functionality.

control-myfeatures/src/main/java/org/oskari/control/myfeatures/ImportMyFeaturesHandler.java
ZakarFin: Agree, on import this is wrong.

service-myfeatures/src/main/java/org/oskari/map/myfeatures/service/MyFeaturesService.java
```diff
@@ -29,6 +30,7 @@ public abstract class MyFeaturesService extends OskariComponent {
 
     public abstract List<MyFeaturesFeature> getFeatures(UUID layerId);
     public abstract List<MyFeaturesFeature> getFeaturesByBbox(UUID layerId, double minX, double minY, double maxX, double maxY);
+    public abstract JSONObject getFeaturesAsGeoJSON(UUID layerId, String srsName);
```
ZakarFin: I feel like we should just format to GeoJSON on control-level instead of having an alternate getter on the service that formats the output of `List<MyFeaturesFeature> getFeatures(UUID layerId);`. Also not sure if we've done that choice before and which way the wind blew that time. I'll dig around the codebase a bit and get back to this.

service-myfeatures/src/main/java/org/oskari/map/myfeatures/service/MyFeaturesServiceMybatisImpl.java
```diff
@@ -193,19 +207,70 @@ public List<MyFeaturesFeature> getFeaturesByBbox(UUID layerId, double minX, doub
...
 
+    @Override
+    public JSONObject getFeaturesAsGeoJSON(UUID layerId, String srs)  {
+        try (SqlSession session = factory.openSession()) {
+            MyFeaturesMapper mapper = getMapper(session);
+
+            List<MyFeaturesFeature> features = mapper.findFeatures(layerId);
+            JSONObject featureCollection = this.toGeoJSONFeatureCollection(features, srs);
+            return featureCollection;
+        }
```
ZakarFin: Probably no need to duplicate the behaviour of the existing method?
```suggestion
            List<MyFeaturesFeature> features =  getFeatures(layerId);
            JSONObject featureCollection = this.toGeoJSONFeatureCollection(features, srs);
            return featureCollection;
```

service-myfeatures/src/main/java/org/oskari/map/myfeatures/service/MyFeaturesService.java
DenverCoder544: Last time around the wind blew from [here ](https://github.com/oskariorg/oskari-server/blob/develop/service-myplaces/src/main/java/org/oskari/myplaces/service/mybatis/MyPlacesFeaturesServiceMybatisImpl.java#L267) and this is pretty much just a mindless copy of that behaviour. We can of course change it around like suggested.

ZakarFin: We can have the functionality on the service (maven module) level, but lets move the GeoJSON formatting from the MybatisImpl to some helper class that just takes in `List<MyFeaturesFeature>` as param and formats it as GeoJSON.

control-myfeatures/src/main/java/org/oskari/control/myfeatures/ImportMyFeaturesHandler.java
```diff
@@ -107,6 +107,8 @@ public class ImportMyFeaturesHandler extends RestActionHandler {
     private static final int MAX_SIZE_MEMORY = 128 * KB;
 
     private static final int MAX_RETRY_RANDOM_UUID = 100;
+    private static final String NATIVE_SRS = "oskari.native.srs";
+    private static final String FALLBACK_NATIVE_SRS = "EPSG:3857";
```
ZakarFin: Not use here?

```suggestion
```

control-myfeatures/src/main/java/org/oskari/control/myfeatures/MyFeaturesFeatureHandler.java
```diff
@@ -133,6 +133,10 @@ public void handlePut(ActionParameters params) throws ActionException {
         ResponseHelper.writeJsonResponse(params, om, feature);
     }
 
+    protected MyFeaturesService getService() {
+        return service;
+    }
+
```
ZakarFin: Is this needed now that the new Handler is not extending this?

```suggestion
```

service-myfeatures/src/main/java/org/oskari/map/myfeatures/service/MyFeaturesService.java
```diff
@@ -8,6 +8,7 @@
 import fi.nls.oskari.domain.map.myfeatures.MyFeaturesFeature;
 import fi.nls.oskari.domain.map.myfeatures.MyFeaturesLayer;
 import fi.nls.oskari.service.OskariComponent;
+import org.json.JSONObject;
```
ZakarFin:
```suggestion
```

PR #1261, opened 2026-04-13
Log4j & tomcat patch version bump

DenverCoder544:


ZakarFin:
Sorry, these were already patched in #1259 and #1258

PR #1265, opened 2026-04-24
Migrate myplaces/userlayers references in mapfull bundles

jampukka:
Builds on the work done in https://github.com/oskariorg/oskari-server/pull/1264

V1_1__migrate_published_myplaces replaces `myplaces_${category_id}` layers used in mapfull bundle conf.layers and state.selectedLayers with their already migrated myfeatures layer `myf_${layer_id}` counterpart.

V1_1__migrate_published_userlayers does the same thing for userlayers.

These V1_1__ migrations rely on the progress tracking mapping tables (myfeatures_categories and myfeatures_user_layer) generated by V1_0__ migrations. Therefore, if you already run V1_0__ migrations long before running these and continued allowing users to generate new myplaces/userlayers and their usage in published maps then there's a chance these migrations will miss some layers (as versioned migrations are run only once).

Additional note: this actually MIGRATES the data - it doesn't just create a copy, like the #1264 migrations do.

### DIFF
--- a/content-resources/src/main/java/flyway/myfeatures_myplaces/V1_1__migrate_published_myplaces.java
+++ b/content-resources/src/main/java/flyway/myfeatures_myplaces/V1_1__migrate_published_myplaces.java
@@ -1,0 +1,130 @@
+package flyway.myfeatures_myplaces;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


jampukka:
Made this a Draft for now as writing down the PR comment made me realize there's nothing limiting this to `PUBLISHED` typed AppSetups (as the name of the migration files would suggest). However, I'm not sure if such limitation is necessary or even wanted.

edit: It's probably better not limiting ourselves to `PUBLISHED` as this also migrates "My Map Views".

PR #1266, opened 2026-04-29
Announcements rss feed

DenverCoder544:
Basis for reading the feed and storing to database. We still need to figure out if and admin may update these external announcements in app ui. In that case we need to somehow decide which fields are allowed to be updated in app and which are not.  

If updating is not allowed, we need to add at least an end date in the feed itself. Or make sure whoever is updating the feed collects garbage on a regular basis.

### DIFF
--- a/control-announcements/src/main/java/org/oskari/announcements/jobs/AnnouncementsRssImportJob.java
+++ b/control-announcements/src/main/java/org/oskari/announcements/jobs/AnnouncementsRssImportJob.java
@@ -1,0 +1,205 @@
+package org.oskari.announcements.jobs;
+
+import fi.nls.oskari.annotation.Oskari;
+import fi.nls.oskari.log.LogFactory;
+import fi.nls.oskari.log.Logger;
+import fi.nls.oskari.util.IOHelper;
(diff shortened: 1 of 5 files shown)


control-announcements/src/main/java/org/oskari/announcements/jobs/AnnouncementsRssImportJob.java
```diff
@@ -0,0 +1,205 @@
...
+        localized.put("title", title == null ? "" : title);
+        localized.put("link", link == null ? "" : link);
+        final JSONObject locale = new JSONObject();
+        locale.put(localeKey, localized);
+        announcement.setLocale(locale);
+
+        final JSONObject options = new JSONObject();
+        options.put("showAsPopup", false);
+        if (guid != null) {
+            options.put(OPTIONS_EXTERNAL_ID, guid);
```
ZakarFin, line 140: Not sure if we should just fail/skip if we don't have guid (or even title/link)

DenverCoder544: That would probably make sense.

PR #1270, opened 2026-05-07
UserLayer/MyFeatures: Add support for direct file upload w/out zip wrapping

jampukka:
Sometimes you don't want to zip a single file.


ZakarFin:
Looks good for merging, but this only adds it to userlayers (not myfeatures)?

jampukka:
That's true, I was thinking about MyFeatures currently using shared code with UserLayer import logic, but that doesn't apply to the control/Handler level obviously. I'll readjust.

PR #1272, opened 2026-05-18
SearchResult: Add contentURL to JSON

jampukka:
Currently contentURL is not written to the client.


ZakarFin:
Related to https://github.com/nlsfi/oskari-server-extras/pull/42

PR #1275, opened 2026-05-20
Layeradmin coverage

DenverCoder544:
* bring coverage info from layer metadata as part of LayerAdmin - action route response
* add ignoreMetadataCoverage flag which porevents using coverage info from metadata as default coverage for layer

related pr in oskari-frontend: https://github.com/oskariorg/oskari-frontend/pull/2992

### DIFF
--- a/control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminHandler.java
+++ b/control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminHandler.java
@@ -111,14 +111,11 @@ public class LayerAdminHandler extends AbstractLayerAdminHandler {
         boolean capabilitiesUpdated = updateCapabilities(ml);
         MapLayerAdminOutput output = getLayerForEdit(params.getUser(), ml);
         final JSONObject attributes = ml.getAttributes();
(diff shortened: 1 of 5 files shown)


control-admin/src/main/java/fi/nls/oskari/control/admin/LayerAdminHandler.java
```diff
@@ -102,9 +106,20 @@ private VectorStyleService getVectorStyleService() {
     @Override
     public void handleGet(ActionParameters params) throws ActionException {
         final int layerId = params.getRequiredParamInt(PARAM_LAYER_ID);
+        String srs = params.getHttpParam(PARAM_SRS, PropertyUtil.get("oskari.native.srs", "EPSG:4326"));
         OskariLayer ml = getMapLayer(params.getUser(), layerId);
         boolean capabilitiesUpdated = updateCapabilities(ml);
         MapLayerAdminOutput output = getLayerForEdit(params.getUser(), ml);
+        final JSONObject attributes = ml.getAttributes();
+        if (!attributes.optBoolean(LayerJSONFormatter.KEY_ATTRIBUTE_IGNORE_COVERAGE, false)) {
```
ZakarFin: I think we might always want to give admin the chance to see the coverage areas, even when they are not used for end-users.

Lets always include coverage and coverageMetadata for admin UI

PR #1289, opened 2026-08-18
Refactor CSWISORecordParser

jampukka:
Refactor from XPath to XmlHelper.
Additionally adds various MD_Metadata parsing improvements.
Also drop the 3-arg constructor of `CSWService` as not used and not intended to be used.

### DIFF
--- a/service-csw/src/test/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParserTest.java
+++ b/service-csw/src/test/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParserTest.java
@@ -6,13 +6,10 @@ import fi.nls.oskari.csw.domain.CSWIsoRecord.DataQualityConformanceResult;
 import fi.nls.oskari.csw.domain.CSWIsoRecord.DataQualityObject;
 import fi.nls.oskari.csw.domain.CSWIsoRecord.DataQualityQuantitativeResult;
 import org.oskari.xml.XmlHelper;
-import org.geotools.referencing.CRS;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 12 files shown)


service-csw/src/test/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParserTest.java
```diff
@@ -21,108 +18,69 @@
...
-            }
-            List <String> lineages = object.getLineageStatements();
-            Assertions.assertTrue(lineages.size() == 1, "There should be only one lineage statement");
-            //Check that lineage doesn't contain localized (SV, EN) content 
-            Assertions.assertTrue((lineages.get(0).length() == 1313), "FI lineage statement shold contain 1313 chars");
-        } catch (Exception e) {
-            e.printStackTrace();
-        }
+        CSWIsoRecord record = CSWISORecordParser.parse(metaDataNode, locale);
+        CSWISORecordDataQualityParser dqParser = new CSWISORecordDataQualityParser();
```
ZakarFin, line 36: Is this used? I can't find a reference to it.

ZakarFin:
Looks good to me, the code is much more readable now and we'll probably have to come back to this when we start doing the planned metadata rendering changes for the frontend. Lets merge and check if we notice any changes now when this is the only thing changed regarding metadata.

service-csw/src/test/java/fi/nls/oskari/csw/helper/CSWISORecordDataQualityParserTest.java
ZakarFin: I'll merge this. The possibly unused var in test can be removed later when we see if this needs any changes.

jampukka: Right, this is legacy as CSWISORecordDataQualityParser is not meant to be constructed, parseDataQualities is now a public static function

PR #1290, opened 2026-08-19
Migrate userguide config

ZakarFin:
The sample-app uses/used a configuration for having just one content "tab" for user guide and referenced a flyoutClazz that has been removed with the React-migration.

This breaks applications that have relied on this functionality with the new React-based frontend implementation. However everything works like before by just removing the flyoutClazz from any configs that have it. For safety, this only removes the flyoutClazz that was used by the sample-app as any other settings might refer to a implementation that is provided by the app itself.

### DIFF
--- a/content-resources/src/main/java/flyway/oskari/V3_4_2__update_userguide.java
+++ b/content-resources/src/main/java/flyway/oskari/V3_4_2__update_userguide.java
@@ -1,0 +1,102 @@
+package flyway.oskari;
+
+import fi.nls.oskari.log.LogFactory;
+import fi.nls.oskari.log.Logger;
+import fi.nls.oskari.util.JSONHelper;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.oskari.helpers.BundleHelper;
+import org.json.JSONObject;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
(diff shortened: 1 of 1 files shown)


content-resources/src/main/java/flyway/oskari/V3_4_2__update_userguide.java
```diff
@@ -0,0 +1,100 @@
...
+        Connection connection = context.getConnection();
+        List<UserGuide> guides = getUserGuides(connection);
+        AtomicInteger errorCount = new AtomicInteger();
+        guides.forEach(guide -> {
+            boolean neededUpdate = removeFlyoutClazz(guide.config);
+            if(neededUpdate) {
+                try {
+                    updateGuide(connection, guide);
+                } catch (SQLException e) {
+                    errorCount.getAndIncrement();
```
jampukka: How would this throw SQLException only for some of the `oskari_appsetup_bundles` rows?

ZakarFin: True, it probably can't. I was initially thinking if there's some problem with the userguide config not being JSON, but removeFlyoutClazz returns false on such case so it should always succeed. I guess something might go wrong with the db connection in the middle of migration, but I guess the try/catch in the loop is rather pointless.

jampukka: Yep, but if something goes wrong at DB connection level then I'd say it's better not to try catch that (the flyway controlled transaction is bound to fail at that point anyway)

ZakarFin: True true, changed the loop to not use lambda so try catch isn't required :)

jampukka (approved):
Looks good to me

PR #1297, opened 2026-08-31
Ignore charset when comparing content type header value

ZakarFin:
There's a similar method in IOHelper to get the charset from content-type header so this is mostly a copy of that one. Not sure if we should throw an exception if content-type is missing, but decided to return it as null if it's not included.

This fixes an issue where https://github.com/oskariorg/oskari-server/blob/3.3.0/service-wfs3/src/main/java/org/oskari/service/wfs3/OskariWFS3Client.java#L133 tries to check the content-type but fails if the charset is included:
ade085b0-ec62-47f8-a5a1-73d8bedef216

### DIFF
--- a/service-base/src/main/java/fi/nls/oskari/util/IOHelper.java
+++ b/service-base/src/main/java/fi/nls/oskari/util/IOHelper.java
@@ -447,12 +447,13 @@ public class IOHelper {
         }
 
         if (expectedContentType != null) {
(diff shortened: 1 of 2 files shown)


jampukka:
Looks already good, but should we go further and make `IOHelper.validateResponse` case-insensitive when it comes to Content-Type checking?

PR #1298, opened 2026-09-02
Derive MVT feature id from the feature id instead of a per-tile counter

jampukka:
Fixes features not being recognized as the same feature across tile boundaries.

VectorTileEncoder was left to assign the MVT feature id, which it does from a counter scoped to one tile. This is clearly problematic for features extending to multiple tiles (even points do due to tile buffers). Instead use a stable hash function to construct the MVT feature id from the original String id. The chance for hash collisions with full 63 bit space is low. (_You mean 64, not 63?_ Nope, go read `toNumericId((String id)`) 

The downsize: Tiles grow by up to 10 bytes per feature: a full-range id varint where the old counter cost one. The three byte-size assertions in SimpleFeaturesMVTEncoderTest are updated accordingly

`_oid` is unchanged. Alternatively this could probably have been solved at frontend level with MVT Format option idProperty, but that has it's own problems as it also deletes the property it hoists to id and frontend expects `_oid` property to exist at multiple other places.

### DIFF
--- a/service-mvt/src/main/java/org/oskari/service/mvt/SimpleFeaturesMVTEncoder.java
+++ b/service-mvt/src/main/java/org/oskari/service/mvt/SimpleFeaturesMVTEncoder.java
@@ -1,5 +1,6 @@
 package org.oskari.service.mvt;
 
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 
 import no.ecc.vectortile.VectorTileEncoder;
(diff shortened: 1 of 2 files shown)


ZakarFin:
Fixes an issue where features tend to have String based ids (with some some string prefix before the numeric feature id) and we need to have a numeric id for the vectortile. I think we've tried removing the prefix and just using the numeric value before, but not all id's are formatted like that. Some might just have a string like a country code as an id. This should fix the problem for good.

